PR #1, opened 2024-08-25
First version of bounding box rabbit tracking.

adamantivm:


jonbinney:
Awesome! Should I do PRs too? I thought it might be easier to just give us all push access and push directly for now, since we're working on separate folders.

jonbinney:
What OS are you on? On 24.04 I pip3 doesn't have your ghetto old version of pytorch :-P
```
ERROR: Could not find a version that satisfies the requirement torch==1.9.0 (from versions: 2.2.0, 2.2.1, 2.2.2, 2.3.0, 2.3.1, 2.4.0
```

jonbinney:
It looks like the annotations file i was using is empty. Is this the right one? https://storage.cloud.google.com/deep-rabbit-hole/rabbit_10min_10fps_div2.mp4

jonbinney:
Oops i mean this : https://storage.cloud.google.com/deep-rabbit-hole/export-data-rabbit-2024-08-19T11%3A24%3A40.154301Z/video_object_tracking/rabbit_vot-6134357890292514816/data-00001-of-00001.jsonl

adamantivm:
> Awesome! Should I do PRs too? I thought it might be easier to just give us all push access and push directly for now, since we're working on separate folders.

I don't know, I don't have a strong opinion. In part, I was playing with how to do this from VSCode, which I didn't use to do. It was an interesting experience, actually! We can discuss what to do during the call tomorrow.

adamantivm:
> What OS are you on? On 24.04 pip3 doesn't have your ghetto old version of pytorch :-P
> 
> ```
> ERROR: Could not find a version that satisfies the requirement torch==1.9.0 (from versions: 2.2.0, 2.2.1, 2.2.2, 2.3.0, 2.3.1, 2.4.0
> ```

Oops, I'm on 20.04. Those versions where chosen automatically by github copilot. I haven't even used pytorch yet. We can make it anything we want.

adamantivm:
> After install just the coco and opencv dependencies, visualization.py runs without errors! I don't see any rectangles in the output video - I can debug a bit. I did just notice that there are TWO rabbits in frame in that video at 8min 20secs!

Yes! there are two for a little while! I actually annotated both a bit so as to see what the annotation file looks like with more than one rabbit.

The correct annotation file is this one: deep-rabbit-hole/rabbit_10min_25fps_div2.mp4.coco.json

It's still imperfect, needs the last 25% push to get it fully properly annotated, but it does its job. And it does have two rabbits at some point - bonus point: identify the second rabbit in the annotation file

jonbinney:
Works! Does _div2 in the file name mean it is at half resolution? 

A couple notes / questions about the annotation:
- is format of the "attributes" section of the annotations part of the COCO format? I didn't see it described in the spec, but it is super useful, especially the "track_id"
- the bounding boxes and areas are floating point. i guess CVAT doesn't keep you to pixel boundaries?

adamantivm:
> Works! Does _div2 in the file name mean it is at half resolution?
Yes, half resolution. I made it up, happy to adopt something better.
> 
> A couple notes / questions about the annotation:
> 
> * is format of the "attributes" section of the annotations part of the COCO format? I didn't see it described in the spec, but it is super useful, especially the "track_id"
I think the whole of the file is the COCO format. The specs available online suck, that's why I wrote my own documentation [here](https://docs.google.com/document/d/1UcaaOx8sj6mJe5jopb8WLIXKv7BPNHIsj8bjxUrWfOo/edit#heading=h.7zpt78gab6st) , annotating the bits that matter

> * the bounding boxes and areas are floating point. i guess CVAT doesn't keep you to pixel boundaries?
Yeah, that was problematic for OpenCV. Funny given that CVAT is maintained by OpenCV, oh well.

In any case, I much rather wished it was fractional from width and height instead, to make it independent from resolution. Likewise, it would be nicer for the frames to be identified by timestamp instead of sequence number, to make it independent of throttling. Turns out that the export format used by Google Vertex AI does both things, but it isn't a recognized format with a python API as far as I could tell so far, so that's why I stuck with Coco for now.

PR #2, opened 2024-08-29
Inference Script scaffold

adamantivm:
Created an inference script that works (even though the inference is useless) and creates a valid annotation file, which at least CVAT.ai will accept and import into a dataset for visualization purposes.

Lots of learnings. Also lots of instructions on how to set-up things are missing, unfortunately.


jonbinney:
- Are you installing grounding dino from source?
- How are you running the test file? Is it as simple as running test_inference.py as a script? Or does python have some tool for running tests?

PR #4, opened 2024-08-31
App for running background subtraction.

jonbinney:
Provides a QT gui for testing different background subtraction parameters. Results are still pretty bad since I spent all of my time on the GUI instead of the algorithm :-)  Adds a "dataset" directory to the root of the repo as a standard place for putting downloaded datasets. Adds a vscode config that allows running of the background subtraction app via a launch configuration, and a .env file to set the PYTHONPATH to find the camera_control python package.

A few notes:
- I've copied @adamantivm's original 10min labeled dataset into our dataset directory, since I needed a datset with more rabbit movement in it for testing.
- The python package is still called "camera_control" because eventually it should become the app to control the camera and zoom in on animals. That will require adding a live "mode" to the app that reads from the camera instead a videofile, and sends PTZ commands.
- At first I opened VSCode with `deep_rabbit_hole/camera_control` as the workspace, and put the vscode config in that directory, and the python virutalenv there too. This was pretty simple,  but i had to open a separate vscode window to look at `object_tracker_0`. Now I'm opening the root of this repo as the VSCode workspace, and I've put a `.vscode` config directory there. I'm thinking we can use one virtual env for all the projects to make things simple. I'm not 100% sure this is the right approach though. Curious to hear what you think.
- I think it will simplify our lives to have a standard folder to put datasets in when we're working on them locally - so I created the `deep_rabbit_hole/datasets` directory. Also not 100% sure this is the best way to do things, but it did could simplify things like launch configurations since we would know the root directory for all the data.
- I'm assuming the name of the video file in our datasets is just "video.mp4" since the dataset name is already long. Happy to change this if we want to go with some other convention.

This code should be *runnable* as long as the python libraries are available at new enough versions. The steps to run should be:
- download the datasets `gcloud storage cp -r gs://deep-rabbit-hole/datasets/* ./` from the datasets directory 
- create a virtual environment in `deep_rabbit_hole/.venv` and `pip install -r camera_control/requirements.txt`
- open the root folder of the repo in vscode
- run the "Camera Control App" launch configuration in vscode. Choose the name of the dataset you want to use - for now choose `rabbits_2024_08_12_25_10min_div2` since the I haven't implemented image resizing in the gui.
- use the slider on the bottom to move to somewhere in the video (can't be the very starts since it needs some frames to model the background)
- click the "Run detection" button int he bottom right. Some info about the detection algorithm will scroll through the console, and if it finds anything, it will draw bounding boxes in the image.
- Change to the "Detection Mask" tab to see which pixels it thinks are non-background

Screenshot from 2024-08-31 18-45-56


camera_control/src/camera_control/camera_control_app.py
```diff
@@ -0,0 +1,114 @@
...
+        frame_number = int((value / 100) * self.total_frames)
+        self.cap.set(cv2.CAP_PROP_POS_FRAMES, frame_number)
+
+        ret, frame = self.cap.read()
+        if ret:
+            frame = cv2.cvtColor(frame, cv2.COLOR_BGR2RGB)
+            height, width, num_channels = frame.shape
+            bytes_per_line = num_channels * width
+            q_img = QImage(
+                frame.data, width, height, bytes_per_line, QImage.Format_RGB888
```
alejandromarcu, line 90: This gives me an error.  I think it's just that we are using different versions of PySide, based on this post: https://forum.qt.io/topic/154279/no-qimage-format_rgb888-pyqt6/2
I'm using PySide 6.7.2 (the latest one), and now that class is found in `QImage.Format.Format_RGB8881`

camera_control/requirements.txt
```diff
@@ -1,3 +1,3 @@
 imutils
 opencv-python
-
+PySide6
```
alejandromarcu, line 3: We should use specific versions here to make sure we don't run into issues for using different versions

camera_control/src/camera_control/camera_control_app.py
```diff
@@ -0,0 +1,114 @@
...
+        self.app_signals.frame_number_updated.connect(self.detection_control_widget.set_frame_number)
+
+        self.running = True
+                
+        self.cap = cv2.VideoCapture(self.video_path)
+        if not self.cap.isOpened():
+            raise ValueError(f"Failed to open video: {self.video_path}")
+        self.total_frames = int(self.cap.get(cv2.CAP_PROP_FRAME_COUNT))
+
+        self.layout = QHBoxLayout(self.central_widget)
```
alejandromarcu, line 64: I get an error here as well, because layout is a method ([code](https://github.com/python-qt-tools/PySide6-stubs/blob/30a43a39c4eda735003097444247876ee18408b9/PySide6-stubs/QtWidgets.pyi#L9253)).  Maybe in the version you are using there was no method?  I fixed it by renaming it to main_layout

camera_control/src/camera_control/camera_control_app.py
jonbinney: I'm also on pyside 6.7.2, how odd! I'll look into this a bit more.

camera_control/requirements.txt
jonbinney: Agreed - does pip let you download any version on any OS? Or do we need to pick a version that is old enough to be available on the oldest OS version that any of us are running? I'm on Ubuntu 24.04 if that matters.

jonbinney:
Yes, merging fast and doing a separate PR for fixes sounds good!

camera_control/requirements.txt
alejandromarcu: I'm on MacOS 14.6.1.  My guess is that if the version is not available for your OS it would just fail.

camera_control/src/camera_control/camera_control_app.py
alejandromarcu: This is what I see if I go to QIMage source. Do you see something similar?
image

camera_control/requirements.txt
alejandromarcu: What version of Python are you using?  I'm on 3.12.2

alejandromarcu: BTW, I just rebased master and tried it again, and now even though VSCode is still showing the errors, the app is working.  Before it was crashing when trying to run predictions

camera_control/src/camera_control/camera_control_app.py
jonbinney: Yep, exactly the same! I agree that `QImage.Format.Format_RGB888` is the right name to use. I'm still curious why this works though. I looked at `QImage.__getattribute__` in the debugger, and it is `<slot wrapper '__getattribute__' of 'Shiboken.Object' objects>` so there's some magic happening...... and indeed `getattr(QImage, "Format_RGB888")` works just fine, even though `Format_RGB888` isn't in `QIMage.__dict___`

camera_control/requirements.txt
jonbinney: What crash were you seeing before? The `Format_RGB888` problem, or something else?

camera_control/src/camera_control/camera_control_app.py
jonbinney: I'm guessing that they did some hack to make enum values appear at their parent scope to try and replicate the way things work in C

camera_control/requirements.txt
alejandromarcu: The crash message was not very clear, but it didn't explicitely mention `FormatRTGB888`. I didn't dig at the time, just fixed the issues that VSCode was showing me.  
I think what could be happening is that I have the Pylance plugin installed, which is the one showing the errors in VSCode.   It may give a false error if they did some hack thing that the static checker can't follow, and if you don't have the plug in (do you?), you wouldn't see the error.

camera_control/src/camera_control/camera_control_app.py
jonbinney: The method exists in my version too, but since it's python it happily allows setting reference to anything :-P

jonbinney:
Added a couple fixes in https://github.com/jonbinney/deep_rabbit_hole/pull/6

PR #8, opened 2024-09-03
Adapts the code to use the Hugging Face Transformers lirary

adamantivm:
Works like a charm. No Grounding DINO set-up needed.

I updated the existing `inference.py` file so that we can see a diff, but also re-added the old implementation as `inference_gd.py` to make side-by-side comparisons and copy/pasting a bit easier

### DIFF
--- a/object_tracker_0/README.md
+++ b/object_tracker_0/README.md
@@ -50,5 +50,5 @@ but basically it comes down to:
  - Activate your venv (VSCode does it for you I think)
  - Go to external/GroundingDINO
  - pip install -e .
- if that rails, try:
+ if that fails, try:
  - python setup.py install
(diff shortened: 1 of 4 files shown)


object_tracker_0/README.md
```diff
@@ -46,4 +49,6 @@ but basically it comes down to:
 
  - Activate your venv (VSCode does it for you I think)
  - Go to external/GroundingDINO
- - pip install -e .
\ No newline at end of file
+ - pip install -e .
+ if that rails, try:
```
alejandromarcu, line 53:
```suggestion
 if that fails, try:
```

alejandromarcu (approved):

jonbinney:
Woohoo, it runs for me! Presumably on the CPU since I haven't fixed my CUDA setup yet, but it outputs bounding boxes!

PR #10, opened 2024-09-04
Improve metrics code

alejandromarcu:
Improved the code for metrics and added unit tests.   I'm using pytest for the tests (seems better than unittest), so I also switched the existing test to it

### DIFF
--- a/object_tracker_0/src/inference.py
+++ b/object_tracker_0/src/inference.py
@@ -118,7 +118,7 @@ def test_model(video_path, annotation_path):
         'categories': [
             {'id': 1, 'name': 'rabbit'}
         ],
-        'images': list(map(lambda x: {'id': x['image_id'], 'file_name': f"frame_{x['image_id'] - 1:06d}.PNG" }, annotations)),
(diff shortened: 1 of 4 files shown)


object_tracker_0/src/inference.py
```diff
@@ -118,7 +118,7 @@ def test_model(video_path, annotation_path):
         'categories': [
             {'id': 1, 'name': 'rabbit'}
         ],
-        'images': list(map(lambda x: {'id': x['image_id'], 'file_name': f"frame_{x['image_id'] - 1:06d}.PNG" }, annotations)),
+        'images': [{'id': x + 1, 'file_name': f"frame_{x:06d}.PNG" } for x in range(frame_count)],
```
alejandromarcu, line 121: This change is because I was getting duplicated entries for images when they had more than 1 annotation.  Also, I think if an image didn't have an annotation it wouldn't appear

adamantivm: Good catch, thank you!

adamantivm (approved):

PR #11, opened 2024-09-07
WIP: Cropping via sliding window

adamantivm:
Adds the ability to process each image in batches, using a sliding window of the configured model input size, to avoid resizing and use the maximum available resolution for object detection.
As a result, it's of course much slower (it requires multiple inference calls instead of one).

TODO: Make all these configurable and move them to arguments:
 - box and label thresholds
 - model input size (need to figure out how to modify the model input file, if that's possible at all)
 - every how many frames to run object detection
 - whether to use tiling or not (aka use resizing)

### DIFF
--- a/object_tracker_0/src/inference.py
+++ b/object_tracker_0/src/inference.py
@@ -52,24 +52,29 @@ def calculate_croppings(pil_frame, sub_frame_height, sub_frame_width):
             croppings.append((left, top))
     return croppings
 
-def detect_objects(model, processor, frame):
+def detect_objects(model, processor, frame, tiling):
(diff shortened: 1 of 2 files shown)


jonbinney:
Do you want us to merge this? Or wait until you remove the WIP?

alejandromarcu:
Looks great, I think the pending things can be done in a follow up PR and you can just merge this one

adamantivm:
ok! I will merge and then follow-up. I was in a hurry yesterday, I just wanted to push what I had on my computer before leaving the coffee shop and computer time for some time.

PR #12, opened 2024-09-08
Live camera capture for camera_control app

jonbinney:
Now camera_control can get frames from _either_ a video file on disk, _or_ a live RTSP stream from a camera. Also added automatic image resizing with window size changes, so this should be usable for high res images. The app doesn't do anything yet with the live captured images; applying the motion detection to them to pick camera targets is my next task.

There are instructions for running the code in `camera_control/README.md` if you're curious. But  unless you've got a camera that supports RTSP, the only change you'll notice is resizing of the images when you resize the app window.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -2,14 +2,25 @@
     "version": "0.2.0",
     "configurations": [
         {
-            "name": "Camera control App",
+            "name": "Camera control App (Live)",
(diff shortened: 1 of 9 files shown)


adamantivm:
Cool stuff. I'll try it later today. I might be able to get it to work with my CCTV cameras.

PR #13, opened 2024-09-11
Adds timing to the inference result

adamantivm:
Records and saves timing and some other stuff in the annotations file, so that we can compare different results as we go.

Also updates the evaluation scripts to use a newer manually labeled file


alejandromarcu (approved):
yey!

PR #14, opened 2024-09-11
Script to describe annotations and update visualization

alejandromarcu:
Provide a new script that given an annotation file outputs a text describing the actions (currently just appear, dissapear and movements).
It also updates the visualize script to show the text, so we can easily see how it matches the video.

I tried it with:
```
python src/visualize.py -v ../datasets/rabbits_2024_08_12_25_10min_div2/video/video.mp4 -a ../datasets/rabbits_2024_08_12_25_10min_div2/annotations/manual_labeling.json -d ../datasets/rabbits_2024_08_12_25_10min_div2/annotations/annotation_description.txt
```

Here's an example of how the visualization looks:
image

### DIFF
--- a/object_tracker_0/tests/test_inference_metrics.py
+++ b/object_tracker_0/tests/test_inference_metrics.py
@@ -1,0 +1,76 @@
+import pytest
+import numpy as np
+from inference_metrics import compute_iou, compute_confusion_matrix_per_image_one_category, compute_confusion_matrix_per_image, compute_confusion_matrix
+from collections import defaultdict
+import json
+import tempfile
+
+@pytest.fixture
+def mock_data():
+    ground_truth = [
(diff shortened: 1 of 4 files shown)


object_tracker_0/tests/test_inference_metrics.py
```diff
@@ -0,0 +1,76 @@
+import pytest
```
alejandromarcu, line 1: This file is unrelated, but realized that I missed it in another PR so just piggybacking it here

adamantivm:
Awesome. One question @alejandromarcu , is it using track_id if available to keep track of which rabbit we're talking about? What happens if the annotations are missing the track_id?

alejandromarcu:
yeah, it uses track_id, so for example in the video there are 2 rabbits, with track_ids 0 and 1, so the rabbits are named rabbit_0 and rabbit_1.  We can later let the LLM name the rabbits.
If there's no track_id it will fail, I'll fix it now so that it would default to 0

PR #20, opened 2024-09-17
Quick hack to make it work for long videos

alejandromarcu:
Since it loads all the frames of the video in memory, a long enough video will exhaust all the memory 🤦 
I followed this very easy hack to fix it: https://github.com/facebookresearch/segment-anything-2/issues/288
It uses the lazy loader, and it changes it to actually be lazy.
It's not ideal to directly change the code in external, since we would need to re-do the changes when we upgrade the version, but my hope is that this will be fixed before we need to upgrade, and worst case we can re-do it or figure out something else.
Another option to not directly change the code would be to partition the video and just work with batches, but I don't think it's worth doing for now if we can just fix it with this hack.

There's also people working on directly supporting video, so eventually one day we'll be able to use it: https://github.com/facebookresearch/segment-anything-2/pull/46/files

### DIFF
--- a/.env
+++ b/.env
@@ -1,3 +1,3 @@
 # So that extensions in VSCode can find our modules.
-PYTHONPATH=./object_tracker_0/src:./camera_control/src:./external/GroundingDINO:${PYTHONPATH}
+PYTHONPATH=./object_tracker_0/src:./camera_control/src:./external/GroundingDINO:./external/segment-anything-2:${PYTHONPATH}
 CUDA_HOME=/usr/local/cuda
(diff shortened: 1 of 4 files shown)


.env
```diff
@@ -1,3 +1,3 @@
 # So that extensions in VSCode can find our modules.
-PYTHONPATH=./object_tracker_0/src:./camera_control/src:./external/GroundingDINO:${PYTHONPATH}
+PYTHONPATH=./object_tracker_0/src:./camera_control/src:./external/GroundingDINO:./external/segment-anything-2:${PYTHONPATH}
```
alejandromarcu, line 2: I needed to add it for it to find sam2.  Are you guys doing it in another way?

adamantivm (approved):
Thanks!

Some update suggestions:
 - Fix the proper passing of the tiling param
 - Idea: add a param to avoid processing ALL frames, instead only pick one every two or five or something like that. Like a throttling. Maybe it's not necessary to use ALL frames for the result to be similar, but with a lot less processing WDYT?

PR #22, opened 2024-09-19
Inference: keep track of ids

alejandromarcu:
When we find new rabbits, we need to reset the state and re-add them all, so that SAM keeps track.  Also, this handles the ids to keep them consistent

### DIFF
--- a/object_tracker_0/src/describe_annotations.py
+++ b/object_tracker_0/src/describe_annotations.py
@@ -30,7 +30,7 @@ def directions_of_movement(prev_x, prev_y, x, y, THRESHOLD = 1):
     else:
         horizontal = ""
 
-    return (horizontal, vertical)    
+    return (horizontal, vertical)
 
 def describe_movement(h, v):
     d = []
@@ -47,16 +47,16 @@ def describe_movement(h, v):
     return " and ".join(d)
 
 
-    
+
 def describe_annotations(filename: str, output_filename: str):
 
     def log(file: TextIOWrapper, image_id: int, actor: str, message: str):
         file.write(f"{image_id:5};{actor};{message}\n")
-    
+
 
     with open(filename, 'r') as f:
         data = json.load(f)
-    
+
     categories_map = {category['id']: category['name'] for category in data['categories']}
 
     annotation_map = defaultdict(lambda: {})
@@ -68,7 +68,7 @@ def describe_annotations(filename: str, output_filename: str):
     min_image_id = min(annotation_map.keys())
     max_image_id = max(annotation_map.keys())
 
-    
+
     position_map = {}
     last_seen = {}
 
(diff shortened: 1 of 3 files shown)


object_tracker_0/src/describe_annotations.py
```diff
@@ -102,13 +102,17 @@ def log(file: TextIOWrapper, image_id: int, actor: str, message: str):
                     # TODO: give more information of where it appeared, e.g. "appeared in the top left corner" or if it just popped up in the middle of the screen
                     log(output_file, image_id, name, "appeared")
 
+            removed_tracks = []
             for track_id, image_id_last_seen in last_seen.items():
                 if image_id - image_id_last_seen > 32:
-                    name = f"{categories_map[annotation_map[image_id][track_id]['category_id']]}_{track_id}"
```
alejandromarcu, line 107: This was a bug, should have been image_id_last_seen, and also you can't alter a collection while iterating

object_tracker_0/src/inference.py
```diff
@@ -69,7 +70,7 @@ def calculate_croppings(pil_frame, sub_frame_height, sub_frame_width):
             croppings.append((left, top))
     return croppings
 
-def do_object_detection(model, processor, pil_frame, tiling = False):
+def do_object_detection(model, processor, pil_frame, tiling = True):
```
alejandromarcu, line 73: I forgot to revert it, but we can leave it set to True by default until we fix the config

object_tracker_0/src/inference.py
```diff
@@ -57,7 +57,8 @@ def convert_bbox_format_2(bbox, shape):
     x0, y0, x1, y1 = bbox
     return [ x0, y0, x1 - x0, y1 - y0 ]
 
-def is_similar(box1, box2, threshold=10):
+# TO DO: probably we should use IoU to be more precise.  E.g. using an absolute threshold, it could be too lenient for small objects and too strict for big ones.
```
adamantivm, line 60: I agree. We should move the IoU calc to an utils file probably and reuse from there.

object_tracker_0/src/inference.py
```diff
@@ -174,7 +175,7 @@ def do_track_objects(predictor, state, bboxes, starting_frame=0, max_frames=10):
     print(f"Adding {len(bboxes)} boxes from tracking from frame {starting_frame} to {starting_frame + max_frames}")
 
     with torch.inference_mode(), torch.autocast("cuda", dtype=torch.float16):
-        for obj_id, bbox in enumerate(bboxes):
+        for obj_id, bbox in bboxes.items():
```
adamantivm, line 178: what's the difference between enumerate and items?

object_tracker_0/src/inference.py
```diff
@@ -246,32 +247,36 @@ def perform_object_tracking(video_path, annotation_path, working_dir, frame_batc
...
             frame_number += frame_batch_size
             continue
 
         # Remove from the already known rabbits from the list of new rabbits to track
-        obj_detect_bboxes = list(filter(lambda newbbox: not any(is_similar(newbbox, oldbbox) for oldbbox in last_track_bboxes), obj_detect_bboxes))
-        if len(obj_detect_bboxes) > 0:
-            print(f"Found new rabbits: {obj_detect_bboxes}")
+        current_bboxes = last_track_bboxes
+        new_bboxes = list(filter(lambda newbbox: not any(is_similar(newbbox, oldbbox) for oldbbox in last_track_bboxes.values()), obj_detect_bboxes))
+        if len(new_bboxes) > 0:
```
adamantivm, line 269:
```suggestion
        # If any new rabbits were found, reset the tracker state and restart tracking from here with the new boxes
        if len(new_bboxes) > 0:
```

adamantivm (approved):
Thanks, Cucu. Can you merge it if my suggestion is good? I also had some questions.

object_tracker_0/src/inference.py
alejandromarcu: I'll do this in a separate PR

object_tracker_0/src/inference.py
alejandromarcu: Initially bboxes was a list of coordinates, so enumerate would just return tuples containing the index and the coordinates.  I changed boxes to be a dict that contains { track_id => coordinates }.  Items returns the pair (track_id, coordinates).  That way, we don't lose the track_id (when using enumerate, the first box was always 0, and so on)

PR #25, opened 2024-09-20
Narration with Llama 3.1 first working version

adamantivm:
It works if running with at least some GPU.

It requires first generating the description (run Describe Annotations launch file) and it generates a corresponding narration.

This is just a first version that works. Many things to build upon, added as comments in the narration.py file.

### DIFF
--- a/object_tracker_0/requirements.txt
+++ b/object_tracker_0/requirements.txt
@@ -5,6 +5,7 @@ huggingface-hub==0.24.6
 pytest
 transformers
 Pillow
+accelerate
 
 # install this one manually
 # -e ../external/segment-anything-2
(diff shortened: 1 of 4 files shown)


object_tracker_0/requirements.txt
```diff
@@ -5,6 +5,7 @@ huggingface-hub==0.24.6
 pytest
 transformers
 Pillow
+accelerate
```
alejandromarcu, line 8: It's not being used right now, is it?

alejandromarcu (approved):
lgtm, and we can discuss in the chat the next steps

object_tracker_0/requirements.txt
adamantivm: It is actually required. When you try to run it without this dependency, it complains and asks to have it installed.

PR #27, opened 2024-09-22
Some playing with Pytorch tensors

adamantivm:
Just doing some operations to understand how to use the basics and get a feel of acceleration.

Some interesting timings happen if I avoid any use of min / max operations on the GPU:

```
GPU timing: Total: 0.03s (1 x 0.03s)
CPU timing: Total: 48.61s (1 x 48.61s)
```


alejandromarcu:
Wow, that's 1500 times faster! 🤯 

Here's what ChatGPT told me about min / max being slow:

Min and max operations can be relatively slow on GPUs for a few key reasons:

1. Branching and Divergence in Control Flow

	•	Min and max operations often involve conditional checks (e.g., if statements) to compare values. In GPU computing, many threads (or warps) execute in parallel. If different threads within a warp follow different branches due to conditional statements, it causes warp divergence. This can make the GPU execute these branches serially, reducing parallelism and thus slowing down execution.

2. Lack of Parallelization in Reduction Operations

	•	Min and max operations are a type of reduction operation, where the result depends on comparing and reducing many values into a single result. While GPUs excel at parallelizing independent computations, reduction operations like min and max require synchronization and a sequence of dependent steps. This limits the extent of parallelism.
	•	For example, finding the minimum of a large array requires comparing pairs of elements, then comparing results of those comparisons, and so on until a single result is obtained. This inherently involves some sequential dependencies.

3. Memory Access Patterns

	•	Min and max operations can involve non-coalesced memory accesses. GPUs are highly optimized for contiguous memory access patterns, where many threads access adjacent memory locations. However, min and max operations might access scattered memory locations, leading to poor memory access efficiency and more latency due to the resulting memory bottleneck.

4. Synchronization Overhead

	•	For reduction operations like min and max, threads must sometimes synchronize to ensure correct results, which can introduce overhead. Synchronization points (such as those used in parallel reductions) can force the GPU to stall while waiting for all threads to finish their work, further reducing the efficiency.

5. Atomic Operations

	•	If min and max are implemented using atomic operations (for instance, atomic min or atomic max), they can be slow because multiple threads may try to update the same value simultaneously, leading to contention. Atomic operations prevent multiple threads from updating the same memory location simultaneously, which introduces serialization, further reducing performance.

Optimizations for Min and Max on GPU

While min and max operations can be slow due to the reasons above, they can be optimized using specific parallel reduction techniques (like tree-based reductions or warp-level primitives) to minimize synchronization, control flow divergence, and memory access inefficiencies.

adamantivm:
I would have thought that pytorch would have already implemented some of those optimizations :man_shrugging:

PR #28, opened 2024-09-22
Basic MLFlow logging for inference script

jonbinney:


jonbinney:
Very easy to setup, and pretty useful. I'll give a quick tour of this on our call tomorrow. One interesting thing it does that I hadn't thought of is logging of system usage:

Screenshot from 2024-09-22 17-35-28

adamantivm:
This is cool!

I have the following questions / next step suggestions: 
 - We should host it somewhere so we can all share results
 - We should also call scoring and annotation of the video within inference.py so that it all ends as results in mlflow
 - We should probably allow dry runs or some sort of way to decide whether to track the run or not
 - I wonder if there's a not too difficult way to offload artifacts to GCS while keeping pointers and easy access to them.

PR #29, opened 2024-09-23
Inference: fix box addition

alejandromarcu:
I realized that if there are no new rabbits we are still re-adding the boxes (without resetting).   SAM2 would take it as refinements, and this probably was causing the OOM.
I just was able to run the 10 min inference in the cloud without OOM!
It run in `Timings => Detection: Total: 945.98s (16000 x 0.06s), Inference: Total: 875.83s (10330 x 0.08s)`.

Metrics:
```
Confusion matrix:
[[ 7954 11406]
 [ 7495     0]]
Precision: 0.41
Recall   : 0.51
F1       : 0.46
```


adamantivm:
That's great! Now we can test things. 
I would guess that the reason you got those results is because IIUC you run it with Tiling and threshold = 0.15 which is too low. I think the magic values are Tiling=True, Threshold=0.35.
It would be great to run it again with these values and also to incorporate the tracking with MLFlow to see this more easily.

PR #31, opened 2024-09-30
Small nits for test runs as of today

adamantivm:
- Check that the input files exist before running documentary :sob: 
 - Use proper default combination of params for object detection (tiling -> 0.35)
 - Use new MLFlow server


jonbinney (approved):
LGTM!

PR #32, opened 2024-09-30
Make it Mac Friendly

alejandromarcu:


adamantivm:
Adapts the remote_setup script so that it works on MacOS

PR #34, opened 2024-09-30
Updates Narration to run locally

adamantivm:
Updated the models to use smaller and newer verisons, it now runs locally in only a few seconds.

See example run: https://soft-pond-5082.ploomberapp.io/#/experiments/0/runs/2aeea555807a4512a99fc6f7a801fba5


jonbinney (commented):
Excellent!

PR #35, opened 2024-10-01
Add another set of annotations to the visualization

alejandromarcu:
With this we can make the visualization show both the results of inference and the ground truth, so that we can compare.
I was able to see that the last 2 minutes or so of ground truth are incorrect, I think we didn't label them yet, Julian will teach me

E.g.

image


jonbinney (approved):
Very useful!

PR #36, opened 2024-10-02
Use video recording time and add it to annotations

adamantivm:
Updated inference to pick the date from the video.
Also included some silly scripts to use against MLFlow

### DIFF
--- a/object_tracker_0/requirements.txt
+++ b/object_tracker_0/requirements.txt
@@ -9,6 +9,7 @@ accelerate
 mlflow
 psutil # So that mlflow can log system metrics
 pynvml # So that mlflow can log GPU metrics
+ffmpeg-python  # To extract video recording time from mp4 file
 
 # install this one manually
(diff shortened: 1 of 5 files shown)


object_tracker_0/requirements.txt
```diff
@@ -9,6 +9,7 @@ accelerate
 mlflow
 psutil # So that mlflow can log system metrics
 pynvml # So that mlflow can log GPU metrics
+ffmpeg-python  # To extract video recording time from mp4 file
```
jonbinney, line 12: I didn't know ffmpeg had a python api - very handy!

jonbinney (approved):
LGTM

object_tracker_0/requirements.txt
adamantivm: I was a bit skeptikal at first, I thought it might launch entire processes to answer small things, but it doens't look like it's doing that, it felt pretty swift on my computer

PR #37, opened 2024-10-02
Script to plot time delta between frames in video

jonbinney:


jonbinney:
frame_deltas

alejandromarcu:
I haven't tried this, but I think we may be able to get timing from python using something like `video.get(cv2.CAP_PROP_POS_MSEC)`, which would have the advantage to run in 1 step vs having to run ffmpeg first.

Another thing that could be interesting to plot would the the cumulative time. In the ideal world of a constant FPS this should be a line, and this would make it easier to see how much it deviates from a line.

jonbinney:
I read online somewhere that `cv2.CAP_PROP_POS_MSEC` doesn't read the actual time from the file, just computes the (frame_number/framerate)`

PR #39, opened 2024-10-08
Automatic closeups of "things of interest"

jonbinney:
The camera control app now has an "autozoom" checkbox which enables automatic detection of moving objects in live images, pans the camera to point at them, zooms in, and captures some video. The camera then returns to a "home" position, specified as preset one on the camera itself.

There are way too many false positives in the motion detection based "thing of interest" detecter which chooses what to look at. Next I plan to try using grounding dino for this.

In adding these features I've broken some of the functionailty when running on a video from disk (instead of live camera images). I'll fix those regressions in another PR.

A couple other changes included in this PR:
- Fixed spacing of detection parameter panel
- Removed gstreamer-python dependency, which was super annoying. The app now uses opencv to read the camera's rtsp stream.
- Added a separate shell script for starting a gstreamer pipeline that writes the camera feed to disk as videos. This is nice because now I can restart the python app without interrupting the pipeline.


jonbinney:
@adamantivm @alejandromarcu

camera_control/src/camera_control/camera_control_app.py
```diff
@@ -102,26 +103,32 @@ def seek_frame(self, frame_number):
...
+    def __init__(self, app_signals, camera_config, detection_parameters):
         super().__init__()
         self.app_signals = app_signals
+        self.detection_parameters = detection_parameters
+        self.autozoom_on = False
         camera_uri = f"rtsp://{camera_config['camera_username']}:{camera_config['camera_password']}@" + \
             f"{camera_config['camera_address']}:{camera_config['camera_rtsp_port']}" + \
             f"/{camera_config['camera_rtsp_path']}"
-        self.camera_streamer = GStreamerCameraCapture(camera_uri)
+        self.camera_capture = cv2.VideoCapture(camera_uri)
```
adamantivm, line 129: nice!

adamantivm (approved):

PR #41, opened 2024-10-09
Install google cloud python package

jonbinney:
And fix typo in cloud login command.

Without these changes, the inference script dies at the end because mlflow can't upload the artifacts, and the GCS login in the cloud script fails with
```
ERROR: (gcloud.auth) Invalid choice: 'application-default-login'.
Maybe you meant:
  gcloud auth application-default login
  gcloud auth login
  gcloud auth application-default print-access-token
  gcloud auth application-default revoke
  gcloud auth application-default set-quota-project

To search the help text of gcloud commands, run:
  gcloud help -- SEARCH_TERMS

```


adamantivm (approved):
Oops

PR #42, opened 2024-10-10
Updates descriptions and narration to use time references and more of a sports commentator style

adamantivm:
This PR changes all the pipeline to preserve the original video times such that the narration refers to specific times when things happen. The idea is to make the description useful to go then find the referred actions in the video.

The result is far for impressive when the LLM produces its output, but I think at this point what is needed now is to work on getting a better LLM, or using one online instead.

### DIFF
--- a/object_tracker_0/src/narration.py
+++ b/object_tracker_0/src/narration.py
@@ -41,8 +41,15 @@ def narrate(description_path, narration_output_path):
         actions = f.readlines()
 
     messages = [
-        {"role": "system", "content": "You are David Attenborough."},
-        {"role": "user", "content": f"Narrate the following actions that occur in a backyard in Raleigh, NC using a descriptive and imaginative way. Name each rabbit that appears with a cute rabbit name: {actions}."}
+        # {"role": "system", "content": "You are David Attenborough."},
+        # {"role": "user", "content": f"Narrate the following actions that occur in a backyard in Raleigh, NC using a descriptive and imaginative way. Name each rabbit that appears with a cute rabbit name: {actions}."}
+        {"role": "system", "content": "You are a sports commentator."},
+        {"role": "user", "content": f"""
+         Narrate the following actions that occur in a backyard in Raleigh, NC.
+         Name each rabbit that appears as rabbit_<number> with a cute rabbit name instead of referring them as rabbit_0, etc.
+         Give human understandable time references for actions, insteade of specifying the timestamp at the beginning of each line.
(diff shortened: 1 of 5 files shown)


scripts/remote_setup
```diff
@@ -35,7 +35,7 @@ fi
 
 # Copy local repo to machine
 echo "Transferring files..."
-rsync -a --progress --exclude=.venv,.git $PWD $CLOUD_USER@$CLOUD_ADDRESS:
+rsync -a --progress --exclude='.venv' --exclude='.git' $PWD $CLOUD_USER@$CLOUD_ADDRESS:
```
jonbinney, line 38: We talked about also transferring `.git` over, right?

object_tracker_0/src/narration.py
```diff
@@ -41,8 +41,15 @@ def narrate(description_path, narration_output_path):
...
 
     messages = [
-        {"role": "system", "content": "You are David Attenborough."},
-        {"role": "user", "content": f"Narrate the following actions that occur in a backyard in Raleigh, NC using a descriptive and imaginative way. Name each rabbit that appears with a cute rabbit name: {actions}."}
+        # {"role": "system", "content": "You are David Attenborough."},
+        # {"role": "user", "content": f"Narrate the following actions that occur in a backyard in Raleigh, NC using a descriptive and imaginative way. Name each rabbit that appears with a cute rabbit name: {actions}."}
+        {"role": "system", "content": "You are a sports commentator."},
+        {"role": "user", "content": f"""
+         Narrate the following actions that occur in a backyard in Raleigh, NC.
+         Name each rabbit that appears as rabbit_<number> with a cute rabbit name instead of referring them as rabbit_0, etc.
```
jonbinney, line 49: Is `<number>` a special thing, or does the LLM just figure out what it means?

adamantivm: It's nothing spectial, I'm just making up how to tell the LLM what to do, with not great results

scripts/remote_setup
adamantivm: Hmm ... git actually takes a while to transfer. I think what we actually want is some way to be able to push changes from the VM to git. Does it make sense to keep transferring the whole git repo before we have a way to push it? Can we save ourselves the transfer in the meantime, and add it back when we implement the mechanism to push commits from the VM?

jonbinney: Good point.

PR #43, opened 2024-10-11
Save the detection score in the annotations

alejandromarcu:
When we do object detection, we get a score for each box.  This change saves it in the annotations, so we can use it later in visualization or somewhere else to be able to calibrate the threshold or better understand what is going on.
Notice that this doesn't happen in every frame, only every (currently) 15.

### DIFF
--- a/object_tracker_0/src/inference.py
+++ b/object_tracker_0/src/inference.py
@@ -304,6 +304,7 @@ def perform_object_tracking(video_path, annotation_path, working_dir, frame_batc
         current_bboxes = last_track_bboxes
         new_bboxes = list(filter(lambda newbbox: not any(is_similar(newbbox, oldbbox) for oldbbox in last_track_bboxes.values()), obj_detect_bboxes))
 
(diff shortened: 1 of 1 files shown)


object_tracker_0/src/inference.py
```diff
@@ -224,7 +226,11 @@ def do_create_annotations(bboxes):
                 },
                 'iscrowd': 0,
                 'segmentation': []
-            })
+            }
+            score = scores.get(frame_idx, {}).get(obj_id, -1)
+            if score > 0:
+                annotation['attributes']['score'] = float(int(score * 100) / 100)
+            annotations.append(annotation)
```
adamantivm, line 233: @alejandromarcu did you try uploading the result to CVAT to confirm it's still compatible? It would be good to keep our format compatible so that we can use that tool if we want to

adamantivm (approved):

object_tracker_0/src/inference.py
alejandromarcu: I just tried and it worked.

PR #44, opened 2024-10-12
Add scores to the visualize script

alejandromarcu:
Since we now write in the annotations the scores when doing object detection, we can easily add the score in the visualization


jonbinney:
Looks good!

PR #47, opened 2024-10-24
Switch to transforms v2 and add requirements.txt.

jonbinney:
transforms.v2 is now recommended, and most of the data augmentation examples use it. I tried various transforms for augmentation, but none of them helped with test performance, so I took them back out. I tried gaussian blur, gaussian noise, and some geometric transforms.

### DIFF
--- a/deep_water_level/src/data.py
+++ b/deep_water_level/src/data.py
@@ -1,16 +1,16 @@
 import os
 import json
 from PIL import Image
-from torchvision import transforms
+from torchvision.transforms import v2
 import torch
 from torch.utils.data import Dataset
 from typing import Tuple
 
 class WaterDataset(Dataset):
-    def __init__(self, annotations_file, images_dir, transform=None):
+    def __init__(self, annotations_file, images_dir, transforms=None):
         self.annotations_file = annotations_file
         self.images_dir = images_dir
-        self.transform = transform
+        self.transforms = transforms
         self.data = self.load_annotations()
 
     def load_annotations(self):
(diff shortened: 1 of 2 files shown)


jonbinney:
@adamantivm @alejandromarcu

deep_water_level/src/data.py
```diff
@@ -46,11 +46,14 @@ def get_data_loaders(
     batch_size: int = 32,
     train_test_split: Tuple[int, int] = [0.8, 0.2],
 ):
-    # Define a transform to apply to the data
-    transform = transforms.Compose([transforms.ToTensor()])
+    transforms = v2.Compose(
+    [
+        v2.ToImage(),
+        v2.ToDtype(torch.float32, scale=True), # convert to float32 and normalize to 0, 1
```
adamantivm, line 52: how does it know how to normalize to 0, 1? Does it use the whole dataset to learn the range or does it make an assumption about it?

My recollection when I implemented Grounding Dino is that I had to pre-calculate the mean and stdev per channel from the dataset in order to normalize correctly. See https://github.com/jonbinney/deep_rabbit_hole/blob/49c35bac4d232497f06aec9df5fdeb025776c65a/object_tracker_0/src/inference_gd.py#L48

Maybe v2 does this automatically or something? do you know @jonbinney ?

jonbinney: Oh, not that NN-specific kind of normalization. This is just scaling the values so that they are between 0 and 1. So for uint8 channels, dividing by 255.

PR #48, opened 2024-10-25
Gradio app to test the model manually

adamantivm:
Adds a quick way to test the model with images manually and see how it does.

TL;DR: It does terribly bad

### DIFF
--- a/deep_water_level/src/train.py
+++ b/deep_water_level/src/train.py
@@ -25,6 +25,8 @@ def load_data():
 def do_training(
         dataset_dir: str,
         annotations_file: str,
+        n_epochs: int,
+        learning_rate: float,
         ):
     # Set-up environment
     device = 'cuda' if torch.cuda.is_available() else 'cpu'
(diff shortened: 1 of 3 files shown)


deep_water_level/src/train.py
```diff
@@ -88,6 +90,8 @@ def do_training(
     parser = argparse.ArgumentParser(description='Train a model on the Deep Water Level dataset')
     parser.add_argument('--dataset_dir', type=str, default='datasets/water_2024_10_19_set1', help='Path to the dataset directory')
     parser.add_argument('--annotations_file', type=str, default='manual_annotations.json', help='File name of the JSON file containing annotations')
+    parser.add_argument('--n_epochs', type=int, default=50, help='Number of epochs to train the model')
+    parser.add_argument('--learning_rate', type=float, default=0.001, help='Learning rate for training the model')
```
jonbinney, line 94: have you noticed much of an effect on the result when changing the learning rate?

adamantivm:
Ehmm ... I only surfaced the parameter. I didn't experiment with different
values, unless I set a different default, in which case it was completely
by accident.

PR #49, opened 2024-10-28
Log dataset parameters

jonbinney:
Logs the list of training and test images used for the run, as well as the transforms used on them.


adamantivm:
@jonbinney are you going to piggiback adding the scatterplot as an artifact to this or should we merge it?

jonbinney:
You can merge this and I'll make a separate pr

PR #50, opened 2024-10-30
Script for plotting water level training data

jonbinney:


jonbinney:
Here is what the output looks like: 
image

The x axis is the time of day and the y axis is  the depth. @adamantivm is this presentation useful? What else should I plot?

adamantivm:
I was thikning more of a continuous timeline where we could see how water level evolves over time, as well as for which time frames we have images, and which of those are labeled.
Then, make it easy to quickly label any continuous swath by inspecting continguous images until we find a change

jonbinney:
Ok I'll add that. It will require downloading the raw data locally, right? Or at least downloading a file listing.

adamantivm:
I think requiring the data, or at least like you say the listing (and the annotations file) will be important.
Having the files locally would help also browse through them as you annotate / explore

jonbinney:
Almost done with an updated tool, I'll open a new PR for that in a bit.

PR #52, opened 2024-11-05
Option to crop images for water level training

jonbinney:
Also added an option to save transformed images to disk, to make it easier to debug the cropping. I tried logging the images to mlflow, but that was way too slow, even for small cropped images.

I didn't see much improvement in test performance from just the cropping, but maybe it will work when combined with other data augmentation methods.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -2,6 +2,20 @@
     "version": "0.2.0",
     "configurations": [
         {
+            "name": "Train water level model",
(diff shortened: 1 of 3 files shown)


adamantivm:
Looks like, reasonably son, you changed the model to accommodate the different size, but I don't see the updated model.py file. Maybe missing a commit @jonbinney ?

adamantivm:
Ah, what great unintended foresight!

PR #53, opened 2024-11-06
Some code generalization and fixes

adamantivm:
- adds cropping to the infer scripts so that it can be played with manually
 - simplifies the calculation of the linear layer in the model
 - adds output normalization as an option

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -2,6 +2,30 @@
     "version": "0.2.0",
     "configurations": [
         {
+            "name": "Test water level model (dataset)",
+            "type": "debugpy",
+            "request": "launch",
+            "program": "deep_water_level/src/infer.py",
+            "console": "integratedTerminal",
+            "args": [
+                "--crop_box", "130", "275", "140", "140",
+                "--model_path", "model.pth",
+                "--dataset_dir", "datasets/water_test_set3",
+                "--annotations_file", "filtered.json",
+            ]
+        },
+        {
+            "name": "Test water level model (single image)",
+            "type": "debugpy",
+            "request": "launch",
+            "program": "deep_water_level/src/infer.py",
+            "console": "integratedTerminal",
+            "args": [
+                "--crop_box", "130", "275", "140", "140",
+                "--model_path", "model.pth",
+            ]
+        },
+         {
             "name": "Train water level model",
             "type": "debugpy",
             "request": "launch",
@@ -12,7 +36,7 @@
                 "--test_dataset_dir", "datasets/water_test_set3",
                 "--annotations_file", "manual_annotations.json",
                 "--crop_box", "130", "275", "140", "140",
-                "--log_transformed_images", "True",
+                "--log_transformed_images", "False",
             ]
         },
         {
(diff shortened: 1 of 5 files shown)


.vscode/launch.json
```diff
@@ -2,6 +2,30 @@
...
+            ]
+        },
+        {
+            "name": "Test water level model (single image)",
+            "type": "debugpy",
+            "request": "launch",
+            "program": "deep_water_level/src/infer.py",
+            "console": "integratedTerminal",
+            "args": [
+                "--crop_box", "130", "275", "140", "140",
```
jonbinney, line 24: I wonder if there is an easy way to store extra parameters like the cropping box in the `model.pth` file since we always want to use the ones we trained with.

deep_water_level/src/model.py
```diff
@@ -9,31 +10,41 @@ def __init__(self, image_size: Tuple[int, int, int] = (3, 810, 510)):
...
-        self.pool2 = nn.MaxPool2d(kernel_size=2, stride=2)
+        self.pool2 = nn.MaxPool2d(kernel_size=2, stride=1)
         self.flatten = nn.Flatten()
-        self.fcn1 = nn.Linear(in_features=self.linear_line_size, out_features=120)
+
+        # Calculate linear size assuming fixed image size
+        self.linear_size = self.calculate_linear_size(image_size)
+
+        self.fcn1 = nn.Linear(in_features=self.linear_size, out_features=120)
+        self.dropout = nn.Dropout(p=0.2)
```
jonbinney, line 24: Did you mean to add a hardcoded dropout here? Does it help performance noticeably?

deep_water_level/src/model.py
```diff
@@ -9,31 +10,41 @@ def __init__(self, image_size: Tuple[int, int, int] = (3, 810, 510)):
...
+    
+    def calculate_linear_size(self, image_size):
+        dummy_input = torch.randn(1, *image_size)
+        output = self.aux_conv_forward(dummy_input)
+        return output.shape[1:][0]
 
-    def forward(self, x):
-        # print(f'start {x.size()}')
+    
+    def aux_conv_forward(self, x):
```
jonbinney, line 33: Does "aux" here mean "auxilliary"?  Why split this part of the convolutions out?

.vscode/launch.json
adamantivm: I was wondering exactly the same. Looks like soem way to store the model parameters is important.
I think it's also important in order to make it easier to run training or inference without having a gazillion arguments to the script

deep_water_level/src/model.py
adamantivm: ah, no. This is an accidental commit, thanks for the catch. I'll remove it for now. If anything it should be added optionally based on a parameter

deep_water_level/src/model.py
adamantivm: Yes. Poor name maybe. I just wanted to abstract away the convolutional portion of the forward pass in order to be able to use it to calculate the linear input size automatically.

PR #55, opened 2024-11-11
Adds Dropout as an optional param to the model

adamantivm:
Not great results after half a test, but wen can test more as needed with this - or not.


adamantivm:
NOTE: this is piggibacked on #54.

alejandromarcu (approved):

PR #56, opened 2024-11-11
Water level labeling tool

jonbinney:
Loads annotated data from a dataset as well as raw images that haven't been labeled yet. Allows viewing the images, as well as interpolating depth values over time for unlabeled images using the known labeled images. Images which don't have labeled depth yet show up as having zero depth on the graph.

Commands:
- left click on a point on the graph to display the corresponding image
- hold down ctrl and left click and drag on the graph to select a time interval of images
- press "i" to interpolate depth values for unlabeled images in the selected interval
- press "w" to save the new labeled depth values to a new annotations file.

We can then use the existing annotation_utils `merge.py` script to add those new annotations to the existing dataset, or create a new dataset. I've got a branch which modifies `merge.py` to also copy the raw images into the appropriate dataset directory, which I'll PR tomorrow.

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -162,6 +162,7 @@ cython_debug/
 
 # Training data 
 /datasets/*/
+/raw_data/*/
 
 # Models
 /models/**/*
--- a/deep_water_level/requirements.txt
+++ b/deep_water_level/requirements.txt
@@ -1,5 +1,7 @@
+matplotlib
 mlflow
 psutil # for mlflow
 pynvml # for mlflow
+pytz
 torch
 torchvision
--- a/deep_water_level/src/labeling_tool.py
+++ b/deep_water_level/src/labeling_tool.py
@@ -1,10 +1,12 @@
 import argparse
+import json
 from matplotlib import pyplot as plt
 import matplotlib.dates
 import numpy as np
 import pandas as pd
 from pathlib import Path
 import re
+import shutil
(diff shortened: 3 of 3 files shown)


jonbinney:
btw, pandas' Dataframe.merge() is pretty amazing. It's basically an SQL join for merging two arrays in various ways.

alejandromarcu (approved):
This tool looks great! I wonder if we can extend it to do what I'm doing for Q&A (measure the distance between the window and reflection) and annotate it somewhere else to keep both values and see the relationship.  We can discuss it in our next meeting.

adamantivm:
This is really cool!
I already have some feature requests, although probably I need to learn more how to use them.
 - I would like  Next / Previous keys, including: next/previous image and next/previous day
 - I would like to be able to zoom in and out / look at subsets of the dataset

I looked at the code to see if I could add some of this but I think it will take me quite some time to do.
Looking forward to chatting about this today!

jonbinney:
You can already zoom in and out on either the data plot or the image using the Matplotlib zoom and home buttons at the bottom. Adding forward/back keys is easy - would you want that to move through just labeled images, or the combination of not labeled and raw images? There are a lot of raw images, so moving through them could take a long time.

adamantivm:
well, I think I need an online matplotlib class. We can do that today.

PR #57, opened 2024-11-11
Script to visualize CNN

alejandromarcu:
This new scripts allows to visualize the kernels in the convolutional layers and the feature maps (i.e. the images passed through those layers)

image


adamantivm:
awesome, thanks!

PR #58, opened 2024-11-13
Script to generate synthetic images

alejandromarcu:
The script has a function to generate the images and annotations, and then generates 3 train/test sets.   If we want to experiment more, we can keep adding variations (e.g. smooth edges, rotations, superimposing other images, etc).
See image below as an example.

After generating the 3 examples, I ran the training and inference.  Here's what happened:
1) This set doesn't add any noise or variation, so for each level there's only 1 possibility.  I used more levels than we have in reality to have enough training data. Also, the test set has levels that the train set doesn't, otherwise it could just exactly memorize it. [MLFlow Link](https://soft-pond-5082.ploomberapp.io/#/experiments/566228862856116089/runs/1ac003f52cab4daa830d462cdb1376d9/model-metrics)
The result seemed reasonable, it was good especially in the middle values, since the initial and last values started to diverge.  I think it may be related to having few samples, maybe it's underfitting.  In any case, not sure if this was a good example, but including it since I did it anyway.

2) Now I introduced more variation with noise, some offset in the top and color variations.  This means that I can have as many images per level as I want, since they will be all different.  [MLFlow Link](https://soft-pond-5082.ploomberapp.io/#/experiments/566228862856116089/runs/80183772aa8c49caa7d52e2ed757d262/model-metrics)
The result seem pretty good, I don't think it's over/under fitting.  When running the inference script I was able to see that most errors are less than 1.0 in the level (BTW, it would be good to make a graph of predicted vs actual to easily see how much it resembles a line, I'll probably do that soon).
It's interesting to see that this result seemed better than 1, even though we had noise and other variations.  I think it's because we had multiple images per level.

3) This is similar to the above but I increased even more the variations. I'm not sure why the loss is so much smaller at the beginning.  The final result doesn't seem too different, so I think it's actually understanding what matters. [MLFlow Link](https://soft-pond-5082.ploomberapp.io/#/experiments/566228862856116089/runs/93d47eb0d0734cb7916c29adb9279ccf/model-metrics).  

I visualized the kernels and they still seem random:
image

And this is how one image look when passing through the first convolutional layer.  Some results seem to be highlighting the "water", but given the inputs that we have, that may not indicarte much.
image

Here's an example image:
image

### DIFF
--- a/deep_water_level/src/generate_images.py
+++ b/deep_water_level/src/generate_images.py
@@ -1,0 +1,99 @@
+import cv2
+import numpy as np
+import os
+import time
+import csv
+
+IMG_WIDTH = 810
+IMG_HEIGHT = 510
+
+
+def generate_data(dir: str, levels: list[float], min_images_per_level = 1, max_images_per_level = 1, pixels_per_level: int = 20, max_noise = 0, max_top_level_offset = 0, max_water_color_variation = 0):
+    """
+    Generates images and annotations and saves them in the given directory.
+
+    Args:
+        dir: the base directory,
+        levels: A list of water level
+        min_images_per_level: the minimum number of images to generate for each level, the exact number will be random
+        max_images_per_level: the maximum number of images to generate for each level, the exact number will be random
+        pixels_per_level: how many pixels represents a level
+        max_noise: the maximum noise to add to the image, or 0 for no noise
+        max_top_level_offset: the maximum random offset to add to the top level, or 0 for no offset
+        max_water_color_variation: the maximum color variation to add to the water, or 0 for no variation
+    """
+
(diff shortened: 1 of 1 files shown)


adamantivm:
How did you choose the traiing parameters? I see you used 0.1 dropout, no cropbox, etc. Any reasoning or just picked a random combination and sticked with it?

Also, in order to better evaluate how it works, it would be nice to know what is the test loss of a "random" model that picks random numbers or always picks the average depth, so that we have a comparison point against the results.
In the case of the real data test sets we know it should be lower than 1.

alejandromarcu:
For the training parameters, I just left the default values from the script.

PR #59, opened 2024-11-16
Move forward/back using keys

jonbinney:
A few changes to the labeling tool:
- l/h for moving forward/back one image
- w/b for moving forward/back one day (saving the annotations file is now ctrl+s)
- selected point(s) on plot are now red instead of drawing a region. this works better for showing single selected regions, whereas before you had to always select an entire region

From the doc string at the top of the file, the commands are:
```
  l - move forward one image
  h - move backward one image
  w - move forward one day
  b - move backward one day
  i - interpolate depths for selected image or region
  ctrl+s - save new annotations as new file
  ctrl+(left click and drag) - select region
```

### DIFF
--- a/deep_water_level/src/labeling_tool.py
+++ b/deep_water_level/src/labeling_tool.py
@@ -1,3 +1,15 @@
+"""
+GUI Tool for labeling water level data
+
+Hotkeys:
+  l - move forward one image
+  h - move backward one image
+  w - move forward one day
+  b - move backward one day
+  i - interpolate depths for selected image or region
+  ctrl+s - save new annotations as new file
+  ctrl+(left click and drag) - select region
+"""
 import argparse
 import json
 from matplotlib import pyplot as plt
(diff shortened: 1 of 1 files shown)


adamantivm:
awesome!

adamantivm:
This is already useful @jonbinney , thank you! I used it for a bit and already spotted an opportunity for improving the latest dataset.
I did one tweak that I pushed directly to main: I added the filename to the title, so I can easily idenify the image to step out of the tool.
I can think of 10000000 improvements, but I think this is not the right thing to do as a whole.
One more important and unfortunately probably pretty difficult limitation to solve is that very soon after using it for a bit of time, it starts lagging and taking too long to move from picture to picture. I see some timings being printed so I image you've already been struggling with it. 
anyway, pretty cool, we can see later where to go from here.

jonbinney:
Yes, the reason it is slow is that I use a naive way of showing which point is selected - it redraws the entire plot. I think there are a few ways to fix this.

PR #63, opened 2024-11-21
Use ruff for python formatting

jonbinney:
We were using the black formatter, but it was configured incorrectly and so it wasn't actually running. Now we're using ruff, since it also provides an import formatter. You'll need to install the ruff extension for vscode (you should be prompted to do so) To format code, press ctrl-shift-I. To organize imports, press alt-shift-O.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,10 +1,7 @@
 {
     "[python]": {
-        "editor.defaultFormatter": "ms-python.black-formatter"
+        "editor.defaultFormatter": "charliermarsh.ruff",
     },
-    "black-formatter.args": [
-        "[\"--line-length\", \"119\"]"
-    ],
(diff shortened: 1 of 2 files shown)


jonbinney:
I set the max line length to 120. This formatter (and the black formatter) are pretty opinionated about the code format, which I like because each project doesn't end up with its own quirky formatting.

alejandromarcu:
Does it format on save? I like when it does, so you don't have to think about doing it each time.

jonbinney:
It can be setup to format on save. I didn't enable that yet since it would make a lot of changes (our files have lots of odd formatting) and I thought we might want to choose when to do that, but I'm fine with enabling it.

adamantivm:
I personally prefer to have more control on when the formatting happens, to avoid surprises, but I can adapt and learn from the experience.

PR #65, opened 2024-11-30
Pushing better default params as of today

adamantivm:
I used the results of various Ray Tune runs, knowing the limitations of our data and metrics, to select at least somewhat better results and also save one "best known good model" to date.

Some notes of how I chose the params: https://docs.google.com/document/d/1_h3u6VsC2pquxDMW0ZL5JwdQC6ZxJw4-CwmKrUR303M/edit?tab=t.0#heading=h.kf9zzdqi9p86


.vscode/launch.json
```diff
@@ -37,7 +37,9 @@
                 "--annotations_file", "filtered.csv",
                 "--crop_box", "130", "275", "140", "140",
                 "--log_transformed_images", "False",
-                "--dropout_p", "0.3",
+                "--n_conf_layers", "3",
```
alejandromarcu, line 40: 3 layers of configuration?  😆 

There was a way to suggest a change but I can't find it

alejandromarcu (commented):
See the comment, there's a typo that needs to be fixed

.vscode/launch.json
adamantivm, line 40:
```suggestion
                "--n_conv_layers", "3",
```

adamantivm: I keep getting that wrong!

PR #69, opened 2024-12-03
Improving Evaluation of Test Results

adamantivm:
I'm looking into why some results that I thought were good enough turn out to be bad.

To start, I'm adding more visualization on the current results. 
Look at this model ([MLFlow Run](https://soft-pond-5082.ploomberapp.io/#/experiments/566228862856116089/runs/537820884eb94f2886b3c478ee4b183a)), which produces as test loss of **0.8976** (seemingly low), but when we look at the results visually, they are pretty bad:

image

After that, I tried also creating a manufactured dataset where the error is always randomly up to 1 (more or less):

image

This led me to observe that reasonable results always have a **test loss lower than 0.35**  <-- This is what we should aim for

### DIFF
--- a/deep_water_level/src/infer.py
+++ b/deep_water_level/src/infer.py
@@ -13,11 +13,12 @@
 # - Add and image and click Submit to get the predicted water level
 import gradio as gr
 import torch
-from PIL import Image
 from model import BasicCnnRegression, BasicCnnRegressionWaterLine
 from data import get_transforms, get_data_loader
 import argparse
-import cv2
+from misc import filename_to_datetime
+import matplotlib.pyplot as plt
+import pandas as pd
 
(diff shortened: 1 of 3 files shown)


deep_water_level/src/infer.py
```diff
@@ -13,11 +13,12 @@
...
 import gradio as gr
 import torch
-from PIL import Image
 from model import BasicCnnRegression, BasicCnnRegressionWaterLine
 from data import get_transforms, get_data_loader
 import argparse
-import cv2
+from misc import filename_to_datetime
+import matplotlib.pyplot as plt
+import pandas as pd
```
jonbinney, line 21: +1 for pandas!

jonbinney (approved):
LGTM!

PR #71, opened 2024-12-05
Format all deep_water_level files

alejandromarcu:
It was driving me crazy that when I did a small change in a file and saved, a lot changed due to format, because then I can't track what are the actual changes.  In this PR I just re-saved all the files so they are formatted with Ruffus.

### DIFF
--- a/deep_water_level/src/data.py
+++ b/deep_water_level/src/data.py
@@ -18,32 +18,31 @@ class WaterDataset(Dataset):
         self.normalize_output = normalize_output
         self.data = self.load_annotations(use_water_line)
 
-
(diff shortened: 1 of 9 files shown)


adamantivm:
Merged it before it could get any old

jonbinney:
Ah, but you didn't sort the imports (alt+shift+o). VSCode treats import sorting as a separate operation from formatting. I can look into that if we want our code to be completely standardized.

alejandromarcu:
I didn't realize this about the imports.  I personally don't care as much.  My problem with the formatting was that once I saved a file, a lot changed and it was very hard to see what were my actual changes vs what was only formatting.

PR #72, opened 2024-12-05
Option to show training data on inference scatter plot

jonbinney:
If the results on the training data looks much better than the test data, then we know we're over fitting. I also added tool tips to show the filename for each point on the scatter plot so that we can track down outliers more easily. To include the inference results on the training dataset, you pass in the `-train_dataset_dir` to `infer.py`:
```
infer.py  --train_dataset_dir=datasets/water_train_set4
```
See the changes in the bottom plot:
Figure_1

### DIFF
--- a/deep_water_level/requirements.txt
+++ b/deep_water_level/requirements.txt
@@ -1,6 +1,7 @@
 gradio
 matplotlib
 mlflow
+mplcursors
 opencv-python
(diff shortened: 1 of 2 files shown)


jonbinney:
This adds a new dependency on mplcursors, so you'll need to `pip install -r deep_water_level/requirements.txt`

PR #73, opened 2024-12-06
When using water line, also use depth

alejandromarcu:
As we discussed, include the depth when training for the water_line

### DIFF
--- a/annotation_utils/misc.py
+++ b/annotation_utils/misc.py
@@ -1,7 +1,10 @@
 # Various utility functions used to manipulate the pilecam annotations
-import json
 import datetime
+import json
+
 import cv2
+import torch
+
 
 def bad_name(dataset_dir, image):
     """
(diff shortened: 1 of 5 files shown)


annotation_utils/misc.py
```diff
@@ -1,63 +1,69 @@
 # Various utility functions used to manipulate the pilecam annotations
```
alejandromarcu, line 1: This file is mostly formatting, when I sent the PR to format the code I missed this file because it's in a different directory.  I only added the `my_device()` function here

deep_water_level/src/infer.py
```diff
@@ -132,7 +132,9 @@ def t2s(t):
         loss += mse
         n_images += len(images)
 
-    print(f"Dataset: {dataset_dir}, Average loss: {loss / len(dataset)}, images: {n_images}, dataset size: {len(dataset)}")
+    print(
```
alejandromarcu, line 135: Not sure why some format things are changing 🤷

annotation_utils/misc.py
```diff
@@ -92,93 +98,109 @@ def get_coordinates_from_segmentation(segmentation):
...
 
     print(f"Total images: {len(data['images'])}")
 
-if __name__ == '__main__':
+
+def my_device():
+    if torch.cuda.is_available():
+        return "cuda"
+
+    if torch.backends.mps.is_available():
```
adamantivm, line 196: interesting! What's MPS?

adamantivm (approved):

adamantivm:
I'm curious if you can find anything visually interesting with this.
In particular, by looking at what the waterline looks like when de depth is particularly wrong, especially if there are many consecutive images with a similar wrong value. That can be seen in the new plots I added on infer.py that show the predicted value over time.

annotation_utils/misc.py
alejandromarcu: It's the GPU in Macs

deep_water_level/src/infer.py
jonbinney: Hopefully it is reaching an equilibrium :-)

PR #74, opened 2024-12-07
Cleanup inference script

jonbinney:
Our inference script has gotten a bit fragile and I keep running into crashes or odd results depending on the arguments I pass it. This PR cleans up a few things:

1. Use the dataset class directly instead of get_data_loader to avoid batching (which is not needed for inference).
2. Store inference results directly in the pandas dataframe instead of having extra lists.
3. Handle multiple outputs smoothly for both inference and plotting so that we don't need to special case when use_water_line is True.

@adamantivm could you confirm that my assumption in point 1 is correct before merging? (that there isn't any point in using get_data_loader and having batching when doing inference). Until now we've two nested loops in the run_dataset_inference function in infer.py and I always get confused about why...... it seems that was because we were iterating over batches, then iterating over the images in each batch. After this PR there are no batches in the inference script.
@alejandromarcu with this PR you should be able to use the plotting when using the water line; one column of plots will be generated for each output of the model. Here's an example with use_water_line=True (ignore the actual values, i made a csv with all zeros for the waterline to test)

Figure_1

### DIFF
--- a/deep_water_level/src/data.py
+++ b/deep_water_level/src/data.py
@@ -73,7 +73,7 @@ class WaterDataset(Dataset):
         data = [parse_line(line) for line in lines]
 
         # Remove any data points with None values
-        data = [d for d in data if all(d[1])]
+        data = [d for d in data if (None not in d[1])]
 
         return data
 
@@ -131,10 +131,10 @@ def get_data_loaders(
     # Create PyTorch DataLoaders for train and test splits
     return (
         torch.utils.data.DataLoader(
-            train_dataset, batch_size=batch_size, shuffle=True, normalize_output=normalize_output
+            train_dataset, batch_size=batch_size, shuffle=True,
         ),
         torch.utils.data.DataLoader(
-            test_dataset, batch_size=batch_size, shuffle=False, normalize_output=normalize_output
+            test_dataset, batch_size=batch_size, shuffle=False,
         ),
     )
 
--- a/deep_water_level/src/infer.py
+++ b/deep_water_level/src/infer.py
@@ -12,15 +12,17 @@
 # - Run this script, then open the Gradio app at http://127.0.0.1:7860/
 # - Add and image and click Submit to get the predicted water level
 import argparse
+from pathlib import Path
 
 import gradio as gr
 import matplotlib.pyplot as plt
 import mplcursors
 import pandas as pd
 import torch
-from data import get_data_loader, get_transforms
+from data import WaterDataset, get_transforms
 from misc import filename_to_datetime, my_device
 from model import BasicCnnRegression, BasicCnnRegressionWaterLine
+import numpy as np
 
 VERBOSE = False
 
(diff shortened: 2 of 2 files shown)


jonbinney:
Looking into merge conflicts....

deep_water_level/src/infer.py
```diff
@@ -69,116 +71,128 @@ def run_gradio_app(model, crop_box=None):
...
-                    "actual": t2s(depth),
-                }
-            )
-            mse += error**2
-        mse /= len(images)
-        if VERBOSE:
-            print(f"MSE ({i}): {mse}")
-        loss += mse
-        n_images += len(images)
+        # Convert tensors to 1d numpy arrays
```
jonbinney, line 102: Converting every thing to numpy arrays instead of tensors here made it easier to store in a dataframe and manipulate later.

deep_water_level/src/infer.py
```diff
@@ -202,15 +216,15 @@ def on_add(sel):
 
     # If these arguments are provided, then the model will be run against the dataset, showing results.
     parser.add_argument(
-        "--dataset_dir", type=str, default="datasets/water_test_set5", help="Path to the dataset directory"
+        "--dataset_dir", type=Path, default="datasets/water_test_set5", help="Path to the dataset directory"
```
jonbinney, line 219: I'm slowly changing all of our path manipulation over to `pathlib.Path` :-)

deep_water_level/src/data.py
```diff
@@ -131,10 +131,10 @@ def get_data_loaders(
     # Create PyTorch DataLoaders for train and test splits
     return (
         torch.utils.data.DataLoader(
-            train_dataset, batch_size=batch_size, shuffle=True, normalize_output=normalize_output
+            train_dataset, batch_size=batch_size, shuffle=True,
```
jonbinney, line 134: I don't know how this ever worked - DataLoader complains that it doesn't expect "normalize_output" as an argument, which makes sense. Am i missing something?

jonbinney:
Merge conflicts resolved, ready to review.

deep_water_level/src/data.py
```diff
@@ -73,7 +73,7 @@ def parse_line(line):
         data = [parse_line(line) for line in lines]
 
         # Remove any data points with None values
-        data = [d for d in data if all(d[1])]
+        data = [d for d in data if (None not in d[1])]
```
alejandromarcu, line 76: is this because 0's won't work with all?

deep_water_level/src/infer.py
```diff
@@ -69,116 +71,128 @@ def run_gradio_app(model, crop_box=None):
...
+            return f"{a.item():.4f}"
+        else:
+            return "[" + ", ".join([f"{x:.4f}" for x in a]) + "]"
+
+    transforms = get_transforms(crop_box=crop_box)
+    dataset = WaterDataset(
+        dataset_dir / "annotations" / annotations_file,
+        dataset_dir / "images",
+        transforms=transforms,
         normalize_output=normalized_output,
```
alejandromarcu, line 93: I think normalize_output is not used anymore in WaterDataset, in that case it should be removed

deep_water_level/src/data.py
alejandromarcu: I had the same question.  But anyway, we're not using cooc anymore for this, so I think it's ok to remove, and also remove it everywhere from this file, no?

deep_water_level/src/infer.py
```diff
@@ -231,13 +245,18 @@ def on_add(sel):
         "--use_water_line",
         type=bool,
         default=False,
-        help="If set, do inference of the water level coordinates as well as depth",
+        help="If set, do inference of the water level coordinates as output instead of depth",
     )
 
-    parser.add_argument("--verbose", type=bool, default=False, help="Print error for each image in the dataset")
+    parser.add_argument(
```
alejandromarcu, line 251: I think one of my PRs switched this to 1 line.  Maybe we have something configured differently?

in ruff.toml I have 'line-length = 120', do you have the same?

deep_water_level/src/infer.py
alejandromarcu: yey, nice clean up!

alejandromarcu (commented):
lgtm, see comments

deep_water_level/src/data.py
jonbinney: Ah yes, exactly. Which i found out because i lazily made an annotations file with `0,0,0,0` for the water line :-)

deep_water_level/src/data.py
jonbinney: `cooc` = `coco`? Whether we use coco or csv doesn't affect this line of code, does it?

deep_water_level/src/infer.py
jonbinney: Agreed - I left it in this PR but we can remove it in another.

deep_water_level/src/infer.py
jonbinney: I changed this line and made the argument `store_true` so that you can call `infer.py --verbose` instead of needing `infer.py --verbose=1`, so i don't think this is a case of the formatter being random. Yeah i also have `line-length=120`

alejandromarcu (approved):

adamantivm:
> Use the dataset class directly instead of get_data_loader to avoid batching (which is not needed for inference).
This is totally fine. It was using batches half because I was lazy to find a better way and half because I kind of liked using the defined interface. I think your change is just fine.

Thanks for the clean-up!

For next time you feel like you want to clean things up: we need to change all our interfaces to take only one "dataset_annotations_path" argument instead of two, as we discussed in the past.

jonbinney:
@adamantivm do you mean removing the `dataset_dir` argument? The one thing that has kept me from doing that so far is that then we'd have to interpret the image paths relative to the location of the annotations file, e.g. `<annotations_file_path>/../images/<filename>` which feels a bit hacky. I'm not against it though.

PR #75, opened 2024-12-09
Preprocessing params feat. Histogram Equalization

adamantivm:
Adds equalization as a parameterizable transform.
Also takes the opportunity to save preprocessing parameters with the model so that it's easier to run inference later on.


deep_water_level/src/data.py
```diff
@@ -91,16 +91,28 @@ def __getitem__(self, index):
...
     transforms_array.extend(
         [
             v2.ToImage(),
             v2.ToDtype(torch.float32, scale=True),  # convert to float32 and normalize to 0, 1
+            v2.Normalize(mean=[0.18653404, 0.20722116, 0.19524723], std=[0.15864415, 0.13673791, 0.12532181]),
         ]
     )
 
+    # From water_train_set4:
+    # Mean: [0.18653404 0.20722116 0.19524723], std: [0.15864415 0.13673791 0.12532181]
```
adamantivm, line 114: @jonbinney here are the mean and std values for our specific data, in case you want to try this for normalization on the test battery you prepared.

You can see how I calculated this on `misc.py` in this PR

deep_water_level/src/deep_water_level/data.py
```diff
@@ -93,18 +93,45 @@ def __getitem__(self, index):
...
+        # IMPORTANT: This works with input in the range [0, 255]
+        transforms_array.append(v2.functional.equalize)
+
     transforms_array.extend(
         [
             v2.ToImage(),
             v2.ToDtype(torch.float32, scale=True),  # convert to float32 and normalize to 0, 1
+            # TODO(adamantivm) Enable normalization. Make it work with equalization, which requires
+            # moving equalization to later and including the necessary dtype conversions back and forth
+            # v2.Normalize(mean=[0.10391959, 0.1423446, 0.13618265], std=[0.05138122, 0.04945769, 0.045432]),
```
jonbinney, line 119: It seems like doing equalization would undo some of the work of the normalization - maybe it wouldn't make sense to do both anyway?

deep_water_level/src/deep_water_level/infer.py
```diff
@@ -37,13 +37,14 @@ def load_model(model_path: Path, train_water_line: bool = False):
 
     checkpoint = torch.load(model_path, weights_only=False, map_location=my_device())
     model_args = checkpoint["model_args"]
+    preprocessing_args = checkpoint["preprocessing_args"] if "preprocessing_args" in checkpoint else None
```
jonbinney, line 40: Ah nice, this will make life much easier

jonbinney (approved):
LGTM!

PR #76, opened 2024-12-10
Importable file structure and Pathlib

jonbinney:
- deep_water_level and annotation_utils now have their code in the folder structure `src/<package_name>/...` and the src directory for each is in our python path so that we can import them by their package name from other code
- code that used `os.path` or `str` for paths has been updated to use `pathlib.Path`
- imports have been ordered using ruff (to do this press `alt-shift-o` in vscode

I didn't touch our older rabbit packages since that would require more testing to avoid breaking things. I've tested our launch configurations and running of individual python files in deep_water_level and annotation_utils. 

I think we're reaching a more standardized state for our code format, which should make things easier. Sorry for the big PR!


adamantivm (approved):
LGTM

PR #77, opened 2024-12-15
Exampe experiment script

jonbinney:
Simple example script so that we have the directory and naming convention. The idea is that rather than modifying `launch.json` or the default args for `infer.py`/`train.py` in place, we can each add python scripts to the experiments directory when we're trying new things. In that script we can call training, inference, and visualization functions directly.

For the scripts I chose the naming convention `year_month_day_author_description.py`. This keeps the scripts roughly sorted so that it is easy to see what people are working on lately, and lets us know who wrote the script.

### DIFF
--- a/deep_water_level/src/deep_water_level/infer.py
+++ b/deep_water_level/src/deep_water_level/infer.py
@@ -75,10 +75,10 @@ def run_dataset_inference(
     model,
     dataset_dir: Path,
     annotations_file: str,
-    normalized_output,
-    crop_box = None,
-    use_water_line = False,
-    equalization: bool = False
+    normalize_output,
+    crop_box=None,
+    use_water_line=False,
+    equalization: bool = False,
 ):
     # Convert numpy array to string for printing
     def a2s(a):
@@ -93,13 +93,13 @@ def run_dataset_inference(
         dataset_dir / "annotations" / annotations_file,
         dataset_dir / "images",
         transforms=transforms,
-        normalize_output=normalized_output,
+        normalize_output=normalize_output,
         use_water_line=use_water_line,
     )
 
     # Run inference on each image in the dataset
     data = []
-    for i, (image, depth, filename) in enumerate(dataset):
+    for image, depth, filename in dataset:
         output = run_inference(model, image)
 
         # Convert tensors to 1d numpy arrays
--- a/experiments/2024_12_09_jon_larger_model.py
+++ b/experiments/2024_12_09_jon_larger_model.py
@@ -1,0 +1,65 @@
+from pathlib import Path
+
+from matplotlib import pyplot as plt
+
+from deep_water_level.infer import load_model, plot_inference_results, run_dataset_inference
+from deep_water_level.train import do_training
+
+annotations_file = Path("filtered.csv")
+train_dataset_path = Path("datasets/water_train_set4")
+test_dataset_path = Path("datasets/water_test_set5")
+crop_box = [130, 275, 140, 140]
+train_water_line = False
+model_filename = Path("model.pth")
+parent_output_dir = Path("../dwl_output")
+
+output_dir = parent_output_dir / f"large_conv"
+output_model_path = output_dir / model_filename
+output_dir.mkdir(parents=True, exist_ok=True)
+
+do_training(
+    # Dataset parameters
+    train_dataset_path,
+    test_dataset_path,
+    annotations_file,
+    # Training parameters
+    n_epochs=4,
+    learning_rate=1e-3,
+    normalize_output=False,
+    crop_box=None,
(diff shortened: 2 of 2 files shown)


deep_water_level/src/deep_water_level/infer.py
```diff
@@ -75,10 +75,10 @@ def run_dataset_inference(
     model,
     dataset_dir: Path,
     annotations_file: str,
-    normalized_output,
-    crop_box = None,
-    use_water_line = False,
```
jonbinney, line 80: @adamantivm do you have format-on-save enabled? I think this file might not have been formatted last time you commited.

deep_water_level/src/deep_water_level/infer.py
```diff
@@ -93,13 +93,13 @@ def a2s(a):
...
         dataset_dir / "images",
         transforms=transforms,
-        normalize_output=normalized_output,
+        normalize_output=normalize_output,
         use_water_line=use_water_line,
     )
 
     # Run inference on each image in the dataset
     data = []
-    for i, (image, depth, filename) in enumerate(dataset):
```
jonbinney, line 102: `i` was unused so i removed the "enumerate"

deep_water_level/src/deep_water_level/infer.py
```diff
@@ -93,13 +93,13 @@ def a2s(a):
         dataset_dir / "annotations" / annotations_file,
         dataset_dir / "images",
         transforms=transforms,
-        normalize_output=normalized_output,
```
jonbinney, line 96: We had this arg named differently in different places.

experiments/2024_12_09_jon_larger_model.py
```diff
@@ -0,0 +1,65 @@
...
+do_training(
+    # Dataset parameters
+    train_dataset_path,
+    test_dataset_path,
+    annotations_file,
+    # Training parameters
+    n_epochs=4,
+    learning_rate=1e-3,
+    normalize_output=False,
+    crop_box=None,
```
alejandromarcu, line 29: Shouldn't this be `crop_box = crop_box`?

experiments/2024_12_09_jon_larger_model.py
```diff
@@ -0,0 +1,65 @@
...
+    n_conv_layers=2,
+    channel_multiplier=2.0,
+    conv_kernel_size=8,
+    conv_stride=2,
+    conv_padding=1,
+    max_pool_kernel_size=2,
+    max_pool_stride=1,
+    # Configuration parameters
+    log_transformed_images=False,
+    train_water_line=False,
```
alejandromarcu, line 41: Same here, I think if you define the variable above, I'd expect it to be used here

experiments/2024_12_09_jon_larger_model.py
```diff
@@ -0,0 +1,65 @@
...
+    output_model_path=output_model_path,
+)
+
+# Load the model and run inference using it
+model, model_args, preprocessing_args = load_model(output_model_path, train_water_line)
+
+train_df = run_dataset_inference(
```
alejandromarcu, line 48: this takes a crop_box argument, I think it will be better to pass it, otherwise using it above will break things.  Same below

experiments/2024_12_09_jon_larger_model.py
```diff
@@ -0,0 +1,65 @@
...
+    model,
+    train_dataset_path,
+    annotations_file,
+    **preprocessing_args,
```
alejandromarcu, line 52: If I'm following this correctly, processing_args will have the parameters used in `create_model` (e.g. image_size, dropout_p, n_conv_layers, etc), which this method doesn't take.  Any reason to pass this? I'd think it would break if it contains any argument that is not defined in `run_dataset_inference`

deep_water_level/src/deep_water_level/infer.py
adamantivm: Would you remind me how to enable it?

experiments/2024_12_09_jon_larger_model.py
jonbinney: Oops! Yeah I'll update this PR once i fix this

experiments/2024_12_09_jon_larger_model.py
jonbinney: Yeah I think I'm using these wrong - lets talk about it on the call

deep_water_level/src/deep_water_level/infer.py
jonbinney: In your user settings
```
   "[python]": {
        "editor.formatOnSave": true
    },
```

PR #79, opened 2024-12-17
Change loss calculation for training

alejandromarcu:
The loss for test is calculated by averaging per batch, but for training, it was just taking the last loss of the batch.  Now the graph looks smoother, and I think maybe that can be the cause that sometimes the test loss is less than the training loss

### DIFF
--- a/deep_water_level/src/deep_water_level/train.py
+++ b/deep_water_level/src/deep_water_level/train.py
@@ -143,6 +143,7 @@ def do_training(
     for epoch in range(n_epochs):
         # Train for the epoch
         model.train()
+        train_loss = 0
         for i, (inputs, labels, filenames) in enumerate(train_data):
             optimizer.zero_grad()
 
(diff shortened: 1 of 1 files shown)


adamantivm:
Are you sure this is the correct way to do it? I always thought the loss should be calculated per batch, not epoch. Did you search online for good practices on training code?

adamantivm:
I mean, what you see is not a bug. It may be wrong, but it was intentional.

alejandromarcu:
The loss is actually still calculated per batch, and then we do a backprop and an optimizer step.  I didn't change that. What I changed was that we need to output the current loss to the terminal and mlflow, and train and test were doing it differently: train was using just the loss of the last batch in the epoch, while test was averaging the batches in the epoch.  I think they both should be done equally to be able to better compare the losses.
In this example they print the losses every 1000 batches, so no here or there: https://pytorch.org/tutorials/beginner/introyt/trainingyt.html#the-training-loop

PR #81, opened 2024-12-24
DRAFT: include a confidence value

alejandromarcu:
Just to show what I was trying for getting a confidence value


deep_water_level/src/deep_water_level/train.py
```diff
@@ -125,25 +126,35 @@ def do_training(
...
             optimizer.zero_grad()
 
             inputs = inputs.to(device)
             labels = labels.to(device)
             outputs = model(inputs)
-            loss = criterion(outputs, labels)
+            # loss = criterion(outputs, labels)
+            mu, log_var = outputs[:, 0], outputs[:, 1]
+
+            loss = 0.5 * log_var + 0.5 * ((labels[:, 0] - mu) ** 2) / torch.exp(log_var)
```
alejandromarcu, line 144: This is the important part

jonbinney: Where does the `torch.exp` come from in the denominator of the second part? Shouldn't it be `log_var**2`?

jonbinney: oh nevermind.... log_var is already the log of the variance so it `torch.exp` just results in the variance. But shouldn't it be squared?

jonbinney: ok ok i see..... log var is "log of variance" is already `log(sigma**2)` so the squared is already in there.

PR #83, opened 2024-12-28
Manual early termination

adamantivm:
Adds a signal handler so that if you interrupt the call, it will finish the current epoch and save the model.


jonbinney (approved):
LGTM

PR #84, opened 2025-01-10
Fine-tuning for DeepLabV3 Segmentation

alejandromarcu:
Fine tune this existing model and parameters to be able to recognize the water.
This result is with only 5 epochs on a training set of 391 images, on an image that is neither in training or test.

image

Based on the test loss increasing, seems to be overfitting. Maybe we need to use augmentation to reduce the overfitting, but still seems like a promising start.

image

### DIFF
--- a/deep_water_level/src/deep_water_level/deeplabv3/train.py
+++ b/deep_water_level/src/deep_water_level/deeplabv3/train.py
@@ -1,0 +1,115 @@
+import time
+from pathlib import Path
+import os
+import torch
+import torch.nn as nn
+from annotation_utils.misc import my_device
+from deep_water_level.deeplabv3.utils import SegmentationDataset
+from torch.optim.adam import Adam
+from torch.utils.data import DataLoader
+from torchvision import models
+from torchvision.models.segmentation import DeepLabV3_ResNet50_Weights
+from torchvision.models.segmentation.deeplabv3 import DeepLabHead
+from torchvision.transforms import functional as F
+
+import mlflow
+
+
+def do_training(
+    # Dataset parameters
+    train_dataset_dir: Path,
+    test_dataset_dir: Path | None,
+    # Training parameters
+    n_epochs: int = 40,
+    batch_size: int = 8,
+    learning_rate: float = 1e-3,
+):
+    if test_dataset_dir is None or test_dataset_dir == train_dataset_dir:
+        # Split the train dataset in test and train datasets
+        dataset = SegmentationDataset(train_dataset_dir, train_dataset_dir / "annotations" / "filtered.json")
+
+        train_dataset, test_dataset = torch.utils.data.random_split(dataset, [0.8, 0.2])
+    else:
+        train_dataset = SegmentationDataset(train_dataset_dir, train_dataset_dir / "annotations" / "filtered.json")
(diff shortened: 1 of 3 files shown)


deep_water_level/src/deep_water_level/deeplabv3/utils.py
```diff
@@ -0,0 +1,78 @@
...
+
+import torch
+from PIL import Image
+from torch.utils.data import Dataset
+from torchvision.transforms import functional as F
+
+
+def preprocess_image(image):
+    transform = F.to_tensor(image)
+    transform = F.normalize(transform, mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225])
```
adamantivm, line 12: did you look at the latest versions of what we're doing to preprocess in the existing codebase?
I think equalization worked much better than normalization. And also for normalization, we need to adapt it to the means and stddev of our specific dataset.

deep_water_level/src/deep_water_level/deeplabv3/train.py
```diff
@@ -0,0 +1,115 @@
...
+
+    print(f"Training dataset size: {len(train_dataset)}, Test dataset size: {len(test_dataset)}")
+
+    device = my_device()
+
+    model = models.segmentation.deeplabv3_resnet50(weights=DeepLabV3_ResNet50_Weights.DEFAULT)
+    model.classifier = DeepLabHead(2048, num_classes=2)
+    model.to(device)
+
+    criterion = nn.CrossEntropyLoss()
```
adamantivm, line 44: This is the loss because in the end it's classifying between different classes, right?

I wonder if it would be possible to adapt it to only work on two classes: water and "not water". In that case the criterion may need to be Binary Cross Entropy loss, right?

adamantivm (approved):
Cool stuff!

One high-level question I have is: was there a strong reason for you to replicate everything (data loader, training, testing, utils) on a separate directory?
I like the idea to have clear separation between one experiment and the other, but on the other hand I feel we're losing much of what we learned to do (e.g.: the preprocessing, or configuring MLFlow to store artifacts) because you started from scratch again.
I have a slight impression that generalizing the code we have a bit to run with this different model may yield a slightly better result, at the cost of maybe lack of clarity of what is where. Perhaps that can be addressed with a bit of clever refactoring and documentation?

deep_water_level/src/deep_water_level/deeplabv3/utils.py
alejandromarcu: yeah, good point, the values I used are for the pretrained models, but if we're fine-tunning I guess other values could be better

alejandromarcu:
> One high-level question I have is: was there a strong reason for you to replicate everything (data loader, training, testing, utils) on a separate directory?
 
Yeah, good point about this.  What happened was that we (ChatGPT and I :p) started writing code from scratch for this, and it was pretty different, just a quick prototype to see if this would remotely work.  Then when I started making the code more decent I did a lot of copy-paste from our existing code.  Looking again now I can see how they have a lot more in common than differences, so I will work on refactoring it.

adamantivm:
FWIW I think that starting from scratch to go faster is a good idea. 
Now if it starts to look promising perhaps it's worth the effort to see what can be re-used.

PR #86, opened 2025-01-10
Fine tuning ResNet50+Places365 adapted to Image Regression

adamantivm:
In this experiment, I'm using a Resnet-50 netwok pre-trained from scratch with the Places 365 dataset (which includes swimming pools) as a base, and then fine-tuning it by re-training it after replacing the last output layer for a 1-output FCN.

We'll see how it goes. I may still tweak options and parameters to try various things.

 I'm saving all this under the experiment name: "Deep Water Level Training - pretrained").

Don't merge until I finished playing around with it please.

### DIFF
--- a/deep_water_level/src/deep_water_level/train.py
+++ b/deep_water_level/src/deep_water_level/train.py
@@ -22,9 +22,22 @@ def signal_handler(signal, frame):
     print("Interrupt requested. Will finish this epoch, save the model and then exit")
 
 
-def create_model(train_water_line, **kwargs):
+def create_model(train_water_line, use_pretrained=False, **kwargs):
     # Create the model
-    if train_water_line:
+    if use_pretrained:
+        # Instantiate the model
+        model = torchvision.models.resnet50(num_classes=365, weights=None)
+        # Load the pretrained weights
+        # TODO: Download from the Web if needed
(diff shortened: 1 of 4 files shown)


deep_water_level/src/deep_water_level/train.py
```diff
@@ -22,9 +22,22 @@ def signal_handler(signal, frame):
...
+        # Load the pretrained weights
+        # TODO: Download from the Web if needed
+        checkpoint = torch.load(
+            "models/weights/resnet50_places365.pth.tar", map_location=my_device(), weights_only=True
+        )
+        # Load the pretrained weights in the model
+        state_dict = {k.replace("module.", ""): v for k, v in checkpoint["state_dict"].items()}
+        model.load_state_dict(state_dict)
+        # Replace the last fully connected layer
+        model.fc = nn.Linear(model.fc.in_features, 1)
```
jonbinney, line 39: Is `fc` a special member variable in pytorch? Or is that just what resnet calls its fully connected layer?

adamantivm: It's the latter. This seems to be a common approach as I found it on tutorials and also in Cucu's PR.

PR #87, opened 2025-01-16
Refactor how we choose the model

alejandromarcu:
Initially we had just one model (BasicCnnRegression), then I added BasicCnnRegressionWaterLine and started a bad pattern by adding the parameter use_water_line to chose between the 2 of them. Julian added another model and used the parameter use_pretrained.  Now I want to use a new model for segmentation, and adding a new parameter would be a mess and is getting harder to follow. Instead, I prefer to pass a model name that can be expanded to include new models and there's no need to be passing new parameters everywhere.

### DIFF
--- a/deep_water_level/src/deep_water_level/train.py
+++ b/deep_water_level/src/deep_water_level/train.py
@@ -11,7 +11,7 @@ from utils import start_experiment  # From object_tracker_0 package
 import mlflow
 from annotation_utils.misc import my_device
 from deep_water_level.data import get_data_loader, get_data_loaders
-from deep_water_level.model import BasicCnnRegression, BasicCnnRegressionWaterLine
+from deep_water_level.model import BasicCnnRegression, BasicCnnRegressionWaterLine, ModelNames
 
 shutdown_requested = False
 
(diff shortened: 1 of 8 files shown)


deep_water_level/src/deep_water_level/train.py
```diff
@@ -37,10 +37,13 @@ def create_model(train_water_line, use_pretrained=False, **kwargs):
...
         model.fc = nn.Linear(model.fc.in_features, 1)
-    elif train_water_line:
+    elif model_name == "BasicCnnRegressionWaterLine":
         model = BasicCnnRegressionWaterLine(**kwargs)
-    else:
+    elif model_name == "BasicCnnRegression":
         model = BasicCnnRegression(**kwargs)
+    else:
+        raise ValueError(f"Unknown model name: {model_name}")
+
```
adamantivm, line 46: If we could return `model_name` as part of `kwargs` here, then the model name would get saved with the model as part of model_args and it wouldn't be necessary to pass it as an additional argument in inference scripts!

adamantivm (approved):
I love it, thank you Cucu!
I added just one suggestion, but that could be done as another PR

deep_water_level/src/deep_water_level/train.py
alejandromarcu: Sounds good, I'll do it in another PR

PR #89, opened 2025-01-17
Save the model name in the model file

alejandromarcu:
With this, when we do inference we can just pass the model file name and it will get the model name from the file


adamantivm:
awesome, thanks Cucu!

PR #92, opened 2025-01-22
Add DeepLabV3 model for segmentation

alejandromarcu:
This is a follow up on https://github.com/jonbinney/deep_rabbit_hole/pull/84
Now that the code has been refactored, I can reuse the scaffolding.  There's an experiment to fine tune the segmentation and a new script to show the inference results for segmentation, since it's different from our usual infer

### DIFF
--- a/deep_water_level/src/deep_water_level/data.py
+++ b/deep_water_level/src/deep_water_level/data.py
@@ -28,6 +28,8 @@ class WaterDataset(Dataset):
     def load_annotations(self):
         suffix = self.annotations_file.suffix
         if suffix == ".json":
+            if self.model_name == "DeepLabV3":
+                return self.load_coco_for_segmentation()
             return self.load_coco()
         elif suffix == ".csv":
             return self.load_csv()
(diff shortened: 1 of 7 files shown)


deep_water_level/src/deep_water_level/data.py
```diff
@@ -222,11 +266,13 @@ def get_data_loaders(
             train_dataset,
             batch_size=batch_size,
             shuffle=True,
+            drop_last=True,
         ),
         torch.utils.data.DataLoader(
             test_dataset,
             batch_size=batch_size,
             shuffle=False,
+            drop_last=True,
```
adamantivm, line 275: what is drop_last, here and in the constructor below?

deep_water_level/src/deep_water_level/train.py
```diff
@@ -224,17 +229,22 @@ def do_training(
...
 
-    torch.save(
-        {
-            "model_state_dict": model.state_dict(),
-            "model_args": model_args,
-            "preprocessing_args": preprocessing_args,
-        },
-        output_model_path,
-    )
+    if model_name == "DeepLabV3 QQ can I remove?":
```
adamantivm, line 237: LOL KEDIZE?

deep_water_level/src/deep_water_level/data.py
alejandromarcu: If the last batch has size 1, the training breaks, because it's taken as a single element rather than a batch (not sure why exactly, but seemed like a Pytorch thing).   When you have big batches is unlikely to happen, but since I'm using a batch size of 2, I hit this issue.  Using this parameter would drop the last batch if it's not "full"

deep_water_level/src/deep_water_level/train.py
alejandromarcu: ooops, will fix it

PR #94, opened 2025-01-24
Raspi inference code

adamantivm:
I'm including this script that I use for inference on the raspberry pi.

It's missing more information to understand how I'm using it, but noting how I just had it running for some time using the wrong cropping, I thought it's better to include it in the repo so I can have some review from you guys to help me find stupid errors like this one that I just committed.

From now on, I'm not going to modify the script directly in the raspi anymore, I'm always going to modify it here first and then copy it over.


alejandromarcu (approved):
lgtm

PR #95, opened 2025-01-25
[DO NOT MERGE] WIP: Calculate multiple croppings per inference

adamantivm:


adamantivm:
Merging this even if incomplete as we switch to work on something else, so that we don't lose this - it's deployed on the raspberry pi already.

PR #96, opened 2025-02-08
Data Augmentation for masks in segmentation training

alejandromarcu:
The main change is that the transforms can now process an image and a mask at the same time.  The mask goes only through the geometrical transformations, e.g. no color jitter or equalization.  The random parameters are set equally for the image and the mask.  We now allow the crop_box to be a function, so that we can do something more custom, e.g. for this training I allowed the crop box to take random sizes given some constraints.  It also allows to do a resize at the end.

Still, the result is far from great, but we can continue iterating if we want.  E.g. after training for 100 epochs I got:
`Epoch [100/100], Loss: 0.0141, Test loss: 0.5272`

image

### DIFF
--- a/deep_water_level/src/deep_water_level/data.py
+++ b/deep_water_level/src/deep_water_level/data.py
@@ -3,7 +3,7 @@ import json
 import os
 import random
 from pathlib import Path
-from typing import List, Tuple
+from typing import Callable, List, Optional, Tuple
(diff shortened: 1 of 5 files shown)


jonbinney (approved):
@alejandromarcu and I talked through this at our meeting and it looks good!

PR #97, opened 2025-03-01
Quoridor env basic skeleton

adamantivm:
Ok, here's a start of a Petting Zoo environment for Quoridor.
It's missing several things, even for basic rules, but it's a start.
It has the basic state for pieces and walls. It can also render the current state in text now.
It's missing all the rules and whatever needs to be done to provide the action space to the player.
It has a main function to test out by providing manual functions to see how the board behaves.
That's all for now. I reflected some of these TODOs on the code.

### DIFF
--- a/deep_quoridor/src/quoridor_env.py
+++ b/deep_quoridor/src/quoridor_env.py
@@ -139,25 +139,27 @@ class QuoridorEnv(AECEnv):
                     board_str += " .  "
 
                 # Render vertical walls (|)
-                if row < self.wall_size and col < self.wall_size:
-                    if self.walls[row, col, 0]:
+                if col < self.wall_size:
+                    if row < self.wall_size and self.walls[row, col, 0]:
                         board_str += "|"
                     elif row > 0 and self.walls[row - 1, col, 0]:  # Continuation of a vertical wall
                         board_str += "|"
                     else:
                         board_str += " "
+                else:
+                    board_str += " "
 
             board_str += "\n"
 
             # Render horizontal walls (───)
             if row < self.wall_size:
-                for col in range(self.wall_size):
+                for col in range(self.board_size):
                     if col < self.wall_size and self.walls[row, col, 1]:
-                        board_str += "───"  # Wall segment
+                        board_str += "──── "  # Wall segment
                     elif col > 0 and self.walls[row, col - 1, 1]:
-                        board_str += "───"  # Continuation of a wall from the left
+                        board_str += "──── "  # Continuation of a wall from the left
                     else:
-                        board_str += "   "
+                        board_str += "     "
                 board_str += "\n"
 
         print(board_str)
--- a/deep_quoridor/src/quoridor_env.py
+++ b/deep_quoridor/src/quoridor_env.py
@@ -1,3 +1,4 @@
+from typing import Tuple
 from pettingzoo import AECEnv
 from gymnasium import spaces
 import numpy as np
@@ -73,17 +74,39 @@ class QuoridorEnv(AECEnv):
 
         self._next_player()
 
-    def rowcol_to_idx(self, row, col):
-        return row * self.board_size + col
+    def action_params_to_index(self, row: int, col: int, action_type: int = 0) -> int:
+        """
+        Takes action parameters (row, col, movement_type) to an action index
+        movement_type = 0 for moving, 1 for horizontal wall placement, 2 for vertical wall placement
+        """
+        if action_type == 0:
+            return row * self.board_size + col
+        elif action_type == 1:
+            return self.board_size**2 + row * self.wall_size + col
+        elif action_type == 2:
+            return self.board_size**2 + self.wall_size**2 + row * self.wall_size + col
+        else:
+            raise ValueError(f"Invalid action type: {action_type}")
 
-    def idx_to_rowcol(self, idx):
-        return divmod(idx, self.board_size)
+    def action_index_to_params(self, idx) -> Tuple[int, int, int]:
+        """
+        Takes an action index to action parameters (row, col, movement_type)
+        movement_type = 0 for moving, 1 for horizontal wall placement, 2 for vertical wall placement
+        """
+        if idx < self.board_size**2:
+            return divmod(idx, self.board_size)
+        elif idx >= self.board_size**2 and idx < self.board_size**2 + self.wall_size**2:
+            return divmod(idx - self.board_size**2, self.wall_size)
+        elif idx >= self.board_size**2 + self.wall_size**2 and idx < self.board_size**2 + (self.wall_size**2) * 2:
+            return divmod(idx - self.board_size**2 - self.wall_size**2, self.wall_size)
+        else:
+            raise ValueError(f"Invalid action index: {idx}")
 
     def _move(self, agent, action):
         """
         TODO: Only allow valid moves
         """
-        row, col = self.idx_to_rowcol(action)
+        row, col = self.action_index_to_params(action)
         if (row, col) in self.positions.values():
             return  # Invalid move (occupied)
         self.positions[agent] = (row, col)
--- a/deep_quoridor/src/quoridor_env.py
+++ b/deep_quoridor/src/quoridor_env.py
@@ -12,7 +12,7 @@ class QuoridorEnv(AECEnv):
         self.render_mode = "ansi"
 
         self.board_size = 9  # 9x9 grid
-        self.wall_size = 8  # 8x8 grid for walls
+        self.wall_size = self.board_size - 1  # 8x8 grid for walls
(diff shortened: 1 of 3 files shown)


jonbinney:
I'll take a look at this tomorrow as I play with the zoo.

adamantivm:
Great comments @alejandromarcu ! 
I won't get back to this until a couple of days from now. 
If you feel so inclined, feel free to hijack the PR and apply those changes yourself. 
Otherwise, I'm happy to work on those when I get back to this.

alejandromarcu:
You seem excited to work on it, we can wait for you!

deep_quoridor/src/quoridor_env.py
```diff
@@ -0,0 +1,170 @@
+from pettingzoo import AECEnv
+from gymnasium import spaces
+import numpy as np
+
+
+class QuoridorEnv(AECEnv):
+    metadata = {"render_modes": ["ansi"], "name": "quoridor_v0"}
+
+    def __init__(self):
```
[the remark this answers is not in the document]
jonbinney: +1. Would definitely be handy for debugging

deep_quoridor/src/quoridor_env.py
```diff
@@ -0,0 +1,170 @@
...
+        self.possible_agents = ["player_1", "player_2"]
+        self.agent_order = self.possible_agents.copy()
+        self.agent_selection = self.agent_order[0]
+
+        self.observation_spaces = {
+            agent: spaces.Dict(
+                {
+                    "board": spaces.Box(0, 2, (self.board_size, self.board_size), dtype=np.int8),
+                    "walls": spaces.Box(0, 1, (self.wall_size, self.wall_size), dtype=np.int8),
+                    "walls_remaining": spaces.Discrete(self.max_walls + 1),
```
alejandromarcu, line 27: I think we need to have the walls_remaining for both players

deep_quoridor/src/quoridor_env.py
```diff
@@ -0,0 +1,170 @@
...
+        row, col = self.idx_to_rowcol(action)
+        if (row, col) in self.positions.values():
+            return  # Invalid move (occupied)
+        self.board[self.positions[agent]] = 0
+        self.positions[agent] = (row, col)
+        self.board[row, col] = 1 if agent == "player_1" else 2
+
+    def _place_wall(self, agent, action):
+        """
+        Place a wall.
```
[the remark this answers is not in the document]
adamantivm: I didn't understand this comment. Kedize

deep_quoridor/src/quoridor_env.py
adamantivm: This is "... for agent in possible_agents"

adamantivm: But I think that's actually wrong ... I need to rewrite that part actually.

deep_quoridor/src/quoridor_env.py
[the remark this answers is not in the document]
alejandromarcu: I was just saying to say in the comment that the walls take 2 spaces

deep_quoridor/src/quoridor_env.py
alejandromarcu: I think this means that each agent is going to observe the elements in the Dict (as opposed to observing the elements in the Dict for each agent, in which case we would be duplicating the walls and the board)

adamantivm: Yes, but after some additional reading I realized this is just a metadata definition, not state variables, so it seems that it doesn't matter.

deep_quoridor/src/quoridor_env.py
```diff
@@ -0,0 +1,170 @@
...
+        self.board_size = 9  # 9x9 grid
+        self.wall_size = self.board_size - 1  # 8x8 grid for walls
+        self.max_walls = 10  # Each player gets 10 walls
+
+        self.possible_agents = ["player_1", "player_2"]
+        self.agent_order = self.possible_agents.copy()
+        self.agent_selection = self.agent_order[0]
+
+        self.observation_spaces = {
+            agent: spaces.Dict(
```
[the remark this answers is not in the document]
adamantivm: done, although uselss for now

deep_quoridor/src/quoridor_env.py
```diff
@@ -0,0 +1,170 @@
...
+        self.reset()
+
+    def reset(self, seed=None, options=None):
+        self.board = np.zeros((self.board_size, self.board_size), dtype=np.int8)
+        self.walls = np.zeros((self.wall_size, self.wall_size, 2), dtype=np.int8)
+        self.walls_remaining = {agent: self.max_walls for agent in self.possible_agents}
+        # Positions are (row, col)
+        self.positions = {"player_1": (4, 0), "player_2": (4, 8)}
+        self.board[self.positions["player_1"]] = 1
+        self.board[self.positions["player_2"]] = 2
```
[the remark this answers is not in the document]
adamantivm: I just removed the duplicaiton for now.

deep_quoridor/src/quoridor_env.py
```diff
@@ -0,0 +1,170 @@
...
+        self.positions[agent] = (row, col)
+        self.board[row, col] = 1 if agent == "player_1" else 2
+
+    def _place_wall(self, agent, action):
+        """
+        Place a wall.
+        In a 9x9 board:
+        - First we subtract 81 to get the wall index
+        - The first 8x8 = 64 values reprsent vertical walls, the last 8x8 = 64 values represent horizontal walls
+        - For vertical walls, we place the wall starting atht upper right corner of the (row, col) cell
```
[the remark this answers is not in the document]
adamantivm: Fixed

adamantivm:
how about now @alejandromarcu ? Ok to merge this and continue on the next PR?

deep_quoridor/src/quoridor_env.py
```diff
@@ -73,17 +74,39 @@ def step(self, action):
...
+        """
+        Takes an action index to action parameters (row, col, movement_type)
+        movement_type = 0 for moving, 1 for horizontal wall placement, 2 for vertical wall placement
+        """
+        if idx < self.board_size**2:
+            return divmod(idx, self.board_size)
+        elif idx >= self.board_size**2 and idx < self.board_size**2 + self.wall_size**2:
+            return divmod(idx - self.board_size**2, self.wall_size)
+        elif idx >= self.board_size**2 + self.wall_size**2 and idx < self.board_size**2 + (self.wall_size**2) * 2:
+            return divmod(idx - self.board_size**2 - self.wall_size**2, self.wall_size)
```
alejandromarcu: The `return`s here will return row, col, you're missing to return the movement_type

adamantivm: It's fixed in the last commit amend. Refresh?

alejandromarcu:
> how about now @alejandromarcu ? Ok to merge this and continue on the next PR?

I just left you a comment about something, I think it would be good to fix it before merging.  Actually, I wanted to bring something up: why are we using the same repository? just to re-use a few util functions? I think it would be cleaner to start a new repository

deep_quoridor/src/quoridor_env.py
```diff
@@ -0,0 +1,203 @@
...
+        else:
+            raise ValueError(f"Invalid action type: {action_type}")
+
+    def action_index_to_params(self, idx) -> Tuple[int, int, int]:
+        """
+        Takes an action index to action parameters (row, col, movement_type)
+        movement_type = 0 for moving, 1 for horizontal wall placement, 2 for vertical wall placement
+        """
+        action_type, _ = divmod(idx - self.board_size**2, self.wall_size**2)
+        action_type += 1
```
alejandromarcu, line 109: This is a very ingenuous way of computing the action type. (BTW, if you're discarding the mod, you can just use integer division with `//`, e.g. `action_type = idx - self.board_size**2 //self.wall_size**2 + 1`
Having said that, I think it would be clearer (and less probabilities of introducing a bug now or later) to go the dumb way and just put `action_type = 1` in the first if branch and so on.
Anyway, feel free to leave it like this if you prefer, just wanted to give you my AR$20 (which is about $0.02)

adamantivm: Ok, I made it more boring as you requested. Now our remaining 999 engineers will be able to maintain this more easily :-P

adamantivm:
Let's discuss about different vs same repo on Monday.

jonbinney:
Thanks for implementing this! I really need to wrap my head around how petting zoo is used and what it can do.

adamantivm:
Note that this is still very incomplete! I also need to figure it out still.
I think they key is to implement an observe method and also some way to calculate and return the valid actions after each move, probably more.

PR #98, opened 2025-03-09
Next iteration and building the Quoridor Env in Petting Zoo

adamantivm:
For now:
 - Implemented a more proper usage of the API in main.py
 - Tweaked the implementation to comply so that it works
 - Builerplate towards implementing observe, reward and mask functions (this will be the key part)
 - Implemented Observe function
 - Partially implemented action mask: moves are done, wall placements missing (next up)

### DIFF
--- a/deep_quoridor/src/quoridor_env.py
+++ b/deep_quoridor/src/quoridor_env.py
@@ -1,6 +1,20 @@
 """
 Quoridor Gym/PettingZoo Environment implementation.
 
+The board size is parameterizable with:
+ - Board size (int): size of the square board
+ - Max number of walls (int): maximum number of walls that each player can place
+Boards are always square.
+For all explanations below, the default value of 9x9 is assumed.
+
+The state is implemented in a two-hot encoding, as follows:
+ - A 9x9 2D array, each representing a space in the board. All values are 0 except
+   1 where player 1 is placed, and 2 where player 2 is placed.
+   NOTE: This is represented internally in a compact way but converted in the specified
+   observation space format as needed.
+ - An 8x8x2 array, representing the vertical and horizontal walls placed in the board, using
+   one-hot encoding.
+
 The action is represented as follows, taking a 9x9 board as an example:
 - First 9x9 = 81 values represent player positions
 - The next 8x8 = 64 values reprsent vertical walls and the last 8x8 = 64 values represent horizontal walls
@@ -11,6 +25,7 @@ The action is represented as follows, taking a 9x9 board as an example:
 
 from typing import Tuple
 from pettingzoo import AECEnv
+from pettingzoo.utils import wrappers
 from gymnasium import spaces
 import numpy as np
 
@@ -21,7 +36,7 @@ class QuoridorEnv(AECEnv):
     def __init__(self, board_size: int = 9, max_walls: int = 10):
         super().__init__()
 
-        self.render_mode = "ansi"
+        self.render_mode = "human"
 
         self.board_size = board_size  # assumed square grid
         self.wall_size = self.board_size - 1  # grid for walls
@@ -201,4 +216,4 @@ class QuoridorEnv(AECEnv):
 
 # Wrapping the environment for PettingZoo compatibility
 def env():
-    return QuoridorEnv()
+    return wrappers.CaptureStdoutWrapper(QuoridorEnv())
--- a/deep_quoridor/src/quoridor_env.py
+++ b/deep_quoridor/src/quoridor_env.py
@@ -236,7 +236,7 @@ class QuoridorEnv(AECEnv):
                 if col < self.wall_size:
                     if row < self.wall_size and self.walls[row, col, 0]:
                         board_str += "|"
-                    elif row > -1 and self.walls[row - 1, col, 0]:  # Continuation of a vertical wall
+                    elif row > 0 and self.walls[row - 1, col, 0]:  # Continuation of a vertical wall
                         board_str += "|"
                     else:
                         board_str += " "
@@ -250,7 +250,7 @@ class QuoridorEnv(AECEnv):
                 for col in range(self.board_size):
                     if col < self.wall_size and self.walls[row, col, 1]:
                         board_str += "──── "  # Wall segment
-                    elif col > -1 and self.walls[row, col - 1, 1]:
+                    elif col > 0 and self.walls[row, col - 1, 1]:
                         board_str += "──── "  # Continuation of a wall from the left
                     else:
                         board_str += "     "
--- a/deep_quoridor/src/quoridor_env.py
+++ b/deep_quoridor/src/quoridor_env.py
@@ -150,6 +150,11 @@ class QuoridorEnv(AECEnv):
                 self.walls[wall_row, min(col_a, self.wall_size - 1), 1] == 1
                 or self.walls[wall_row, max(col_a - 1, 0), 1] == 1
             )
+        else:
+            raise ValueError(f"Invalid movement from {row_a, col_a} to {row_b, col_b}")
+
+    def _is_in_board(self, row, col):
+        return 0 <= row < self.board_size and 0 <= col < self.board_size
 
     def _get_action_mask(self, agent_id):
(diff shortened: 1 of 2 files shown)


alejandromarcu (approved):
I know it's WIP, but if you prefer you can merge it and continue in a new PR. Hopefully it won't break prod 😆

deep_quoridor/src/quoridor_env.py
```diff
@@ -80,12 +85,89 @@ def step(self, action):
...
+                        board_str += "──── "  # Wall segment
+                    elif col > -1 and self.walls[row, col - 1, 1]:
+                        board_str += "──── "  # Continuation of a wall from the left
+                    else:
+                        board_str += "     "
+                board_str += "\n"
+
+        print(board_str)
+
+    @functools.lru_cache(maxsize=None)
```
jonbinney, line 154: lru_cache is new to me - seems very useful! i guess some parent class does some magic to change the hash of "self" whenever the class is modified? Otherwise it would return the same memoized result even if the environment has changed

deep_quoridor/src/quoridor_env.py
```diff
@@ -19,50 +35,39 @@ class QuoridorEnv(AECEnv):
...
-        self.terminations = {agent: False for agent in self.possible_agents}
-        self.truncations = {agent: False for agent in self.possible_agents}
+        self.rewards = {agent: 0 for agent in self.agents}
+        self.terminations = {agent: False for agent in self.agents}
+        self.truncations = {agent: False for agent in self.agents}
         self.agent_selection = self.agent_order[0]
 
+        # I created these so that env.last() wouldn't complain, but I don't
+        # really understand how should they work or why they are needed
+        self._cumulative_rewards = self.rewards.copy()
```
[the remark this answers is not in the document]
adamantivm: I'm not particularly interested in the cummulative reward, but if I don't add this, then `env.last()` fails, so I had to add it.

About the reward, I have some ideas for incremental rewards, worth discussing online.

deep_quoridor/src/quoridor_env.py
adamantivm: I think this can be used because the return here never changes, even as the environment change. This observation space is only a description of the space characteristics, not the space itself.

deep_quoridor/src/quoridor_env.py
```diff
@@ -80,12 +88,161 @@ def step(self, action):
...
+            old_row, old_col = self.positions[agent_id]
+            new_row, new_col = old_row + row, old_col + col
+
+            # Check if the new position is still inside the board
+            if not (0 <= new_row < self.board_size and 0 <= new_col < self.board_size):
+                continue
+
+            # Is there a wall in the way we're going?
+            blocked_by_wall = False
+            if row == 0:  # Moving vertically - check horizontal walls
```
alejandromarcu, line 149: The comment should be the opposite, since the row is not changing, right?
```suggestion
            if row == 0:  # Moving horizontally - check vertical walls
```

deep_quoridor/src/quoridor_env.py
```diff
@@ -80,12 +88,161 @@ def step(self, action):
...
+                wall_col = min(old_col, new_col)
+                # NOTE: The min and max tricks there are just a lazy way to avoid overindexing
+                # Since the wall grid is one smaller than the board (because boards take two spaces)
+                if (
+                    self.walls[min(new_row, self.wall_size - 1), wall_col, 0] == 1
+                    or self.walls[max(new_row - 1, 0), wall_col, 0] == 1
+                ):
+                    blocked_by_wall = True
+            else:  # Moving horizontally - check vertical walls
```
alejandromarcu, line 158:
```suggestion
            else:  # Moving vertically - check horizontal walls
```

alejandromarcu (commented):
I think it will be good to have some unit tests, since handling the movements and walls is a bit tricky and there are a few corner cases with jumping, especially with walls.  I can volunteer to write unit tests if you want

adamantivm:
> I think it will be good to have some unit tests, since handling the movements and walls is a bit tricky and there are a few corner cases with jumping, especially with walls. I can volunteer to write unit tests if you want

yes please the unit tests!

deep_quoridor/src/quoridor_env.py
```diff
@@ -132,6 +132,25 @@ def _get_observation(self, agent_id):
...
+                self.walls[min(row_a, self.wall_size - 1), wall_col, 0] == 1
+                or self.walls[max(row_a - 1, 0), wall_col, 0] == 1
+            )
+        elif col_a == col_b:  # Vertical movement - check horizontal walls
+            wall_row = min(row_a, row_b)
+            return (
+                self.walls[wall_row, min(col_a, self.wall_size - 1), 1] == 1
+                or self.walls[wall_row, max(col_a - 1, 0), 1] == 1
+            )
+
```
alejandromarcu, line 153: I'd raise an exception here, because if by accident (e.g. when checking jumps) we pass both a different row and column, then `None` will be returned and may be evaluated as `False`, and will be hard to debug for our other 999 developers

deep_quoridor/src/quoridor_env.py
```diff
@@ -146,35 +165,48 @@ def _get_action_mask(self, agent_id):
...
-            # Is the opponent in the target position?
-            opponent_in_target = self.positions[self._opponent(agent_id)] == (new_row, new_col)
+            # Check if that direction is blocked by a wall
+            if self._is_wall_between(old_row, old_col, new_row, new_col):
+                continue
 
-            # TODO: Jumping the opponent, considering walls if there are any
+            # Check if the opponent is in the target position
+            if self.positions[self._opponent(agent_id)] == (new_row, new_col):
+                # Can we jump straight over the component?
```
alejandromarcu, line 174: I assume you mean the opponent?

```suggestion
                # Can we jump straight over the opponent?
```

deep_quoridor/src/quoridor_env.py
```diff
@@ -146,35 +165,48 @@ def _get_action_mask(self, agent_id):
...
+                straight_row = new_row + row
+                straight_col = new_col + col
 
-            # Check if the aren't any walls in the way
-            if not opponent_in_target and not blocked_by_wall:
-                mask[self.action_params_to_index(new_row, new_col, 0)] = 1
+                if (
+                    # That new position falls within the board
+                    0 <= straight_row < self.board_size
+                    and 0 <= straight_col < self.board_size
```
alejandromarcu, line 181: What do you think of making a function, e..g `_in_board` for this? it's short but will be used often

deep_quoridor/src/quoridor_env.py
alejandromarcu: Also, we could rise if the difference between rows or cols is other than 1 (e.g. if we accidentally call it with a jump of 2)

deep_quoridor/src/quoridor_env.py
```diff
@@ -80,22 +90,206 @@ def step(self, action):
...
+                elif self.positions["player_1"] == (row, col):
+                    board_str += " P1 "
+                else:
+                    board_str += " .  "
+
+                # Render vertical walls (|)
+                if col < self.wall_size:
+                    if row < self.wall_size and self.walls[row, col, 0]:
+                        board_str += "|"
+                    elif row > -1 and self.walls[row - 1, col, 0]:  # Continuation of a vertical wall
```
alejandromarcu, line 239: I'm not sure why you changed to `row > -1` from `row > 0`, but I think it should be `row > 0`, otherwise if it's 0, we're getting the last row.  E.g. I run into this when testing dfs, I was placing a vertical wall on the bottom and it was overflowing to the top because of this

adamantivm: I agree. I don't know why this ended up changing. I don't even remember adding the check in the first place!
I agree it needs to be row > 0 but it's for a very subtle reason! It's because otherwise row - 1 can evaluate to -1, which in python means "the last one in the array", not what we want! 
I'm fixing it now.

PR #100, opened 2025-03-11
FInished environment - random agent now possible

adamantivm:
This MR completes a first pass of the agent implementation.
Now the game runs until one of the agents win, taking action moves.
WARNING: It's very frustrating to watch.

I also did a small update to the rendering to make it a tad easier to see the walls.

### DIFF
--- a/deep_quoridor/src/main.py
+++ b/deep_quoridor/src/main.py
@@ -9,42 +9,19 @@ print("Initial Board State:")
 board = game.render()
 print(board)
 
-# Manually select moves (modify these to test different scenarios)
-actions = [
-    # 0- Move forward
-    game.action_params_to_index(1, 4, 0),
-    # 1- Move forward (opponent)
-    game.action_params_to_index(7, 4, 0),
-    # 0- Add a Horizontal wall
-    game.action_params_to_index(6, 4, 2),
-    # 1- Add a vertical wall
-    game.action_params_to_index(0, 3, 1),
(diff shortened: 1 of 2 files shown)


deep_quoridor/src/main.py
```diff
@@ -9,42 +9,20 @@
...
     if termination or truncation:
         action = None
-        print(f"\nGame Over! {agent} wins.")
+        winner = max(game.rewards, key=game.rewards.get)
+        print(f"\nGame Over! {winner} wins.")
+        break
     else:
         # Hardcoded actions for now
-        action = actions.pop(0)
+        action = game.action_space(agent).sample(mask)
```
adamantivm, line 25: We have a baseline agent!

alejandromarcu (approved):
Nice!

deep_quoridor/src/main.py
jonbinney: Is `.sample()` uniform random?

adamantivm: Apparently yes in this case: https://gymnasium.farama.org/api/spaces/#gymnasium.spaces.Space.sample

alejandromarcu: We should write a better heuristic, e.g. follow the shortest path to the target if our path is shorter, or place a wall in the path of the shorter path of the adversary otherwise

jonbinney: I'll try that tonight

PR #102, opened 2025-03-12
Unit tests for wall placement

alejandromarcu:
Add tests to verify that walls can't be placed overlapping other walls or blocking the pass (and also test that walls can be placed otherwise).
With the tests I found a small bug: it would allow a new wall to be placed exactly on top of an existing wall.  I changed the way it's verified in a way that I find easier to understand

### DIFF
--- a/deep_quoridor/src/quoridor_env.py
+++ b/deep_quoridor/src/quoridor_env.py
@@ -232,15 +232,13 @@ class QuoridorEnv(AECEnv):
         if self.walls[row, col, (orientation + 1) % 2] == 1:
             return True
 
-        # Check overlap with a wall in the same orientation before or after
-        deltarow = (orientation + 1) % 2
-        deltacol = orientation
-        if (
-            self.walls[min(row + deltarow, self.wall_size - 1), min(col + deltacol, self.wall_size - 1), orientation]
-            == 1
-            or self.walls[max(row - deltarow, 0), max(col - deltacol, 0), orientation] == 1
-        ):
-            return True
+        # Check that in the 2 wall segments there's no other segment already
+        if orientation == 0:
+            if self.is_wall_between(row, col, row, col + 1) or self.is_wall_between(row + 1, col, row + 1, col + 1):
+                return True
+        else:
+            if self.is_wall_between(row, col, row + 1, col) or self.is_wall_between(row, col + 1, row + 1, col + 1):
+                return True
 
         return False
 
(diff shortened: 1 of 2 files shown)


deep_quoridor/src/quoridor_env.py
```diff
@@ -232,15 +232,13 @@ def _is_wall_overlap(self, row, col, orientation):
...
             return True
 
-        # Check overlap with a wall in the same orientation before or after
-        deltarow = (orientation + 1) % 2
-        deltacol = orientation
-        if (
-            self.walls[min(row + deltarow, self.wall_size - 1), min(col + deltacol, self.wall_size - 1), orientation]
-            == 1
-            or self.walls[max(row - deltarow, 0), max(col - deltacol, 0), orientation] == 1
-        ):
```
alejandromarcu, line 242: it was missing `or self.walls[row, col, orientation] == 1`, but I find this other way easier to reason about (basically check if in the 2 places where we would put segments there's a wall already, which also doesn't need to check for boundaries).

deep_quoridor/src/quoridor_env.py
```diff
@@ -328,7 +326,7 @@ def action_params_to_index(self, row: int, col: int, action_type: int = 0) -> in
     def action_index_to_params(self, idx) -> Tuple[int, int, int]:
         """
         Takes an action index to action parameters (row, col, movement_type)
-        movement_type = 0 for moving, 1 for horizontal wall placement, 2 for vertical wall placement
+        movement_type = 0 for moving, 1 for vertical wall placement, 2 for horizontal wall placement
```
alejandromarcu, line 329: I found a comment bug that confused me a lot

alejandromarcu:
BTW, great job in the implementation of the Env, @adamantivm! you get a reward of +10 points for completing it and -1 point for the bug 🤣

deep_quoridor/src/quoridor_env.py
adamantivm: Off the top of my head, I think this is incomplete, because it doesn't contemplate the case where you have a perpendicular wall going through in the middle of the two spaces. 
Let me see if I can prove this with a unit test, when I have a chance. In that case, you'll get -1 point :)

deep_quoridor/test/quoridor_env_test.py
```diff
@@ -13,34 +13,38 @@ def parse_board(board):
...
     - '*': Potential move positions for player 1
     - '|': Vertical walls
+    - 'v': A vertical wall cannot be placed here (for testing wall placement)
     - ' ': Empty spaces (just for formatting)
     - '+': Vertical wall continuation (just for formatting)
 
     Additionally, between 2 rows for the cells, a row with horizontal walls can be included.
     The position of the wall will be with respect to the row above.  The row is composed by:
     - '-': Horizontal walls
+    - 'v': A horizontal wall cannot be placed here (for testing wall placement)
```
adamantivm, line 23:
```suggestion
    - '>': A horizontal wall cannot be placed here (for testing wall placement)
```

Is this correct?

adamantivm (approved):
Two nits and an erronous observation that I deleted, then done!

PR #103, opened 2025-03-12
Animate game in text mode

alejandromarcu:
I thought it would be nice to see the game animated, and if it's only "AI" playing we don't need a fancy UI, so I did an animation with text which was very easy and fast to do with copilot ;)

### DIFF
--- a/deep_quoridor/src/main.py
+++ b/deep_quoridor/src/main.py
@@ -1,3 +1,5 @@
+import curses
+
 from quoridor_env import env  # Import the environment from your file
 
 # Initialize environment
@@ -10,6 +12,11 @@ board = game.render()
 print(board)
 
 step = 0
+stdscr = curses.initscr()
+curses.noecho()
+curses.cbreak()
+curses.start_color()
+
 for agent in game.agent_iter():
     observation, reward, termination, truncation, info = game.last()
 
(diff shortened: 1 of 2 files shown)


adamantivm:
if you have it handy, can you include a small screen recording of this playing?

alejandromarcu:
Here's the recorder.  Once we have an AI we can make it a bit slower.

https://github.com/user-attachments/assets/20e48aa3-c94d-4181-bd26-0a8307b0d2ed

adamantivm:
Nice! 
I have only one request: can you make it optional or at least allow a runtime argument to disable it and just print to screen, just in case?
Other than that, LGTM

jonbinney:
This is great, thanks @alejandromarcu !

alejandromarcu:
Done, as a bonus I added options to change the board size and max walls.  In the future when we include GUI we may want to do something more fancy with how we choose rendering, but I think this is enough for now

deep_quoridor/src/main.py
```diff
@@ -1,35 +1,71 @@
...
+            action = game.action_space(agent).sample(mask)
+            print(f"\nStep {step + 1}: {agent} takes action {action}")
+
+        game.step(action)  # Apply action
+
+        board = game.render()
+
+        if render == "curses":
+            stdscr.clear()
+            stdscr.addstr(2, 2, board)
```
alejandromarcu, line 48: oops, this should be 2,0,board, otherwise the first row is to the right (I wanted to move the board away from the side).  I'll piggy back the change in my next PR

PR #104, opened 2025-03-13
Split simple agents into classes

jonbinney:
The RandomAgent in main.py moves completely randomly. The SimpleAgent will do some lookahead but still has bugs that need to be worked out.

### DIFF
--- a/deep_quoridor/src/main.py
+++ b/deep_quoridor/src/main.py
@@ -1,7 +1,20 @@
-import curses
 import argparse
+import curses
 
 from quoridor_env import env  # Import the environment from your file
+from simple_agent import SimpleAgent
+
+
+class RandomAgent:
+    def __init__(self):
+        pass
+
+    def get_action(self, game):
+        observation, _, termination, truncation, _ = game.last()
+        mask = observation["action_mask"]
+        if termination or truncation:
+            return None
+        return game.action_space(game.agent_selection).sample(mask)
 
 
 def play(board_size: int | None, max_walls: int | None, render: str):
@@ -11,6 +24,11 @@ def play(board_size: int | None, max_walls: int | None, render: str):
     game = env(**args)
 
     game.reset()
+
+    agents = {
+        "player_0": RandomAgent(),
+        "player_1": RandomAgent(),
+    }
 
     if render == "print":
         print("Initial Board State:")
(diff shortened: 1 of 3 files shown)


alejandromarcu:
If I understand correctly, the simple agent is trying 10 different random sequences of 3 movements and picking up the best based on the rewards, no?
I see a few issues with this approach:
- Usually you can put 100+ walls and have 4 movements, so when chosing randomly, it's going to be rare that it choses the movement, it will chose 25 times to add a wall per 1 time it choses to move
- Even when they move, they move randomly, so less chances of reaching the target
- 3 moves is really short, so per the things above, only by the end of the game any of the sequences will have a reward
- Rewards are not discounted, so it may pick up a much longer path than available.

I think you would need discounted rewards (or it can just be dividing the reward by the length of the path), longer sequences and more sequences to try.  But I still think it won't be very good.  E.g. let's say that the other player is random, the strategy would be to move forward 9 times.  But if you're chosing completely random, the odds that the first move would be to move forward are very slim (until you run out of walls). 
If you made it for example that with probability p (e.g. 0.6) it would move vs putting a wall, I think you would have a significant improvement.  Or even more, make it such that with a probability p it moves on the shortest path.  Notice that this is just an heuristic for exploring, because if your exploration is too bad, because of the branching, you would need to explore a huge space.

adamantivm:
Regardless of the critics to the algorithm, I think the refactoring is the right way to go. Any reason not to merge this? Let's merge it and continue from here!

alejandromarcu:
Sounds good, we can merge and iterate on the agent

jonbinney:
Sounds good, I'll merge and make some fixes to the SimpleAgent today. How do we feel about the convention of having agents be classes with a `get_action(env)` method?

jonbinney:
I think that strictly speaking the agent should only have access to `observation`  but not the entire game state. But probably doesn't matter since this is a complete information game.

alejandromarcu:
> I think that strictly speaking the agent should only have access to `observation` but not the entire game state. But probably doesn't matter since this is a complete information game.

I think it's ok, I see that you needed the game to copy it and explore.  We could re-build the game from the observation in theory, but seems like a lot of extra work and inefficient

adamantivm:
> How do we feel about the convention of having agents be classes with a `get_action(env)` method?

I like it!

PR #105, opened 2025-03-13
Optimize wall block check

alejandromarcu:
The `observe` method needs to check for all the possible wall placements (128) whether the pawn can reach the other side, which was done using DFS, and all this was relatively slow.
E.g using the profiling in this PR and adding `return True` at the beginning of `_is_wall_potential_block`:

```
         1707939 function calls (1460111 primitive calls) in 0.395 seconds

   Ordered by: internal time

251838/4010    0.135    0.000    0.381    0.000 quoridor_env.py:437(dfs)
   286178    0.133    0.000    0.226    0.000 quoridor_env.py:135(is_wall_between)
   572356    0.062    0.000    0.062    0.000 {built-in method builtins.min}
   268422    0.031    0.000    0.031    0.000 {built-in method builtins.max}
   307550    0.024    0.000    0.024    0.000 quoridor_env.py:156(_is_in_board)
```

With this optimization we get:

```
         205243 function calls (180855 primitive calls) in 0.050 seconds

   Ordered by: internal time

    32292    0.015    0.000    0.026    0.000 quoridor_env.py:135(is_wall_between)
24734/346    0.013    0.000    0.038    0.000 quoridor_env.py:436(dfs)
    68632    0.007    0.000    0.007    0.000 {built-in method builtins.min}
    34382    0.004    0.000    0.004    0.000 {built-in method builtins.max}
```

Which is 8x faster, since now we call the dfs function 1/10th of the times.
For further optimization, we could:
- optimize `is_wall_between`, e.g. by keeping in memory a bitmap of the walls including padding, so the check would be as easy as checking for 1 array position
- Look for another algorithm instead of DFS, e.g. an algorithm that we can run it once and then check for all the wall positions
- Look for a different implementation (E.g. https://scikit-image.org/docs/0.23.x/auto_examples/segmentation/plot_floodfill.html)
- Use caching. For example, hash the wall configuration to be used as a key, and as a value we would store all the cells that we can flow from each target.  Then, if the wall configuration doesn't change (e.g. players moved instead of putting a wall), we can quickly check if the pawn is in one of the reachable cells.

### DIFF
--- a/deep_quoridor/test/profile_observe.py
+++ b/deep_quoridor/test/profile_observe.py
@@ -1,0 +1,40 @@
+from deep_quoridor.src.quoridor_env import env
+import cProfile
+import time
+
+e = env(9)
+walls = [
+    ((3, 0), 1),
+    ((3, 2), 1),
+    ((3, 4), 1),
+    ((3, 6), 1),
+    ((2, 7), 0),
+    ((5, 7), 1),
+    ((4, 0), 0),
+    ((4, 2), 0),
+    ((4, 4), 0),
+]
+
+
+def run():
+    st = time.time()
+    for w in walls:
+        e.place_wall("player_0", w[0], w[1])
+        e.observe("player_0")
+        e.observe("player_1")
+
+    et = time.time()
+    print(e.render())
+    print((et - st) * 1000000)
+
+
+# Run in Cucu's computer.  Just for relative reference:
+#
+# No checks at all:                                90 usec
+# Wall overlap check:                           3,000 usec
+# Wall overlap and block:                     200,000 usec
+# Using _is_wall_potential_block optimization: 25,000 usec
+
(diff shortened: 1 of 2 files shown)


deep_quoridor/test/profile_observe.py
```diff
@@ -0,0 +1,40 @@
...
+    print(e.render())
+    print((et - st) * 1000000)
+
+
+# Run in Cucu's computer.  Just for relative reference:
+#
+# No checks at all:                                90 usec
+# Wall overlap check:                           3,000 usec
+# Wall overlap and block:                     200,000 usec
+# Using _is_wall_potential_block optimization: 25,000 usec
```
jonbinney, line 36: Thanks for including these times here.... very useful for getting a rough idea of what helped.

jonbinney (approved):
LGTM!

PR #106, opened 2025-03-15
Optional Incremental rewards

adamantivm:
Implements an option to enable incremental rewards, to accelerate training albeit suboptimally.

~~I still want to test it a bit more, but~~ it's ready for review.

Thanks for the unit tests and the DFS @alejandromarcu , it came very handy

Some clarifications:
 - I changed the default render mode to print, because ncurses once you run it, it ruins the terminal and it's annoying to get back to it (I use it a lot to re-run on the console)
 - The new DFS param `any_path` is meant so that if the distance is not needed, the amount of comparations remains the same as before when it only returned True / False
 - For the incremental step value, I did some runs and it looked like when a player wins by a large margin, the rewawrd right before winning is still lower than 1 (but getting close to 1)

### DIFF
--- a/deep_quoridor/src/quoridor_env.py
+++ b/deep_quoridor/src/quoridor_env.py
@@ -37,7 +37,7 @@ import numpy as np
 class QuoridorEnv(AECEnv):
     metadata = {"render_modes": ["ansi"], "name": "quoridor_v0"}
 
-    def __init__(self, board_size: int, max_walls: int, step_rewards: bool = False):
+    def __init__(self, board_size: int, max_walls: int, step_rewards: bool):
         super(AECEnv, self).__init__()
 
         self.render_mode = "human"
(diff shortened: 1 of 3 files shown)


alejandromarcu:
DFS doesn't give you the shortest path, you need BFS.  I'm happy to write the bfs function for you if you want.

jonbinney:
@alejandromarcu I'm working on a PR to use the networkx graph library to speed things up and make it easy to check things like distance between parts of the board - I should have that ready tomorrow and then we can use A* or djikstra or whatever we want.

alejandromarcu:
Actually I just realized that Julian changed DFS to get the shortest path

alejandromarcu:
I'm not convinced about this way of finding the shortest path, because we're marking the cells as visited but never unmarking them.  So if the first path to reach is not the shortest but uses some cells that we need to visit the shortest path, we won't be able to travel through the shortest path. If we un-mark them, then it will take the same path over and over.
Let's see if we can use the PR that @jonbinney is writing for the shortest path!

adamantivm:
I'm not too concerned with the optimality of the solution, as long as it more or less it approximates some "betterness" of moves.
Remember this is both optional - and a good RL implementation should disable this feature - and only meant to be able to play with something that beats random.
Is it OK if I merge this, as long as it doesn't break what is already there? Could you review it with that aspect in mind @alejandromarcu ?

deep_quoridor/src/quoridor_env.py
```diff
@@ -101,6 +102,15 @@ def step(self, action):
             self.terminations = {a: True for a in self.agents}
             self.rewards[agent] = 1
             self.rewards[self.get_opponent(agent)] = -1
+        elif self.step_rewards:
+            # Assign rewards as the difference in distance to the goal divided by
+            # twice the board size.
```
alejandromarcu, line 107:
```suggestion
            # 3 times the board size.
```

deep_quoridor/src/quoridor_env.py
```diff
@@ -434,50 +440,78 @@ def _next_player(self):
...
+                dfs = self._dfs(new_row, new_col, target_row, visited)
+                if dfs != -1:
+                    if any_path:
+                        return dfs + 1
+                    if best == -1 or dfs + 1 < best:
+                        best = dfs + 1
+
+        return best
+
+    def can_reach(self, row, col, target_row, any_path=True):
```
alejandromarcu, line 487: now that it returns an int rather than a bool, can_reach is not a good name.  I think similarly to Euclidean Distance or Hamiltonean Distance, this should be named Julianian Distance 🤣 
You could leave the `can_reach` method and make it return boolean, and add another method julianian_distance

alejandromarcu (approved):

PR #107, opened 2025-03-17
WIP: Move core Quoridor implementation into new class

jonbinney:
Not done yet, but I think this is the right direction. I'll try to finish tomorrow. This PR does a couple things:
- use a 3rd party graph library to speed up detection of invalid walls
- split core quoridor rules into their own class that agents can use (so they don't need the entire environment)

I also need to fix merge conflicts with @adamantivm 's last PR.

### DIFF
--- a/deep_quoridor/src/quoridor.py
+++ b/deep_quoridor/src/quoridor.py
@@ -1,0 +1,188 @@
+from dataclasses import dataclass
+from enum import IntEnum, unique
+
+import networkx as nx
+import numpy as np
+
+
+@unique
+class WallOrientation(IntEnum):
+    VERTICAL = 0
+    HORIZONTAL = 1
+
+
+class Action:
+    pass
+
+
+@dataclass
+class MoveAction(Action):
+    destination: tuple[int, int]  # (row, col)
+
+
+@dataclass
+class WallAction(Action):
+    position: tuple[int, int]  # Start of wall (row, col)
+    orientation: WallOrientation
+
+
+def create_board_graph(board_size: int):
+    """
+    Create a graph representation of the board using networkx library.
+    Each node represents a position on the board and each edge represents a possible move.
+    """
+
+    graph = nx.Graph()
+    for row in range(board_size):
+        for col in range(board_size):
+            # Add node for each position on the board
+            graph.add_node((row, col))
+
+            # Add edges for possible moves
+            if row < board_size - 1:
+                graph.add_edge((row, col), (row + 1, col))  # Down
+            if row > 0:
+                graph.add_edge((row, col), (row - 1, col))  # Up
+            if col < board_size - 1:
+                graph.add_edge((row, col), (row, col + 1))  # Right
+            if col > 0:
+                graph.add_edge((row, col), (row, col - 1))  # Left
+
+    return graph
+
+
+class CombinedGrid:
+    """
+    Grid which has both walls and possible pawn positions.
+
+    Pawn positions are on even rows and columns, walls are on odd rows and columns.  Occupied pawn and
+    wall positions are marked as True.
+    """
+
+    def __init__(self, board_size: int):
+        self._grid = np.zeros((board_size * 2 - 1, board_size * 2 - 1), dtype=np.bool)
+
+    def set_player_cell(self, position: tuple[int, int], value: bool):
+        self._grid[position[0] * 2, position[1] * 2] = value
+
+    def get_player_cell(self, position: tuple[int, int]) -> bool:
+        self._grid[position[0] * 2, position[1] * 2]
+
+    def set_wall_cells(self, position: tuple[int, int], orientation: WallOrientation, value: bool):
+        if orientation == WallOrientation.VERTICAL:
+            self._grid[position[0] * 2 + 1 : position[0] * 2 + 4, position[1] * 2 + 1] = value
+        else:
+            self._grid[position[0] * 2 + 1, position[1] * 2 + 1 : position[1] * 2 + 4] = value
+
+    def get_wall_cells(self, position: tuple[int, int], orientation: WallOrientation) -> bool:
+        if orientation == WallOrientation.VERTICAL:
+            return self._grid[position[0] * 2 + 1 : position[0] * 2 + 4, position[1] * 2 + 1]
+        else:
+            return self._grid[position[0] * 2 + 1, position[1] * 2 + 1 : position[1] * 2 + 4]
+
+    def get_walls_between(self, position1: tuple[int, int], position2: tuple[int, int]) -> bool:
+        """
+        Return a sequence of occupancy values (both wall and player positions) between two player positions
+
+        The first element of the sequence will be player, then wall, etc.
+        The last element of the sequence will be the player position at position2.
+        """
+        start = (position1[0] * 2, position1[1] * 2)
+        destination = (position2[0] * 2, position2[1] * 2)
+        return self._grid[start[0] : destination[0] + 1, start[1] : destination[1] + 1]
+
+    def __str__(self):
+        return "\n".join("".join("X" if cell else "." for cell in row) for row in self._grid)
+
+
+class Quoridor:
+    def __init__(self, board_size: int, max_walls: int):
+        self.board_size = board_size
+        self.max_walls = max_walls
+        self.walls = np.zeros((board_size - 1, board_size - 1, 2), dtype=np.int8)
+        self.player_positions = [(0, board_size // 2), (board_size - 1, board_size // 2)]
+        self.walls_remaining = [self.max_walls, self.max_walls]
+        self.current_player = 0
+
+        # Occupancy grid which has both walls and possible pawn positions.
+        # Occupied pawn and wall positions are marked as True.
+        self._combined_occ_grid = CombinedGrid(board_size)
+
+    def step(self, action: Action, validate: bool = True):
+        """
+        Execute the given action and update the game state.
+
+        Turning off validation can save some computation if you're sure the action is valid,
+        and also makes it easier to teleport players around the board for testing purposes.
+        """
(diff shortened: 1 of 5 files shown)


deep_quoridor/src/quoridor.py
```diff
@@ -0,0 +1,188 @@
...
+        and also makes it easier to teleport players around the board for testing purposes.
+        """
+        if validate and not self.is_action_valid(action):
+            raise ValueError("Invalid action")
+
+        if isinstance(action, MoveAction):
+            # Move the pawn, mark previous position as empty, and mark new position as occupied.
+            self._combined_occ_grid.set_player_cell(self.player_positions[self.current_player], False)
+            self.player_positions[self.current_player] = action.destination
+            self._combined_occ_grid.set_player_cell(self.player_positions[self.current_player], True)
```
alejandromarcu, line 125: This hopefully would be encapsulated in CombinedGrid and be just 1 function call

deep_quoridor/src/quoridor.py
```diff
@@ -0,0 +1,188 @@
...
+            )
+
+            # This is a valid normal one space move. Just make sure there are no walls in the way.
+            if np.abs(delta).sum() == 1:
+                return True
+
+            # Straight jump. The must be an opponent in between, and no wall.
+            if delta in [(2, 0), (-2, 0), (0, 2), (0, -2)]:
+                middle_position = np.mean([action.destination, self.player_positions[self.current_player]])
+                if not self._combined_occ_grid.get_player_cell((action.destination[0] - 1, action.destination[1])):
```
alejandromarcu, line 156: middle_position was meant to be used here, no?

deep_quoridor/src/quoridor.py
```diff
@@ -0,0 +1,188 @@
...
+                    return False
+
+                path_walls = self._combined_occ_grid.get_walls_between(
+                    self.player_positions[self.current_player], action.destination
+                )
+                if path_walls.any():
+                    # There is a wall in between. Can't jump.
+                    return False
+
```
alejandromarcu, line 165: we still need to check for non-straight jumps

deep_quoridor/src/quoridor.py
```diff
@@ -0,0 +1,188 @@
...
+        elif isinstance(action, WallAction):
+            if self.walls_remaining[self.current_player] < 1:
+                return False
+
```
alejandromarcu, line 169: and here we need to check that the wall is not blocking the path to the target

deep_quoridor/test/quoridor_env_test.py
```diff
@@ -19,21 +19,21 @@ def parse_board(board):
 
     Additionally, between 2 rows for the cells, a row with horizontal walls can be included.
     The position of the wall will be with respect to the row above.  The row is composed by:
-    - '-': Horizontal walls
+    - '-': Horizontal walls====================================================================
```
alejandromarcu, line 22:
```suggestion
    - '-': Horizontal walls
```

deep_quoridor/test/quoridor_env_test.py
```diff
@@ -68,16 +68,16 @@ def parse_board(board):
...
-                        env.place_wall("player_0", (row_n, col_n - 1), 0)
+                    if ch == "|":
+                        game.step(WallAction((row_n, col_n - 1), WallOrientation.VERTICAL))
                     continue
 
                 if ch == "*":
                     potential_moves.append((row_n, col_n))
                 elif ch == "1":
-                    env.positions["player_0"] = row_n, col_n
+                    game.player_positions[0] = (row_n, col_n)
```
alejandromarcu, line 78: here it's not setting the player position on the CombinedGrid, that's why I want it to be encapuslated :)

deep_quoridor/src/quoridor.py
```diff
@@ -0,0 +1,188 @@
...
+        else:
+            self._grid[position[0] * 2 + 1, position[1] * 2 + 1 : position[1] * 2 + 4] = value
+
+    def get_wall_cells(self, position: tuple[int, int], orientation: WallOrientation) -> bool:
+        if orientation == WallOrientation.VERTICAL:
+            return self._grid[position[0] * 2 + 1 : position[0] * 2 + 4, position[1] * 2 + 1]
+        else:
+            return self._grid[position[0] * 2 + 1, position[1] * 2 + 1 : position[1] * 2 + 4]
+
+    def get_walls_between(self, position1: tuple[int, int], position2: tuple[int, int]) -> bool:
```
alejandromarcu, line 83: also return type should be `array[bool]`

alejandromarcu (commented):
I still need to look more into it and better understand the design implications, but I left some comments for now.

deep_quoridor/src/quoridor_env.py
```diff
@@ -134,36 +124,9 @@ def _get_observation(self, agent_id):
         walls = self.walls.copy()
 
         return {
-            "board": board,
-            "walls": walls,
-            "my_walls_remaining": self.walls_remaining[agent_id],
-            "opponent_walls_remaining": self.walls_remaining[self.get_opponent(agent_id)],
+            "game": self._game,
```
alejandromarcu, line 127: I'm not sure about the consequences here of returning the game, because PettingZoo expects us to return based on the format in `observation_space` method, that's usually things like `spaces.Discrete(...)`

deep_quoridor/src/quoridor_env.py
```diff
@@ -297,45 +260,7 @@ def _get_info(self):
...
-                    elif col > 0 and self.walls[row, col - 1, 1]:
-                        board_str += "──── "  # Continuation of a wall from the left
-                    elif col < self.wall_size and self.walls[row, col, 0]:
-                        board_str += "    +"  # Vertical wall going through this intersection
-                    else:
-                        board_str += "     "
-                board_str += "\n"
-
-        print(board_str)
+        print(self._game.render_to_string())
```
alejandromarcu, line 263: I don't see this function defined

deep_quoridor/test/quoridor_env_test.py
```diff
@@ -68,16 +68,16 @@ def parse_board(board):
 
                 if ch == "v" or ch == "|":
                     forbidden_walls.append((row_n, col_n - 1, 0))
-                    if ch == "|" and not env.is_wall_between(row_n, col_n - 1, row_n, col_n):
-                        env.place_wall("player_0", (row_n, col_n - 1), 0)
+                    if ch == "|":
+                        game.step(WallAction((row_n, col_n - 1), WallOrientation.VERTICAL))
```
alejandromarcu, line 72: I'm not sure if this would work as intendeed, because this was a little hack so that the wall is only placed on the top/left part.  If you don't check, you may place another wall that will extend it. E.g., when we find --, we need to place only one wall starting on the left, if you don't check if there's a wall already, you'll place another wall and will be ---

deep_quoridor/src/quoridor_env.py
```diff
@@ -403,22 +328,6 @@ def _move(self, agent, position):
             return  # Invalid move (occupied)
         self.positions[agent] = (row, col)
 
-    def place_wall(self, agent, position, orientation):
```
alejandromarcu, line 406: the step method is still calling this

deep_quoridor/src/quoridor.py
```diff
@@ -0,0 +1,188 @@
...
+        elif isinstance(action, WallAction):
+            self._combined_occ_grid.set_wall_cells(action.position, action.orientation, True)
+            # TODO: Check that the wall doesn't block any player from reaching their goal.
+            self.walls_remaining[self.current_player] -= 1
+        else:
+            raise ValueError("Invalid action type")
+
+        self.current_player = 1 - self.current_player
+
+    def is_action_valid(self, action: Action):
```
alejandromarcu, line 135: This is duplicating the checks in `_get_action_mask`, since they are complex they should be coded only once.
`_get_action_mask` could call on this for each possibility, but the pawn movement is more straightforward to code the other way around, like it's now in `_get_action_mask`.

jonbinney:
@alejandromarcu the comments are useful but this isn't done, as you've noticed. :-)  Lets chat about the design decisions on the call tomorrow.

deep_quoridor/src/quoridor.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def __str__(self):
+        return "\n".join("".join("X" if cell else "." for cell in row) for row in self._grid)
+
+
+class Quoridor:
+    def __init__(self, board_size: int, max_walls: int):
+        self.board_size = board_size
+        self.max_walls = max_walls
+        self.walls = np.zeros((board_size - 1, board_size - 1, 2), dtype=np.int8)
```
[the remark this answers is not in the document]
jonbinney: Nope! I'm in the process of switching from the old representation to the CombinedGrid; still need to finish the transition.

deep_quoridor/test/quoridor_env_test.py
jonbinney: Looks like i fell asleep on my keyboard

deep_quoridor/src/quoridor_env.py
jonbinney: Luckily the tests fail long before this point so i never notice :-P

deep_quoridor/src/quoridor.py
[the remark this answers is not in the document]
alejandromarcu: I'm sorry, I should have reviewed at a higher level since it's WIP

deep_quoridor/test/quoridor_env_test.py
alejandromarcu: At least you're not putting fake `x`s

jonbinney:
Closing this in favor of https://github.com/jonbinney/deep_rabbit_hole/pull/126

PR #108, opened 2025-03-18
Introduce the Arena with plug ins for saving and rendering

alejandromarcu:
The Arena class is where we would run the games.  The plug-ins allow you to add extra functionality. I converted the renderers to plug ins, so we can easily add or change the renderers without polluting the main loop.
I wrote anoter plug in that saves the actions from one or more games to a YAML file.  And hopefully we can re-use plug ins for more stuff
For future work:
- Tournaments where each agent plays against the rest  (or subset) N times 
- Replay a game by passing the name of the yaml file plus the game name inside

E.g. of the first lines of `game.yaml` (I know, problems with player_0 vs player1, we need to fix this).
```
game1:
  player1: RandomAgent
  player2: SimpleAgent
  winner: player_0
  steps: 553
  time_ms: 60903
  actions:
  - 178
  - 83
  - 166
  ```

### DIFF
--- a/deep_quoridor/src/arena.py
+++ b/deep_quoridor/src/arena.py
@@ -1,13 +1,14 @@
 from typing import Optional
 from quoridor_env import env
-
-
-class Agent:
-    def get_action(self, game):
-        raise NotImplementedError("You must implement the get_action method")
+from agents import Agent
 
 
 class ArenaPlugin:
+    """
+    Base class for all arena plugins.
+    The plug in can override any combinantion of the methods below in order to provide additional functionality.
+    """
+
     def start_game(self, game, agent1: Optional[Agent] = None, agent2: Optional[Agent] = None):
         pass
 
@@ -25,7 +26,15 @@ class ArenaPlugin:
 
 
 class CompositeArenaPlugin:
+    """
+    Allows to combine multiple plugins into a single one, calling them sequentially for each method.
+
+    """
+
     def __init__(self, plugins: list[ArenaPlugin]):
+        """
+        For the sake of convenience, the plugin list is allowed to be empty, in which case the plugin will be a no-op.
+        """
         self.plugins = plugins
 
     def start_game(self, game, agent1: Optional[Agent] = None, agent2: Optional[Agent] = None):
--- a/deep_quoridor/src/arena.py
+++ b/deep_quoridor/src/arena.py
@@ -1,0 +1,91 @@
+from typing import Optional
+from quoridor_env import env
+
+
+class Agent:
+    def get_action(self, game):
+        raise NotImplementedError("You must implement the get_action method")
+
+
+class ArenaPlugin:
+    def start_game(self, game, agent1: Optional[Agent] = None, agent2: Optional[Agent] = None):
+        pass
+
+    def end_game(self, game, step):
+        pass
+
+    def start_arena(self, game):
+        pass
+
+    def end_arena(self, game):
+        pass
+
+    def action(self, game, step, agent, action):
+        pass
+
+
+class CompositeArenaPlugin:
+    def __init__(self, plugins: list[ArenaPlugin]):
+        self.plugins = plugins
+
+    def start_game(self, game, agent1: Optional[Agent] = None, agent2: Optional[Agent] = None):
+        [plugin.start_game(game, agent1, agent2) for plugin in self.plugins]
+
+    def end_game(self, game, step):
+        [plugin.end_game(game, step) for plugin in self.plugins]
+
+    def start_arena(self, game):
+        [plugin.start_arena(game) for plugin in self.plugins]
+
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: deep_quoridor/src/arena.py [hunk 1]
```diff
@@ -0,0 +1,91 @@
...
+
+    def start_arena(self, game):
+        pass
+
+    def end_arena(self, game):
+        pass
+
+    def action(self, game, step, agent, action):
+        pass
+
```
alejandromarcu, line 25: TO DO: Type hints

deep_quoridor/src/main.py
```diff
@@ -1,87 +1,46 @@
...
     parser.add_argument("-N", "--board_size", type=int, default=None, help="Board Size")
     parser.add_argument("-W", "--max_walls", type=int, default=None, help="Max walls per player")
-    parser.add_argument("-r", "--render", choices=["print", "curses"], default="print", help="Render mode")
+    parser.add_argument("-r", "--render", choices=["result", "text", "curses"], default="result", help="Render mode")
     parser.add_argument("--step_rewards", action="store_true", default=False, help="Enable step rewards")
+    parser.add_argument(
+        "--games_output_filename",
+        type=str,
+        default="game.yaml",
+        help="Save the played games to a file. Use 'None' to disable saving.",
```
alejandromarcu, line 18: No, you don't really want to save the ouptut to a file named 'None' 🤣

deep_quoridor/src/main.py
```diff
@@ -1,87 +1,46 @@
...
+        "board_size": args.board_size,
+        "max_walls": args.max_walls,
+        "step_rewards": args.step_rewards,
+        "renderer": renderer,
+        "saver": saver,
+    }
+
+    args = {k: v for k, v in args.items() if v is not None}
+    arena = Arena(**args)
+    arena.play_game(RandomAgent(), SimpleAgent())
```
alejandromarcu, line 46: TO DO: this should be also be configured from the parameters, e.g. `--players=RandomAgent,SimpleAgent`

deep_quoridor/src/main.py
```diff
@@ -1,87 +1,46 @@
...
     parser = argparse.ArgumentParser(description="Deep Quoridor")
     parser.add_argument("-N", "--board_size", type=int, default=None, help="Board Size")
     parser.add_argument("-W", "--max_walls", type=int, default=None, help="Max walls per player")
-    parser.add_argument("-r", "--render", choices=["print", "curses"], default="print", help="Render mode")
+    parser.add_argument("-r", "--render", choices=["result", "text", "curses"], default="result", help="Render mode")
     parser.add_argument("--step_rewards", action="store_true", default=False, help="Enable step rewards")
+    parser.add_argument(
+        "--games_output_filename",
+        type=str,
+        default="game.yaml",
```
adamantivm, line 17: I suggest a more evocative default name, such as `game.recording.yml` or `playback.yml`. This name is easily confused with a game definition.

deep_quoridor/src/main.py
```diff
@@ -1,87 +1,46 @@
...
+        help="Save the played games to a file. Use 'None' to disable saving.",
+    )
 
     args = parser.parse_args()
-    play(**vars(args))
+
+    # TO DO: this should be automatically detected from the environment
```
adamantivm, line 23: or use a factory method

deep_quoridor/src/random_agents.py
```diff
@@ -0,0 +1,13 @@
+from arena import Agent
```
adamantivm, line 1: nit: rename this file `random_agent.py` (singular) for consistency with `simple_agent`.
nit follow-up: move the agents to an `agents` folder, so we can keep adding there. And drop the _agent from the name.

deep_quoridor/src/arena.py
```diff
@@ -0,0 +1,91 @@
+from typing import Optional
+from quoridor_env import env
+
```
adamantivm, line 3: can you add a brief description on each "interface" here on what the intention you had in mind is? It will be useful to continue constructing these with the same idea in mind

deep_quoridor/src/main.py
adamantivm: or via a YAML file that describes a particular run with all its details. Sort of like a tournament specification.

adamantivm (commented):
Awesome, thank you. I left some suggestinos.

Also, I think the very first next step before Tournament is: "Play N times and report what percentage was won by each agent". I think that will be very useful for early benchmarking, so we don't have to manually play many times until we know which agent is better - one or two games may be just chance.

alejandromarcu:
I also want to use a factory method for the agents and allow the names to be passed as parameters, but I'll do it in a follow up PR with the N matches and tournaments

PR #109, opened 2025-03-18
Support multiple games in the arena

alejandromarcu:
This PR allows to:
- pass the players that will take part, e.g. `--players=random,simple`
- indicate that you want a tournament of all the player combinations with `-A`
- pass how many time each combination of players will play, e.g. `-t 10`
- print a table summarizing the tournament results.

E.g.:
```
> main.py -A -t=6                 

random vs simple: random won in 177 steps
simple vs random: simple won in 161 steps
random vs simple: simple won in 108 steps
simple vs random: random won in 246 steps
random vs simple: simple won in 198 steps
simple vs random: simple won in 891 steps
random vs bettersimple: bettersimple won in 210 steps
bettersimple vs random: bettersimple won in 179 steps
random vs bettersimple: bettersimple won in 70 steps
bettersimple vs random: random won in 72 steps
random vs bettersimple: bettersimple won in 82 steps
bettersimple vs random: bettersimple won in 97 steps
simple vs bettersimple: bettersimple won in 188 steps
bettersimple vs simple: bettersimple won in 123 steps
simple vs bettersimple: bettersimple won in 54 steps
bettersimple vs simple: bettersimple won in 157 steps
simple vs bettersimple: bettersimple won in 110 steps
bettersimple vs simple: bettersimple won in 65 steps


Player          | Wins   | Wins P1 
===================================
bettersimple    |     11 |      5 
simple          |      4 |      2 
random          |      3 |      1 
```

(bettersimple is just an agent that extends simple that I made for testing)

And the game recording file looks like this:
```
config:
  board_size: 9
  max_walls: 10
  step_rewards: false
games:
  game_0001:
    player1: random
    player2: simple
    winner: random
    steps: 177
    time_ms: 876
    actions:
    - 3
    - 147
    - 133
    - 158
 ```

### CODE REVIEW

File: deep_quoridor/src/agents/agent.py [hunk 1]
```diff
@@ -4,5 +4,26 @@ class Agent:
     Given a game state, the agent should return an action.
     """
 
+    agents = {}
```
adamantivm, line 7: nit: I would move this code to `__init__.py`. That way, when you look at the directory structure, you don't get confused by one file which is not an agent but an interface. I've seen this pattern in other bigger projects too.

deep_quoridor/src/main.py
```diff
@@ -1,15 +1,28 @@
...
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description="Deep Quoridor")
     parser.add_argument("-N", "--board_size", type=int, default=None, help="Board Size")
     parser.add_argument("-W", "--max_walls", type=int, default=None, help="Max walls per player")
     parser.add_argument("-r", "--renderer", choices=Renderer.names(), default="results", help="Render mode")
     parser.add_argument("--step_rewards", action="store_true", default=False, help="Enable step rewards")
+    parser.add_argument(
+        "-p",
+        "--players",
+        type=str,
```
adamantivm, line 16: argparse supports "list" arguments natively. Can you try changing it to that, if I"m not crazy? It will all be much cleaner, having the proper type from the get go

deep_quoridor/src/renderers.py
```diff
@@ -28,20 +28,42 @@ class NoneRenderer(Renderer):
...
-class TextRenderer(Renderer):
-    def start_game(self, game, agent1: Optional[Agent] = None, agent2: Optional[Agent] = None):
+            if r.winner == r.player1:
+                won[r.player1] += 1
+                won_as_p1[r.player1] += 1
+            else:
+                won[r.player2] += 1
+
+        print("\n")
+        print("Player          | Wins   | Wins P1 ")
```
adamantivm, line 53: Hmm ... I have the impression that more usual metrics are "win %" or at least "wins / total", otherwise it doesn't say much.
Also, I think a table showing pairwise results are more typical. I'm not an expert, but I suggest doing a quick search to see what's usual.

At the very minimum please include wins out of how many.

adamantivm (approved):
Cool! Left some suggestions, but feel free to merge whenever you want.

muralx (approved):
Thanks Cucu, some minor comments.

deep_quoridor/src/main.py
```diff
@@ -1,15 +1,28 @@
...
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description="Deep Quoridor")
     parser.add_argument("-N", "--board_size", type=int, default=None, help="Board Size")
     parser.add_argument("-W", "--max_walls", type=int, default=None, help="Max walls per player")
     parser.add_argument("-r", "--renderer", choices=Renderer.names(), default="results", help="Render mode")
     parser.add_argument("--step_rewards", action="store_true", default=False, help="Enable step rewards")
+    parser.add_argument(
+        "-p",
+        "--players",
```
muralx, line 15: Thinking at loud, we might want to have same agent with different model files playing each other. So this might need to decouple agentid from agent implementation. (no action needed, just something for us to think)

deep_quoridor/src/agents/agent.py
```diff
@@ -4,5 +4,26 @@ class Agent:
...
+    def __init_subclass__(cls, **kwargs):
+        friendly_name = Agent._friendly_name(cls.__name__)
+        Agent.agents[friendly_name] = cls
+
+    def name(self):
+        return Agent._friendly_name(self.__class__.__name__)
+
+    @staticmethod
+    def _friendly_name(class_name: str):
+        return class_name.replace("Agent", "").lower()
```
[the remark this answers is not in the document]
alejandromarcu: The result would be empty only if the name is Agent, AgentAgent, etc. Is that what you mean? in that case I don't think it's an issue

deep_quoridor/src/main.py
alejandromarcu: We'll want agents playing against themselves, but I don't think as part of the tournament would be helpful, rather just N games between them
I think the use would be for training, would be interesting to see how the game develops, and to see how much advantage there is in being first or second.
Right now you can use the same agent twice and it will work, but it's not clear who won:

```
(.venv) (base) amarcu@MacBookPro deep_rabbit_hole % /Users/amarcu/code/deep_rabbit_hole/.venv/bin/python /Users/amarcu/code/deep_rabbit_hole/deep_quoridor/src/main.py --players random random -t 2
game_0001: random vs random - random won in 101 steps
game_0002: random vs random - random won in 87 steps
```
We probably should do it so that if the player plays against itself, instead of giving you the friendly name of the agent, it just tells you player 1 or player 2.

Another thing we may run into is to run the same agent with different parameters. We could do that either by subclassing the agent, but may not be flexible enough.  Another possibility would be to be able to pass parameters when specifying the players, e.g. `--players dqn(20) dqn(30)` but we'd need some parsing

deep_quoridor/src/renderers.py
```diff
@@ -28,20 +28,42 @@ class NoneRenderer(Renderer):
...
-    def end_game(self, game, step):
-        print(f"Game over! {self.match}: {game.winner()} won in {step} steps")
+    def end_game(self, game, result: GameResult):
+        print(f"{self.match}: {result.winner} won in {result.steps} steps")
 
+    def end_arena(self, game, results: list[GameResult]):
+        won = {}
+        won_as_p1 = {}
+        for r in results:
+            for player in [r.player1, r.player2]:
```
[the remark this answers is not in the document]
alejandromarcu: I think it's fine here, because I wanted to do an action for the 2 players in the game, and instead of copy pasting the action and changing the player I did a loop with the 2 players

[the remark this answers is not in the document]
alejandromarcu: Also I'm removing this code anyway

deep_quoridor/src/renderers.py
alejandromarcu: Ok, the new version will show for example:

```
+------------+--------+--------+------------+--------+
|  P1 \ P2   | random | simple | simpleplus | Total  |
+------------+--------+--------+------------+--------+
|   random   |   -    |  60%   |    40%     |  50%   |
|   simple   |  80%   |   -    |     0%     |  40%   |
| simpleplus |  40%   |  80%   |     -      |  60%   |
|   ======   | ====== | ====== |   ======   | ====== |
|   Total    |  60%   |  70%   |    20%     |  50%   |
+------------+--------+--------+------------+--------+
```

alejandromarcu: Actually, that format doesn't work well.
I'm switching to something like this in another PR:

```
+---------------+--------+--------+------------+-------+
|     Player    | random | simple | simpleplus | Total |
+---------------+--------+--------+------------+-------+
|   P1 random   |   -    |  20%   |    60%     |  40%  |
|   P2 random   |   -    |  20%   |    20%     |  20%  |
|   P1 simple   |  80%   |   -    |    80%     |  80%  |
|   P2 simple   |  80%   |   -    |    20%     |  50%  |
| P1 simpleplus |  80%   |  80%   |     -      |  80%  |
| P2 simpleplus |  40%   |  20%   |     -      |  30%  |
+---------------+--------+--------+------------+-------+
```

deep_quoridor/src/main.py
adamantivm: I agree on two things:
 - It will become handy to be able to specify parameters for agents (e.g.: which snapshot to load)
 - It would be useful to always print: "Player 1: random" or something like that

For the agent parameters, it might get complicated enough that a program argument won´t  suffice and instead we'll need to move to a "tournament specification" (yaml?) or something like that. Especially if we want to specify parameters for a long list of agents participating in the tournament.

alejandromarcu: I think for getting more complex runs (or things that we want to store), we could just use python scripts under an experiments folder, like we did in the pool project.
E.g. the script could subclass agents to change the parameters, and then call a method with the desired configuration

PR #110, opened 2025-03-19
Windsurf-generated DeepQNetwork and training code - Claude Sonnet 3.7

adamantivm:
Update: I did a few trainings - with horrible / usless but mechanically correct results - and integrated this so that it can run on the Arena without breaking things.

Mind blowing ... it autogenerated all of this with a single prompt, after thinking for a bit.
I only had to correct one small thing and that was enough for the thing to run.
I'm not sure if the network will learn anything, it's oscillating and all, but I wanted to share this first step as I was truly impressed.

image

image


alejandromarcu:
Wow, very impressive! Is it the loss that is oscillating?  the network doesn't seem very big.  Also, I wonder if passing the board in a format similar to Jon's CombinedGrid and using CNNs may be helpful.  Can you consult with Windsurf?

adamantivm:
About the board format: yes, the idea is to start iterating with better architectures, but I wanted to have a trained baseline first.
I want to understand if this is actually doing what it should be doing first, and then after I am convinced it's OK, I will start iterating on the actual model.

deep_quoridor/src/agents/__init__.py
```diff
@@ -18,8 +18,8 @@ def _friendly_name(class_name: str):
         return class_name.replace("Agent", "").lower()
 
     @staticmethod
-    def create(friendly_name: str) -> "Agent":
-        return Agent.agents[friendly_name]()
+    def create(friendly_name: str, **kwargs) -> "Agent":
```
adamantivm, line 21: some agents may need some environment info to be created. The base class should allow for parameters and the Arena should make sure to pass the info down.

deep_quoridor/src/agents/flat_dqn.py
```diff
@@ -0,0 +1,193 @@
...
+        """Save the model to disk."""
+        torch.save(self.online_network.state_dict(), path)
+
+    def load_model(self, path):
+        """Load the model from disk."""
+        self.online_network.load_state_dict(torch.load(path))
+        self.update_target_network()
+
+
+class Pretrained01FlatDQNAgent(FlatDQNAgent):
```
adamantivm, line 179: This is a subclass that loads a specific model weights file. In the future this could be a parameter instead, so that we don't have to create dozens of subclasses per architecture.

adamantivm:
@alejandromarcu @jonbinney OR @muralx  Is it OK if I merge this? 
I still need to learn and research a lot about this code, it's not really "here's an awesome model!" but ruther: "here is some code that doesn't break things and we can play with".
Are you OK?
I want to avoid code divergence and also allow us to work jointly on this.

Here is a play run for your amusement:

```
(.venv) ➜  deep_rabbit_hole git:(dq/deep-q-training) /home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/.venv/bin/python /home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/deep_quoridor/src/play.py --players random pretrained01flatdqn
Loading pre-trained model from /home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/models/dqn_flat_nostep_final.pt
/home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/deep_quoridor/src/agents/flat_dqn.py:175: FutureWarning: You are using `torch.load` with `weights_only=False` (the current default value), which uses the default pickle module implicitly. It is possible to construct malicious pickle data which will execute arbitrary code during unpickling (See https://github.com/pytorch/pytorch/blob/main/SECURITY.md#untrusted-models for more details). In a future release, the default value for `weights_only` will be flipped to `True`. This limits the functions that could be executed during unpickling. Arbitrary objects will no longer be allowed to be loaded via this mode unless they are explicitly allowlisted by the user via `torch.serialization.add_safe_globals`. We recommend you start setting `weights_only=True` for any use case where you don't have full control of the loaded file. Please open an issue on GitHub for any issues related to this experimental feature.
  self.online_network.load_state_dict(torch.load(path))
game_0001: random vs pretrained01flatdqn - pretrained01flatdqn won in 144 steps
Loading pre-trained model from /home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/models/dqn_flat_nostep_final.pt
game_0002: pretrained01flatdqn vs random - random won in 296 steps
Loading pre-trained model from /home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/models/dqn_flat_nostep_final.pt
game_0003: random vs pretrained01flatdqn - pretrained01flatdqn won in 254 steps
Loading pre-trained model from /home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/models/dqn_flat_nostep_final.pt
game_0004: pretrained01flatdqn vs random - random won in 182 steps
Loading pre-trained model from /home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/models/dqn_flat_nostep_final.pt
game_0005: random vs pretrained01flatdqn - random won in 627 steps
Loading pre-trained model from /home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/models/dqn_flat_nostep_final.pt
game_0006: pretrained01flatdqn vs random - random won in 310 steps
Loading pre-trained model from /home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/models/dqn_flat_nostep_final.pt
game_0007: random vs pretrained01flatdqn - pretrained01flatdqn won in 238 steps
Loading pre-trained model from /home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/models/dqn_flat_nostep_final.pt
game_0008: pretrained01flatdqn vs random - pretrained01flatdqn won in 175 steps
Loading pre-trained model from /home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/models/dqn_flat_nostep_final.pt
game_0009: random vs pretrained01flatdqn - pretrained01flatdqn won in 198 steps
Loading pre-trained model from /home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/models/dqn_flat_nostep_final.pt
game_0010: pretrained01flatdqn vs random - random won in 490 steps
+---------------------+---------------------+--------+--------+
|       P1 \ P2       | pretrained01flatdqn | random | Total  |
+---------------------+---------------------+--------+--------+
| pretrained01flatdqn |          -          |  20%   |  20%   |
|        random       |         20%         |   -    |  20%   |
|        ======       |        ======       | ====== | ====== |
|        Total        |         20%         |  20%   |  20%   |
+---------------------+---------------------+--------+--------+

```

deep_quoridor/src/arena.py
```diff
@@ -128,8 +128,8 @@ def play_games(self, players: list[str], times: int):
         for i in range(len(players)):
             for j in range(i + 1, len(players)):
                 for t in range(times):
-                    agent_i = Agent.create(players[i])
-                    agent_j = Agent.create(players[j])
+                    agent_i = Agent.create(players[i], board_size=self.board_size)
+                    agent_j = Agent.create(players[j], board_size=self.board_size)
```
alejandromarcu, line 132: Not for this PR, but I think we should be passing here either the Arena or a new data class ArenaInfo or something.  E.g. we'll probably want the number of walls in the future

alejandromarcu (approved):
LGTM, I think it's ok to merge once the conflicts are fixed

deep_quoridor/src/arena.py
adamantivm: Yes. Note that I added `**kwargs` in the init of the agents, so that should allow passing anything needed.

PR #111, opened 2025-03-20
Fix the results table

alejandromarcu:
I realized that my previous PR displayed a table that had a problem.  The rows where when the P1 won, and I thought the columns would be when P2 won, but it doesn't work that way. I had asked chatGPT but I didn't clarify that we care about who is P1 vs P2.  Now that I clarified it suggested something like this, where you can see how well an agent did as P1 and as P2.

```
+---------------+--------+--------+------------+-------+
|     Player    | random | simple | simpleplus | Total |
+---------------+--------+--------+------------+-------+
|   P1 random   |   -    |  20%   |    60%     |  40%  |
|   P2 random   |   -    |  20%   |    20%     |  20%  |
|   P1 simple   |  80%   |   -    |    80%     |  80%  |
|   P2 simple   |  80%   |   -    |    20%     |  50%  |
| P1 simpleplus |  80%   |  80%   |     -      |  80%  |
| P2 simpleplus |  40%   |  20%   |     -      |  30%  |
+---------------+--------+--------+------------+-------+
```

### DIFF
--- a/deep_quoridor/src/renderers.py
+++ b/deep_quoridor/src/renderers.py
@@ -46,40 +46,33 @@ class ResultsRenderer(Renderer):
         all_players = sorted(set([r.player1 for r in results]) | set([r.player2 for r in results]))
         players = {player: i for i, player in enumerate(all_players)}
 
-        table = PrettyTable()
-        table.field_names = ["P1 \\ P2"] + list(players) + ["Total"]
-
-        N = len(players)
-        wins = np.zeros((N + 1, N + 1))
-        games = np.zeros((N + 1, N + 1))
+        games = defaultdict(lambda: 0)
+        wins = defaultdict(lambda: 0)
 
         for r in results:
-            games[players[r.player1], players[r.player2]] += 1
+            pair1 = (f"P1 {r.player1}", r.player2)
+            pair2 = (f"P2 {r.player2}", r.player1)
+            games[pair1] += 1
+            games[pair2] += 1
             if r.winner == r.player1:
-                wins[players[r.player1], players[r.player2]] += 1
+                wins[pair1] += 1
+            else:
+                wins[pair2] += 1
 
-        # Get the totals per row and column
-        wins[-1, :] = np.sum(wins, axis=0)
-        wins[:, -1] = np.sum(wins, axis=1)
-        games[-1, :] = np.sum(games, axis=0)
-        games[:, -1] = np.sum(games, axis=1)
+        table = PrettyTable()
+        table.field_names = ["Player"] + list(players) + ["Total"]
 
-        # Hacky way of adding an extra column and row for the totals
-        players["Total"] = max(players.values()) + 1
+        for player in players:
+            for pos_player in [f"P1 {player}", f"P2 {player}"]:
+                row_wins = [wins[(pos_player, opponent)] for opponent in players]
+                row_games = [games[(pos_player, opponent)] for opponent in players]
+                row_wins.append(sum(row_wins))
+                row_games.append(sum(row_games))
 
-        # Set the results in the PrettyTable
-        for player1, i1 in players.items():
-            row = [player1]
-            for _, i2 in players.items():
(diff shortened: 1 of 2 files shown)


deep_quoridor/src/renderers.py
```diff
@@ -46,40 +46,33 @@ def perc(wins: int, played: int) -> str:
...
-        # Hacky way of adding an extra column and row for the totals
-        players["Total"] = max(players.values()) + 1
-
-        # Set the results in the PrettyTable
-        for player1, i1 in players.items():
-            row = [player1]
-            for _, i2 in players.items():
-                row.append(perc(wins[i1, i2], games[i1, i2]))
+        table = PrettyTable()
+        table.field_names = ["Player"] + list(players) + ["Total"]
```
muralx, line 63: nit: Player -> [Player ]Winner/Opponent
nit: Total -> Overall

deep_quoridor/src/renderers.py
```diff
@@ -46,40 +46,33 @@ def perc(wins: int, played: int) -> str:
...
 
-        # Hacky way of adding an extra column and row for the totals
-        players["Total"] = max(players.values()) + 1
-
-        # Set the results in the PrettyTable
-        for player1, i1 in players.items():
-            row = [player1]
-            for _, i2 in players.items():
-                row.append(perc(wins[i1, i2], games[i1, i2]))
+        table = PrettyTable()
```
muralx, line 62: Nit: 
I think it could be better to split this render or add a configurable setting to it.
I imagine i would just want to see the final result without the line by line match result (mainly if we run hundreds of them). Also it could be good to display the important values of the arena config as description on top, so we have all the info at hand.

For instance
- Board 9x9
- Number of games: 10
- Number of walls allowed: 8

Results
\<The table here\>

Happy to do it myself if there is agreement in that it is what we want.

muralx (approved):

adamantivm:
Are you sure this makes sense @alejandromarcu ? I'm confused but ...

If you have this (which I think it's what you were trying to do):
 - Row: Player 1 name
 - Column: Player 2 name
 - Cell value: % of times Player 1 wins
Wouldn't that show everything we need?
If you want to know how simple did as P2, you just need to look at the column for "simple" and calculate (1 - cell value).

With this update, isn't the information a bit redundant?
Row: P1 Simple , Col: random == 1 - Row: random, Col: P1 Simple.

I would also leave the values of "random" vs "random". It should always be close to 50% and if it's not, then we have some interesting information (agent prefers P1 when playing against itself?)

Did you research how are these stats displayed normally? Maybe taking a peek at WandB support for multiplayer RL, if it has it? Or other such tools?

alejandromarcu:
@adamantivm, I did some quick searches and finally defaulted to chatgpt, who told me to do it like this:
image

You're right that the previous version will work considering that the column is percentage of loses, but I find it confusing, .e.g.

image

Anyway, I'll do more research later and we can keep discussing

deep_quoridor/src/renderers.py
adamantivm: >  your previous CL
You gave away your Google lineage @muralx

muralx: haha, CL sounds better than PR, it will take me some time.

alejandromarcu:
In the AlphaZero paper they do like this, but it's just 2 players.
image

DeepSeek also sugested an ELO rating.

deep_quoridor/src/renderers.py
alejandromarcu: We could accept more than one renderer, so you can combine them, and there would be for example one to print the result of each match and another one to print the summary.  It's easy to do, but maybe overengineering for now, when we still don't know what we need

alejandromarcu: @muralx , feel free to display the config values

muralx:
Can we have this change submitted? I think it is better than the one we have right now, even if it is redundant.

alejandromarcu:
Yes, doing this in #133

PR #112, opened 2025-03-20
Fix text renderer that was failing

muralx:


adamantivm:
I'll leave it to @alejandromarcu who should know better

alejandromarcu (approved):

PR #113, opened 2025-03-21
Adding arena stats, more renderers, and fixes

muralx:
Add progress bar renderer
Fix flickering in curses renderer
Split match/arena renders . (I don't think we will want match renderers when running large batches)


muralx:
@alejandromarcu ya que son los renderers que vos estuviste trabajando, te lo paso a vos para review.

FYI: Cero python en mi historial

deep_quoridor/src/agents/__init__.py
```diff
@@ -29,5 +29,5 @@ def get_action(self, game):
         raise NotImplementedError("You must implement the get_action method")
 
 
-from agents.random import RandomAgent
-from agents.simple import SimpleAgent
+from agents.random import RandomAgent  # noqa: E402, F401
+from agents.simple import SimpleAgent  # noqa: E402, F401
```
alejandromarcu, line 33: Nice, I didn't know about this noqa.  It's unrelated to this PR, but we can look to see if this is the right way to do the __init__
What linter are you using?

deep_quoridor/src/arena.py
```diff
@@ -27,7 +27,7 @@ def start_game(self, game, agent1: Agent, agent2: Agent):
     def end_game(self, game, result: GameResult):
         pass
 
-    def start_arena(self, game):
+    def start_arena(self, game, total_games: int):
```
alejandromarcu, line 30: We can consider instead of passing the total games to create a new type, e.g. `ArenaData` or something (could be similar to the `GameResult` class), to be able to pass more things.  e.g. a plug in may want to know the names of the player or information about the agents, and that would allow us to pass all information in a structure.  Not feeling strongly about doing it now anyway

deep_quoridor/src/arena.py
```diff
@@ -120,7 +120,7 @@ def _play_game(self, agent1: Agent, agent2: Agent, game_id: str) -> GameResult:
         return result
 
     def play_games(self, players: list[str], times: int):
-        self.plugins.start_arena(self.game)
+        self.plugins.start_arena(self.game, total_games=len(players) * (len(players) - 1) * times // 2)
```
alejandromarcu, line 123: I think the computation doesn't work for odd number of times.  We could also require the times to be even, or just double it and say that instead of games, it's games that each player plays as P1 and P2

alejandromarcu (commented):
Great job, you're already mastering Python!

deep_quoridor/src/arena.py
muralx: Agree, I did not like to pass just that info. If we need some extra info I will change it to data, or passing the arena  (as we do with the game).

deep_quoridor/src/agents/__init__.py
muralx: I don't know much about the style/best practice in python, but I would put the interface in its own file, and use the init file to initilize the agents' module (meaning forcing the load of the subclasses so they get registered) (I guess that is the purpose of these imports?)

deep_quoridor/src/arena.py
muralx: Not sure I follow your comment. The intention is to pass the total number of games the arena will run. I did not think about it too much, but gemini explains it better: https://g.co/gemini/share/5f1d962f904b
Let me know if you meant something else.

muralx:
Adding @jonbinney  and @adamantivm  as optional/FYI

alejandromarcu (approved):
LGTM

adamantivm:
LGTM. Sorry for the conflicts.

PR #114, opened 2025-03-21
Adding replay tool to allow replaying matches from saved yaml files

muralx:


adamantivm:
Looks great! LGTM

jonbinney:
Very cool! Is this so that humans can view old games, or is it part of the machine learning training?

muralx:
Thanks. @jonbinney I think it is for looking at a specific game. Probably, I am missing a delay betwen plays otherwise it it will be hard to actually see anything :) .

deep_quoridor/src/agents/agent_registry.py
```diff
@@ -0,0 +1,34 @@
...
+    @staticmethod
+    def names():
+        return list(AgentRegistry.agents.keys())
+
+    @staticmethod
+    def register(name: str, agent_class):
+        AgentRegistry.agents[name] = agent_class
+
+
+class SelfRegisteringAgent:
```
adamantivm, line 20: shouldn't this extend `Agent`? Or maybe just merge this into `Agent` and make it a single base class that includes self-registration?

deep_quoridor/src/agents/replay.py
```diff
@@ -0,0 +1,30 @@
+from agents import Agent
+
+
+class ReplayAgent(Agent):
```
adamantivm, line 4: Hmm ... why does this not extend SelfRegisteringAgent? Is this so that it can call itself like the original name?
I think it would be OK to make it a normal agent with a name that it creates itself dynamically based on the name, such as ReplayRandom or something like that ....
Otherwise, if you think it's better to keep the registering function out of the replay agent, then instead of merging the agent and selfregisteringagent classes (comment above), it would be fine to just make the selfregisteragent a subclass of agent.

adamantivm (approved):
I think it's ok @muralx . I made some comments.

Generally speaking:
 - I would keep all the base class and registration machinery in init, to avoid cluttering the dir with files that are not actual agentes
 - I would merge the two base classes into one, or otherwise make one subclass from the other (maybe that was a simple omission)

More generally: would merge fast and quick. I think the replay agent could have been merged and then you could have worked on this change to the registration mechanism in a subsequent PR.

deep_quoridor/src/agents/replay.py
muralx: The main goal was to decuple of the Agent interface that is used in Game and Arean from the way Agents are used from the way they are created (or class registration for the purpose of creation). SelfRegisteringAgent should have been an Agent, I missed that while refactoring the code.

deep_quoridor/src/agents/agent_registry.py
muralx: Yes, it should. Fixing this.

deep_quoridor/src/agents/__init__.py
```diff
@@ -1,33 +1,5 @@
-class Agent:
```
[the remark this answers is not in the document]
muralx: Thanks for the comments.

I explained the purpose of separating Agent (interface) from registration for instantiation purposes below (I hope that makes sense).

Regarding the usage of \__init\__: I will move those back into the \__init\__  file but I am adding my thinking about the topic. 

I was reading about it and I found that \__init\__ was mainly designed for module identification, initialization of the module and for exporting symbols from subpackages. Gemini suggested that it could be used to hide from the user some irrelevant code, while at the same time making it clear that it should not be over used for that purpose (The question was explicit about adding interfaces in the init file)

IMHO, as a *non Python programmer*, Agent/(and now AgentRegistry) are used by clients of the package and I find it cumbersome to go look for those types that I need to interact with inside the init file.

PR #115, opened 2025-03-22
Reuse the agent for different games

alejandromarcu:
As Julian mentioned, for agents that load models or do any heavy initialization, creating a new one for every game is inefficient. Instead, use the same instance for all the matches and call a new method `reset()` before starting


deep_quoridor/src/arena.py
```diff
@@ -81,12 +81,12 @@ def __init__(
         self.max_walls = max_walls
         self.step_rewards = step_rewards
         self.game = env(
-            board_size=board_size, max_walls=max_walls, step_rewards=step_rewards
+            board_size=board_size,
+            max_walls=max_walls,
+            step_rewards=step_rewards,
         )
```
alejandromarcu, line 87: @muralx not sure what's going on with the formatting, I tried switching the line length to different values and around 95 seemed closer, but there was always something different.  What are your settings?
I currently have what's in the repo, the `ruff.toml` file containing just:
```
line-length = 120
```
BTW, notice that the file is in deep_rabbit_hole, not in deep_quoridor, so maybe if you're opening deep_quoridor directly is not being taken into account?

adamantivm:
LGTM

deep_quoridor/src/arena.py
```diff
@@ -125,20 +125,24 @@ def _play_game(self, agent1: Agent, agent2: Agent, game_id: str) -> GameResult:
...
         self.plugins.start_arena(self.game)
 
         match_id = 1
         results = []
+        agents = []
+        for p in players:
+            if isinstance(p, Agent):
+                agents.append(p)
+            else:
+                agents.append(AgentRegistry.create(p))
```
alejandromarcu, line 138: @adamantivm , I added this so that we can pass an instance instead of the friendly name, that way we can instantiate it however we want.  It still doesn't allow to create custom instances from the command line, but it may be useful

deep_quoridor/src/arena.py
jonbinney: hmmm.... i don't know if it works if you don't open the root of the repo in vscode (that's what I've been doing, since in the past we've had dependencies between directories (e.g. annotation_utils)

jonbinney (commented):
LGTM

PR #116, opened 2025-03-23
Resending Renderers changes

muralx:
Allowing multiple renderers
Adding progressbar renderer
Fix flickering for curses renderer


alejandromarcu (approved):
LGTM

PR #117, opened 2025-03-23
First working DQN, wins at trivial game

adamantivm:
This MR includes a few fixes, to the point it was successfully proven to work on a trivial game.

The fixes:
 - Get the reward for the appropriate agent (it was fetching the reward to the opponent before)
 - When playing, initialize epsilon to 0 instead of 1 (it was playing randomly!)
Also a hack:
 - Amplify the reward so that it stands out more clearly from the random initialization weights

Some things are still known to be missing:
 - Training is only on blacks at the moment, we should extend it to whites
 - The negative reward for losing is being lost (not pun intended) and not used for training currently, it should be fixed
 - There are tons of opportunities for improvement in DQN architecture
 - Try with a more realistic game!

But at least this is now a baseline that shows to be working!

If you run the training with the following params:

```
/home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/.venv/bin/python /home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/deep_quoridor/src/train_dqn.py -N 3 -W 0 -e 10000 -u 100
```

It reliably produces an agent which can beat random on blacks 100% of the times:

```
(.venv) ➜  deep_rabbit_hole git:(dq/dqn-training-fixes) /home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/.venv/bin/python /home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/deep_quoridor/src/play.py -t 100 -N 3 -W 0 -p random pretrained01flatdqn -r arenaresults
Loading pre-trained model from /home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/models/dqn_agent_final.pt
...
Arena stats
===========
Board size:  3
Max walls:  0
Step rewards:  False
Total number of games:  100 

+---------------------+---------------------+--------+--------+
|       P1 \ P2       | pretrained01flatdqn | random | Total  |
+---------------------+---------------------+--------+--------+
| pretrained01flatdqn |          -          |   0%   |   0%   |
|        random       |          0%         |   -    |   0%   |
|        ======       |        ======       | ====== | ====== |
|        Total        |          0%         |   0%   |   0%   |
+---------------------+---------------------+--------+--------+
```

(the important number is row 2, column 1)

One really interesting thing to note is how things change when we change the reward multiplier (1000, 100, 1), the number of epochs to train, the epsilon decay and how often to save. Tweaking these values produce much more different training experiences. I recommend playing with them to see it first hand.

### DIFF
--- a/deep_quoridor/src/agents/flat_dqn.py
+++ b/deep_quoridor/src/agents/flat_dqn.py
@@ -183,7 +183,7 @@ class Pretrained01FlatDQNAgent(FlatDQNAgent):
 
     def __init__(self, board_size, **kwargs):
         super(Pretrained01FlatDQNAgent, self).__init__(board_size)
-        model_path = "/home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/models/dqn_flat_nostep_final.pt"
+        model_path = "/home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/models/dqn_agent_final.pt"
         if os.path.exists(model_path):
             print(f"Loading pre-trained model from {model_path}")
             self.load_model(model_path)
--- a/deep_quoridor/src/agents/flat_dqn.py
+++ b/deep_quoridor/src/agents/flat_dqn.py
@@ -6,6 +6,7 @@ import torch.optim as optim
 from collections import deque
 import random
 from agents import SelfRegisteringAgent
+from pathlib import Path
 
 
 class DQNNetwork(nn.Module):
@@ -182,8 +183,8 @@ class Pretrained01FlatDQNAgent(FlatDQNAgent):
     """
 
     def __init__(self, board_size, **kwargs):
-        super(Pretrained01FlatDQNAgent, self).__init__(board_size)
-        model_path = "/home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/models/dqn_agent_final.pt"
+        super().__init__(board_size, epsilon=0.0)
+        model_path = Path(__file__).resolve().parents[3] / "models" / "dqn_agent_final.pt"
         if os.path.exists(model_path):
             print(f"Loading pre-trained model from {model_path}")
             self.load_model(model_path)
--- a/deep_quoridor/src/train_dqn.py
+++ b/deep_quoridor/src/train_dqn.py
@@ -19,6 +19,12 @@ def train_dqn(
     """
     Train a DQN agent to play Quoridor.
 
+    Julian notes:
+    - This is for now working for a trivial 3x3 board with no walls
+    - It teaches the agent to use black (player 2) only, against a random agent
+      Note that in a 3x3 board with no walls, black always wins (if it wants)
+    - It's currently not assigning negative rewards for losing
+
     Args:
         episodes: Number of episodes to train for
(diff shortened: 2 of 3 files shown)


deep_quoridor/src/train_dqn.py
```diff
@@ -73,8 +79,13 @@ def train_dqn(
                 # Execute action
                 game.step(action)
 
-                # Get new state, reward, etc.
-                next_observation, reward, termination, truncation, _ = game.last()
+                # Get the observation and rewards for THIS agent (not the opponent)
```
adamantivm, line 82: @alejandromarcu this part is key, otherwise it was getting the rewards for the opponent instead

deep_quoridor/src/train_dqn.py
```diff
@@ -32,11 +32,8 @@ def train_dqn(
     """
     game = env(board_size=board_size, max_walls=max_walls, step_rewards=step_rewards)
 
-    # Calculate action space size
-    action_size = board_size**2 + ((board_size - 1) ** 2) * 2
-
     # Create the DQN agent
-    dqn_agent = FlatDQNAgent(board_size, action_size)
+    dqn_agent = FlatDQNAgent(board_size, epsilon_decay=0.9999)
```
adamantivm, line 36: @alejandromarcu and this is the other bug you found, this attempts to make it decay a bit slower

alejandromarcu: nit, remove the line above computing action_size.
Also, I'm wondering if the epsilon decay should happen after episode rather than after training step

jonbinney:
I like the commit message "First ever trivial training that works?". We should make all of your commit messages questions :-)

deep_quoridor/src/train_dqn.py
```diff
@@ -59,11 +62,14 @@ def train_dqn(
             observation, reward, termination, truncation, _ = game.last()
 
             # If the game is over, break the loop
+            # BUG: Missing recording a negative reward for the DQN agent when the opponent wins
             if termination or truncation:
+                if False:
+                    print(f"Winner: {game.get_opponent(agent_name)}. Episode reward: {episode_reward}")
```
alejandromarcu, line 68: We should use a logging framework so we can keep this things as `log.debug` or something and we can turn on and off from the command line

alejandromarcu (approved):
LGTM

deep_quoridor/src/agents/flat_dqn.py
```diff
@@ -183,7 +183,7 @@ class Pretrained01FlatDQNAgent(FlatDQNAgent):
 
     def __init__(self, board_size, **kwargs):
         super(Pretrained01FlatDQNAgent, self).__init__(board_size)
```
alejandromarcu, line 185:
```suggestion
        super().__init__(board_size, epsilon=0.0)
```
As I mentioned in chat, epsilon should be 0 when playing, otherwise if you don't specify it, it will be 1 so it will be playing randomly.  Additional nit, I removed the parameters to `super`, that was the format in Python 2.0

adamantivm:
This is now ready to merge. i recommend to check it out, it's exciting!

PR #124, opened 2025-03-24
Renderers: split them in files in renderers

alejandromarcu:
I want to do a PyGame renderer and the renderers file is getting too big, so I think it will be better to split it and use a structure similar to agents


alejandromarcu:
Resolves #125

deep_quoridor/src/arena.py
```diff
@@ -151,7 +151,7 @@ def replay_games(self, arena_data: dict, game_ids_to_replay: list[str]):
         This method simulates games using recorded moves from previous matches, allowing for
         replay and analysis of historical games.
         """
-        self.plugins.start_arena(self.game)
+        self.plugins.start_arena(self.game, len(game_ids_to_replay))
```
alejandromarcu, line 154: This was an unrelated thing that the IDE showed me it was missing a parameter

muralx: This was fixed on the last PR I sent you

muralx (approved):

deep_quoridor/src/arena.py
alejandromarcu: Ooops, I swear I read your PR, I forgot that you had fixed it 🤷

PR #126, opened 2025-03-25
Implement basic Quoridor rules in new class

jonbinney:
This implementation is meant to implement the rules in a way that is
easy for agents to use for planning, so that they don't need to copy the
entire environment every time they want to try out actions.

Limitations:
    - The "you can't completely block someone's path to the goal when placing a wall" rule hasn't been implemented yet in the new Quoridor class.
    - This PR doesn't update the environment class to use the new Quoridor class for game logic.
    - To avoid conflicts until the previous item is done, a new tests file has been created that tests just the new Quoridor class.

### DIFF
--- a/deep_quoridor/src/quoridor.py
+++ b/deep_quoridor/src/quoridor.py
@@ -1,0 +1,294 @@
+from dataclasses import dataclass
+from enum import IntEnum, unique
+
+import networkx as nx
+import numpy as np
+
+type Position = np.ndarray[tuple[int, int], np.dtype[np.uint8]]  # (row, col)
+
+
+@unique
+class WallOrientation(IntEnum):
+    VERTICAL = 0
+    HORIZONTAL = 1
+
+
+class Action:
+    pass
+
+
+@dataclass
+class MoveAction(Action):
+    destination: Position
+
+
+@dataclass
+class WallAction(Action):
+    position: Position  # Start of wall
+    orientation: WallOrientation
+
+
+class Board:
+    """
+    Grid which has both walls and possible pawn positions.
+
+    Pawn positions are on even rows and columns, walls are on odd rows and columns. We use negative
+    values to represent free cells and walls, 0 for player 1, and 1 for player 2.
+    integers.
+    """
+
+    # Possible values for each cell in the grid.
+    OUT_OF_BOUNDS = -2  # Returned if a move is off of the board.
+    FREE = -1
+    # The player numbers start at 0 so that they can also be used as indices into the player positions array.
+    PLAYER1 = 0
+    PLAYER2 = 1
+    # Walls are stored as 10 + player number.We leave 2-9 free for implementing more than 2 players in the future.
+    PLAYER1_WALL = 10
+    PLAYER2_WALL = 11
+
+    def __init__(self, board_size: int, max_walls: int):
+        self.board_size = board_size
+        self.max_walls = max_walls
+
+        self._grid = np.full((board_size * 2 - 1, board_size * 2 - 1), Board.FREE, dtype=np.int8)
+        self._player_positions = np.array([(0, board_size // 2), (board_size - 1, board_size // 2)])
+        for player, position in enumerate(self._player_positions):
+            self._grid[*position * 2] = player
+        self._walls_remaining = [self.max_walls, self.max_walls]
+
+    def get_player_position(self, player: int) -> Position:
+        """
+        Get the position of the player's pawn.
+        """
+        try:
+            return self._player_positions[player]
+        except IndexError:
+            raise ValueError(f"Invalid player {player}")
+
+    def move_player(self, player: int, new_position: Position):
+        """
+        Move the player's pawn. Doesn't check if the move is valid according to game rules.
+        """
+        new_position = np.asarray(new_position, dtype=np.uint8)
+
+        try:
+            old_position = self._player_positions[player]
+        except IndexError:
+            raise ValueError(f"Invalid player {player}")
+
+        try:
+            self._grid[*old_position * 2] = Board.FREE
+            self._grid[*new_position * 2] = player
+            self._player_positions[player] = new_position
+        except IndexError:
+            return Board.OUT_OF_BOUNDS
+
+    def get_cell(self, position: Position) -> int:
+        """
+        Get the value of a "player cell" in the grid.
+        """
+        position = np.asarray(position, dtype=np.uint8)
+        try:
+            return self._grid[*position * 2]
+        except IndexError:
+            return Board.OUT_OF_BOUNDS
+
+    def add_wall(self, player: int, position: Position, orientation: WallOrientation):
+        """
+        Mark the grid cells corresponding to the wall as occupied.
+        """
+        position = np.asarray(position, dtype=np.uint8)
+
+        if self._walls_remaining[player] < 1:
+            raise RuntimeError("Cannot place wall, no walls remaining for player {player}")
+        try:
+            self._grid[self._get_wall_slices(position, orientation)] = 10 + player
+        except IndexError:
+            return Board.OUT_OF_BOUNDS
+
+        self._walls_remaining[player] -= 1
+
+    def can_place_wall(self, position: Position, orientation: WallOrientation) -> bool:
+        """
+        Returns True if the wall can be placed at the given position and orientation.
+
+        Checks that the player has walls remaining, the wall is within the bounds of the board, and doesn't overlap with other walls.
+        """
+        position = np.asarray(position, dtype=np.uint8)
+        try:
+            return (self._grid[self._get_wall_slices(position, orientation)] == Board.FREE).all()
+        except IndexError:
+            return False
+
+    def is_wall_between(self, position1: Position, position2: Position) -> bool:
+        """
+        Check if there is a wall between two positions.
+        """
+        position1 = np.asarray(position1, dtype=np.uint8)
+        position2 = np.asarray(position2, dtype=np.uint8)
+        wall_position = position1 + position2
+        return self._grid[*wall_position] == Board.PLAYER1_WALL
+
+    def _get_wall_slices(self, position: Position, orientation: WallOrientation) -> tuple[slice, slice]:
+        """
+        Get a tuple of slices that correspond to the wall's cells in the grid.
+        """
+        if orientation == WallOrientation.VERTICAL:
+            return (slice(position[0] * 2, position[0] * 2 + 3), position[1] * 2 + 1)
+        elif orientation == WallOrientation.HORIZONTAL:
+            return (position[0] * 2 + 1, slice(position[1] * 2, position[1] * 2 + 3))
+        else:
+            raise ValueError("Invalid orientation: {orientation}")
+
+    def __str__(self):
+        """
+        Return a pretty-printed string representation of the grid.
+        """
+        display_grid = np.full(self._grid.shape, " ", dtype=str)
+        display_grid[::2, ::2] = "."
+        display_grid[self._grid == Board.PLAYER1_WALL] = "w"
+        display_grid[self._grid == Board.PLAYER2_WALL] = "W"
+        for player, _ in enumerate(self._player_positions):
+            display_grid[*self._player_positions[player] * 2] = str(player + 1)
+        return "\n".join([" ".join(row) for row in display_grid]) + "\n"
+
+
+class Quoridor:
+    def __init__(self, board: Board):
+        """
+        If you want to start from the initial game state, pass in board=Board(board_size, max_walls).
+        """
+        self.board = board
+        self.current_player = 0
+
+        self._jump_checks = create_jump_checks()
+
+    def step(self, action: Action, validate: bool = True):
+        """
+        Execute the given action and update the game state.
+
+        Turning off validation can save some computation if you're sure the action is valid,
+        and also makes it easier to teleport players around the board for testing purposes.
+        """
+        if validate and not self.is_action_valid(action):
+            raise ValueError("Invalid action")
+
+        if isinstance(action, MoveAction):
+            self.board.move_player(self.current_player, action.destination)
+        elif isinstance(action, WallAction):
+            self.board.add_wall(self.current_player, action.position, action.orientation)
+            # TODO: Check that the wall doesn't block any player from reaching their goal.
+        else:
+            raise ValueError("Invalid action type")
+
+        self.current_player = 1 - self.current_player
+
+    def is_action_valid(self, action: Action):
+        """
+        Check whether the given action is valid given the current game state.
+        """
+        if isinstance(action, MoveAction):
+            player = self.get_current_player()
+            current_position = self.board.get_player_position(player)
+            opponent = 1 - player
+            opponent_position = self.board.get_player_position(opponent)
+            opponent_offset = opponent_position - current_position
+            destination_position = np.asarray(action.destination, dtype=np.uint8)
+            position_delta = tuple(destination_position - current_position)  # Tuple so we can use it as a key.
+
+            # Destination cell must be free
+            is_valid = True
+            if self.board.get_cell(destination_position) != Board.FREE:
+                is_valid = False
+
+            elif np.sum(np.abs(position_delta)) == 1:
+                # Moving to an adjacent cell, just make sure no wall is in the way.
+                if self.board.is_wall_between(current_position, destination_position):
+                    is_valid = False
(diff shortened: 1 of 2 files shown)


jonbinney:
@ablasdel can you review this? Once it is merged you can implement "can't block someone's path completely" rule using the networkx library. I'm really tired right now so it's possible I left something really dumb in this PR.....

jonbinney:
Another feature of the new Quoridor class is the ability to check validity of individual actions, instead of computing the entire action mask like the pettingzoo environment does. I'm hoping this comes in handy for speeding up algorithms within agents who work by randomly exploring some subset of actions at each step (e.g. alphago style monte carlo tree search). For that we may need to move the reward calculation out of the Environment class, but that requires some more thought.

deep_quoridor/src/quoridor.py
```diff
@@ -0,0 +1,293 @@
...
+        Returns True if the wall can be placed at the given position and orientation.
+
+        Checks that the player has walls remaining, the wall is within the bounds of the board, and doesn't overlap with other walls.
+        """
+        position = np.asarray(position, dtype=np.uint8)
+        try:
+            return (self._grid[self._get_wall_slices(position, orientation)] == Board.FREE).all()
+        except IndexError:
+            return False
+
```
alejandromarcu, line 122: if there's a problem with the indexes, that sounds like a bug, we should let the exception go through or detect it early, so we're not masking the bug

deep_quoridor/src/quoridor.py
```diff
@@ -0,0 +1,293 @@
...
+            if self.board.get_cell(destination_position) != Board.FREE:
+                is_valid = False
+
+            elif np.sum(np.abs(position_delta)) == 1:
+                # Moving to an adjacent cell, just make sure no wall is in the way.
+                if self.board.is_wall_between(current_position, destination_position):
+                    is_valid = False
+
+            elif (tuple(opponent_offset), tuple(position_delta)) in self._jump_checks:
+                jump_checks = self._jump_checks[(tuple(opponent_offset), tuple(position_delta))]
```
alejandromarcu, line 210: You can use the walrus operator:
```suggestion
            elif (jump_checks := tuple(opponent_offset), tuple(position_delta)) in self._jump_checks:
```

deep_quoridor/src/quoridor.py
```diff
@@ -0,0 +1,293 @@
...
+        try:
+            return self._grid[*position * 2]
+        except IndexError:
+            return Board.OUT_OF_BOUNDS
+
+    def add_wall(self, player: int, position: Position, orientation: WallOrientation):
+        """
+        Mark the grid cells corresponding to the wall as occupied.
+        """
+        position = np.asarray(position, dtype=np.uint8)
```
alejandromarcu, line 100: do you need this line, since you're just passing position along?

alejandromarcu (commented):
First pass.  It's looking good, I like how compact the representation of the Board is!

The main thing I found is the argument checking, it's not very Pythonic, and also there are mixed things: returning OUT_OF_BOUNDS, raising ValueError, silently returning False, etc.

Some ideas:
- Define a Player(IntEnum) class, so that you don't need to check for players.
- Don't rely on IndexError, since it won't be raised when the value is -1, -2, etc. Instead, check the indices first
- Differentiate between this 2 errors:
    - The argument is out of bounds, e.g. player 3 or a position outside of the board.  (If you use a Player enum, I think it's ok not to check for it, since the linter should catch it).  You could define a helper method like `assert_position(position)` that would check that the position is inside the board,  and you could use Python's assert. They can even be disabled at runtime once we know there are no bugs.  But this is not checking user input, this is checking that the program is working correctly, so I think assert is appropriate
        - An invalid move was attempted (either moving the pawn or placing a wall).   While this would also normally be triggered by a bug, I think it's worth defining a custom exception to be able to easily know that we attempted to do something against the rules of the game.  E.g. in the future we could even train a model-lessa agent, in which case this is not a bug, but if the agent tries to cheat it gets a negative reward so that it learns that it can't be done

deep_quoridor/test/quoridor_test.py
```diff
@@ -0,0 +1,290 @@
...
+
+        else:
+            for i, ch in enumerate(row):
+                if ch == " " or ch == "+":
+                    continue
+
+                if ch == "v" or ch == "|":
+                    forbidden_walls.append((row_n, col_n - 1, WallOrientation.VERTICAL))
+                    if ch == "|":
+                        # We arbitrarily use player 1 to place the walls since it doesn't matter for these tests.
```
alejandromarcu, line 76: Maybe we should be placing walls alternating between player 1 and player 2, since now it does matter (e.g. I think there's a bug because it just checks walls for player 1)

jonbinney: Great point.

deep_quoridor/src/quoridor.py
jonbinney: TIL!

deep_quoridor/src/quoridor.py
```diff
@@ -0,0 +1,293 @@
...
+            raise ValueError(f"Invalid player {player}")
+
+        try:
+            self._grid[*old_position * 2] = Board.FREE
+            self._grid[*new_position * 2] = player
+            self._player_positions[player] = new_position
+        except IndexError:
+            return Board.OUT_OF_BOUNDS
+
+    def get_cell(self, position: Position) -> int:
```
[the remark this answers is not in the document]
jonbinney: Heh that was what I called it originally and then my obsession with fitting all statements on one line took over and I shortened it. I agree the descriptive name is better; I'll change it back.

deep_quoridor/src/quoridor.py
jonbinney: What do you mean by "problem with the indexes"? All jumps that can every be valid are in the `_jump_checks` dictionary, so this is checking whether this is a valid move. For example moving 3 spaces is never valid, and this catches that because it isn't in the dictionary.

deep_quoridor/src/quoridor.py
```diff
@@ -0,0 +1,293 @@
...
+            return False
+
+    def is_wall_between(self, position1: Position, position2: Position) -> bool:
+        """
+        Check if there is a wall between two positions.
+        """
+        position1 = np.asarray(position1, dtype=np.uint8)
+        position2 = np.asarray(position2, dtype=np.uint8)
+        wall_position = position1 + position2
+        return self._grid[*wall_position] == Board.PLAYER1_WALL
```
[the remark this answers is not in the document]
jonbinney: Oof. Great chatch.

jonbinney:
@alejandromarcu I like the idea of a player enum, and of using asserts for cases that are clearly bugs in the code. The negative indices thing is also a good point. I'll take a stab at those things tomorrow.

One thing I struggled with is that numpy and python sometimes have different ways of doing things.... in particular it's really ugly converting numpy arrays to tuples just to make them hashable so they can be used as dictionary keys. Maybe in a few weeks I'll go back and switch it over to a sparse numpy array or something.

deep_quoridor/src/quoridor.py
```diff
@@ -0,0 +1,293 @@
...
+        self._walls_remaining = [self.max_walls, self.max_walls]
+
+    def get_player_position(self, player: int) -> Position:
+        """
+        Get the position of the player's pawn.
+        """
+        try:
+            return self._player_positions[player]
+        except IndexError:
+            raise ValueError(f"Invalid player {player}")
```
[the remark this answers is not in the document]
ablasdel: +1 python reverse/negative indexing is fun but sometimes do unexpected things.

deep_quoridor/src/quoridor.py
```diff
@@ -0,0 +1,293 @@
...
+            old_position = self._player_positions[player]
+        except IndexError:
+            raise ValueError(f"Invalid player {player}")
+
+        try:
+            self._grid[*old_position * 2] = Board.FREE
+            self._grid[*new_position * 2] = player
+            self._player_positions[player] = new_position
+        except IndexError:
+            return Board.OUT_OF_BOUNDS
```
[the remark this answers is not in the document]
ablasdel: +1 In general I believe is bad form to use exceptions for what is essentially simple branching.

deep_quoridor/src/quoridor.py
```diff
@@ -0,0 +1,293 @@
...
+
+    def get_cell(self, position: Position) -> int:
+        """
+        Get the value of a "player cell" in the grid.
+        """
+        position = np.asarray(position, dtype=np.uint8)
+        try:
+            return self._grid[*position * 2]
+        except IndexError:
```
ablasdel, line 93: I prefer explicitly checking the bounds beforehand. Especially if we are going to return instead of reraise.

deep_quoridor/src/quoridor.py
alejandromarcu: The comment here doesn't seem to match the code, since I don't see any jumps here.  
Anyway, it's safer to explicitely check for the case rather than rely on exceptions.  E.g. in this case, if the position is inside the board (which probably should be checked earlier), then I would expect `_get_wall_slices` to return a valid index in the grid.

deep_quoridor/src/quoridor.py
ablasdel: How about we check if it is the proper type or raise an exception instead of converting? Simpler and has much less chance of doing something unexpected if someone passes something that doesn't convert.

deep_quoridor/src/quoridor.py
```diff
@@ -0,0 +1,293 @@
...
+        self._walls_remaining[player] -= 1
+
+    def can_place_wall(self, position: Position, orientation: WallOrientation) -> bool:
+        """
+        Returns True if the wall can be placed at the given position and orientation.
+
+        Checks that the player has walls remaining, the wall is within the bounds of the board, and doesn't overlap with other walls.
+        """
+        position = np.asarray(position, dtype=np.uint8)
+        try:
```
ablasdel, line 118: This try seems like it would be more readable as a simple if else.

deep_quoridor/src/quoridor.py
```diff
@@ -0,0 +1,293 @@
...
+            return Board.OUT_OF_BOUNDS
+
+    def add_wall(self, player: int, position: Position, orientation: WallOrientation):
+        """
+        Mark the grid cells corresponding to the wall as occupied.
+        """
+        position = np.asarray(position, dtype=np.uint8)
+
+        if self._walls_remaining[player] < 1:
+            raise RuntimeError("Cannot place wall, no walls remaining for player {player}")
```
ablasdel, line 103: Why are we raising when we are out of walls but returning an error code when we are out of bounds? 

Seems like the behavior for failing to place should match regardless of reason.

deep_quoridor/src/quoridor.py
ablasdel: +1 here on checking the indexes prior to use instead of catching the index error

deep_quoridor/src/quoridor.py
```diff
@@ -0,0 +1,293 @@
...
+        else:
+            raise ValueError("Invalid orientation: {orientation}")
+
+    def __str__(self):
+        """
+        Return a pretty-printed string representation of the grid.
+        """
+        display_grid = np.full(self._grid.shape, " ", dtype=str)
+        display_grid[::2, ::2] = "."
+        display_grid[self._grid == Board.PLAYER1_WALL] = "w"
```
ablasdel, line 149: Is there a reason we are keeping track of which player placed which wall where? I don't think there is anywhere we will use that info and it complicates the data model.

alejandromarcu:
With respect to numpy and tuples, if I'm not missing anything you are using numpy to do some basic arithmetic with the positions, e.g. multiply both coordinates by 2 at once, or `np.sum(np.abs(position_delta))`.
If that's the case, I think the code may be clearer ir you don't use numpy and just do the operations explicitly in tuples, and maybe define helper methods if you need.

For example, you could do something like this:

```
type Position = Tuple[int, int]

# ....
   
    def _player_to_grid_position(self, pos: Position) -> Position:
        assert 0 < pos[0] < self.board_size
        assert 0 < pos[1] < self.board_size
        return (pos[0] * 2, pos[1] * 2)

    def move_player(self, player: int, new_position: Position):
        new_r, new_c = self._player_to_grid_position(new_position)
        old_r, old_c = self._player_to_grid_position(self._player_positions[player])

        self._grid[old_r, old_c] = Board.FREE
        self._grid[new_r, new_c] = player
        self._player_positions[player] = new_position
```

and you can use `_player_to_grid_position` in a few other methods.  In other parts you would need to be explicit, e.g.  instead of: `wall_position = position1 + position2` you can do `wall_r, wall_c = position1[0] + position2[0], position1[1] + position2[1]` (or have it in 2 lines).  While that's a bit longer, you don't need to be using `np.asarray` conversions.

deep_quoridor/src/quoridor.py
```diff
@@ -0,0 +1,293 @@
...
+            "wall": [((1, 0), (2, 0))],
+            "nowall": [((0, 0), (1, 0)), ((1, 0), (1, 1))],
+        },
+        (1, -1): {
+            "wall": [((1, 0), (2, 0))],
+            "nowall": [((0, 0), (1, 0)), ((1, 0), (1, -1))],
+        },
+    }
+    # Rotate the checks for all 4 possible directions and convert to numpy arrays.
+    checks = {}
```
ablasdel, line 273: This is slick but also a bit hard to read. Provided it is well tested on this function we are. 👍

deep_quoridor/src/quoridor.py
alejandromarcu: I think in general we could:
- Let exceptions propagate (or raise / assert our own) when something shouldn't happen under normal circumstnaces, e.g. we're trying to acess an invalid position in the grid.  A normal behaving agent shouldn't do it, and for a human player, the position needs to be validated before then.
- Return values when it's the normal flow (e.g. if we're asking if the pawn can be moved somewhere or a wall placed somewhere).
- Raise a custom exception (`CheatException`?) when we're trying to do something that goes against the rules of the game, like moving the pawn to an invalid location or placing a conflicting wall.  The reason for doing that is that this shouldn't be happening anyway, but let's say that the agent has a bug and sends something wrong, we don't want to mask the bug and see an unexpected behavior (e.g. the agent had a turn but nothing happened).

deep_quoridor/src/quoridor.py
jonbinney: Done

deep_quoridor/src/quoridor.py
jonbinney: Done

deep_quoridor/src/quoridor.py
jonbinney: It's for display. I think (especially when debugging an agent) it will be useful to be able to color the walls differently based on who placed them.

deep_quoridor/src/quoridor.py
jonbinney: Yeaaaaah. I tried a bunch of different ways to check jumps, and all of them were hard to read. This is a big candidate for refactor at a later date if someone gets a good idea :-)

deep_quoridor/src/quoridor.py
jonbinney: I switched all the type checking to asserts, so that it errors if the wrong type is passed in and we can disable the checks for speed later.

deep_quoridor/src/quoridor.py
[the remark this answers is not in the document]
jonbinney: Fixed by adding a separate member function that properly checks whether positions are in bounds, and raising if not.

deep_quoridor/src/quoridor.py
jonbinney: I did all of these except the custom exception - I'll leave that to a future PR.

jonbinney (commented):
I addressed the comments (except for adding a custom exception type, which we can do later. I'm going to merge this so that @ablasdel can build on it.

jonbinney (commented):
Ah, forgot to say I also didn't remove the use of numpy  for player positions. It got about halfway therough that, and it was making things extra verbose. I'm not against it but the code felt really messy (might be i'm just too used to using numpy)

jonbinney:
Aw crap i forgot to push my commit with the PR fixes before I merged. I'll do a new PR and merge it right now. Sorry for the breakage!

PR #128, opened 2025-03-26
Assign negative rewards using reward backfilling

adamantivm:
I'm not 100% convinced this is good, but opening it up for review and discussion.

In any case, it's an optional flag so I suppose it can be included anyway and not used if it's not useful.

I couldn't get to do step-by-step debugging to confirm it works well.

For the record, here is what ChatGPT had to say. The first response seems very good, but then it was impossible to get a trustworthy reference for it, so I'm not convinced it's made up from hallucination: https://chatgpt.com/share/67e46aab-923c-800c-9c28-0028b301ac73

### DIFF
--- a/deep_quoridor/src/train_dqn.py
+++ b/deep_quoridor/src/train_dqn.py
@@ -13,10 +13,12 @@ def train_dqn(
     update_target_every,
     board_size,
     max_walls,
+    epsilon_decay=0.9999,
     save_path="models",
     model_name="dqn_agent",
     save_frequency=100,
     step_rewards=True,
+    assign_negative_reward=False,
 ):
     """
     Train a DQN agent to play Quoridor.
@@ -41,7 +43,7 @@ def train_dqn(
     game = env(board_size=board_size, max_walls=max_walls, step_rewards=step_rewards)
 
     # Create the DQN agent
-    dqn_agent = FlatDQNAgent(board_size, epsilon_decay=0.9999)
+    dqn_agent = FlatDQNAgent(board_size, epsilon_decay=epsilon_decay)
 
     # Create a random opponent
     random_agent = RandomAgent()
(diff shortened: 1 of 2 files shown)


deep_quoridor/src/train_dqn.py
```diff
@@ -158,23 +165,29 @@ def train_dqn(
...
+    parser.add_argument("-N", "--board_size", type=int, default=3, help="Board Size")
+    parser.add_argument("-W", "--max_walls", type=int, default=0, help="Max walls per player")
+    parser.add_argument("-e", "--episodes", type=int, default=2000, help="Number of episodes to train for")
     parser.add_argument("-b", "--batch_size", type=int, default=64, help="Batch size for training")
     parser.add_argument("-u", "--update_target", type=int, default=100, help="Episodes between target network updates")
     parser.add_argument("--step_rewards", action="store_true", default=False, help="Enable step rewards")
     parser.add_argument("--save_path", type=str, default="models", help="Directory to save models")
     parser.add_argument("--model_name", type=str, default="dqn_agent", help="Base name for saved models")
     parser.add_argument("--save_frequency", type=int, default=500, help="How often to save the model (in episodes)")
+    parser.add_argument("--epsilon_decay", type=float, default=0.999, help="Epsilon decay rate for exploration")
```
muralx, line 177: nit: add short form too

muralx (approved):

deep_quoridor/src/train_dqn.py
```diff
@@ -63,8 +65,13 @@ def train_dqn(
             observation, reward, termination, truncation, _ = game.last()
 
             # If the game is over, break the loop
-            # TODO: Assign negative reward to the DQN agent when the opponent wins
             if termination or truncation:
+                # Update the reward on the last element of the replay buffer (if there is any)
```
adamantivm, line 69: This is the key part. Can I have someone double-check this logic? Does this make sense to you all?
Another thing that I noticed is that when I enable this flag, loss becomes (and remains) super high.

@adamantivm @muralx @alejandromarcu @ablasdel wdyt?

muralx: From a logic perspective looks good to me (Negative reward, marked as done so reward is not adjusted). I will merge your changes and update the base class once you submit.

muralx (approved):

PR #129, opened 2025-03-27
Adding trainable agent, ruff config, fix renderer

muralx:
- Extracted common DQN code into a base class that can be easily inherited and overridden, base class includes hook methods for different aspects of the construction, observation, etc. (CodePilot refactoring)
- Made training to run unaware of what agents are running, meaning multiple trainable agents could run together (there are some additional changes that could be done for making it more generic, for now it is still for us to change code and play with it).
- Including a DExpAgent (Diego Experimental Agent). Not sure how you guys want to do this, but I find it handy to keep a class, I will be playing with, around. Easier to handle commits, changes, versions. Happy to remove if you guys prefer to keep this kind of code out.
- - For DExpAgent I added more layers to the NN, to see whether there was an improvement (Not really without changing other considerations). Also I plan to change representation.
- Fix Curses renderer that had an alignment issue
- Include new Ruff config enabling standard formatting,  import ordering.
- - Reordered imports to avoid warnings.

Sorry about the big PR with multiple unrelated changes.

### DIFF
--- a/deep_quoridor/src/agents/flat_dqn.py
+++ b/deep_quoridor/src/agents/flat_dqn.py
@@ -55,7 +55,7 @@ class FlatDQNAgent(AbstractTrainableAgent):
         opponent_walls = np.array([obs["opponent_walls_remaining"]])
 
         flat_obs = np.concatenate([board, walls, my_walls, opponent_walls])
-        return torch.FloatTensor(flat_obs, device=self.device)
+        return torch.FloatTensor(flat_obs).to(self.device)
 
 
 class Pretrained01FlatDQNAgent(FlatDQNAgent):
--- a/deep_quoridor/src/quoridor_env.py
+++ b/deep_quoridor/src/quoridor_env.py
@@ -27,11 +27,12 @@ Every time we represent a coordinate as a tuple, it is in the form (row, col)
 
 import copy
 import functools
-from typing import Tuple, Optional
+from typing import Optional, Tuple
+
+import numpy as np
+from gymnasium import spaces
 from pettingzoo import AECEnv
 from pettingzoo.utils import wrappers
-from gymnasium import spaces
-import numpy as np
 
 
 class QuoridorEnv(AECEnv):
--- a/.gitignore
+++ b/.gitignore
@@ -171,3 +171,4 @@ cython_debug/
 mlartifacts
 mlruns
 game_recording.yaml
+.vscode/launch.json
--- a/deep_quoridor/src/agents/__init__.py
+++ b/deep_quoridor/src/agents/__init__.py
@@ -1,64 +1,27 @@
-class Agent:
-    """
-    Base class for all agents.
-    Given a game state, the agent should return an action.
-    """
-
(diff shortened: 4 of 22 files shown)


deep_quoridor/src/agents/flat_dqn.py
```diff
@@ -36,145 +35,27 @@ def forward(self, x):
...
-        return random.sample(self.buffer, batch_size)
-
-    def __len__(self):
-        return len(self.buffer)
+class FlatDQNAgent(AbstractTrainableAgent):
+    """Agent that uses Deep Q-Network with flat state representation."""
 
+    def _calculate_action_size(self):
+        """Calculate the size of the action space."""
+        return self.board_size**2 + (self.board_size - 1) ** 2 * 2
```
adamantivm, line 43: I'm thinking, this probably should come from the environment, because it's actually up to the environment how it represents the actions that can be taken there, right?

adamantivm (commented):
Nice refactor!
Can you add a bit of context in the description of the PR to understand the motivation and how it has been used?

I'm curious what's the intent of the difference in dexp vs flat_dqn, and how you've been training this / with what results.

Also, it would be great if you could incorporate the minor changes I did in #128 so that's included in this more abstracted version as well.

deep_quoridor/src/agents/flat_dqn.py
muralx: This was an automatic refactor so I did not think too much about it. I do think that action representation for the game and action representation for the agent/model (same applies to board representation) don't need to match and should be decoupled (I have not done that in this PR, I want to get the new game first from @jonbinney). Once the new game is submitted I can modify the base class to include some kind of adapter between representations.

deep_quoridor/src/quoridor_env.py
```diff
@@ -27,11 +27,12 @@
 
 import copy
 import functools
-from typing import Tuple, Optional
+from typing import Optional, Tuple
+
+import numpy as np
```
jonbinney, line 32: Thanks for ordering these! I've been using `alt+shift+o` to manually tell vscode to reorder the imports - is there a way to have it do that on every save, like the rest of the formatting?

jonbinney: nvm - from your PR description it sounds like you may have done that already in the ruff config?

muralx: It wasn't actually doing it, but at least it was showing it as a problem (I am not familiar with python style, but I know they looked bad). I have just added the following to the settings.json in ~/.config/Code/User/settings.json and it is now fixing the importso on save
    "editor.codeActionsOnSave": {
        "source.fixAll": "always"
    },

adamantivm:
Can we merge this?

muralx:
@adamantivm I have incorporated your changes and I've also added determinism during training for ensuring we can reproduce thing.

adamantivm:
ok, sounds great! Let's get it merged then!

PR #135, opened 2025-03-28
Unify training logic inside agent 

muralx:
One more refactoring :)
- All training logic is now inside the trainable agent and can be tweak independent of the environment
- Added hook methods to the Trainable agent, and calls from standard arena to allow arena to be reused for training purposes
- Removed most logic from train.py except from model saving and rendering of rewards/losses (done with arena plugins: Thanks Cucu for those plug ins)
- Using 1, -1 for win/loss. In theory we can add something in the agent to scale numbers if needed
- I changed the step_rewards to user board_size^2, which seem to work better

Side note
After the refactoring I was able to train DExp(Diego Experimental) Agent on a 5x5b 0w which has a win ration of .95. In theory, you should be able to reproduce it due to (-i 42).

Training cmd: **python3 deep_quoridor/src/train.py  -N 5 -W 0 -d 0.9999 -e 1000 -i 42 -s**
Play cmd:  **python3 deep_quoridor/src/play.py -N 5 -W 0 -t 100  -r matchresults -p simple dexppretrained**

### DIFF
--- a/deep_quoridor/src/agents/core/agent.py
+++ b/deep_quoridor/src/agents/core/agent.py
@@ -11,14 +11,24 @@ class Agent:
     def name(self) -> str:
         return Agent._friendly_name(self.__class__.__name__)
 
-    def get_action(self, game) -> int:
-        raise NotImplementedError("You must implement the get_action method")
+    def is_learning_agent(self) -> bool:
+        """Returns True if the agent is a learning agent, False otherwise."""
+        return False
 
-    def reset(self):
-        """This method is called before starting a new game for the agent
-        to have a chance to reset its state before starting.
+    def start_game(self, game):
+        """This method is called when a new game starts.
+        It allows the agent to reset its state if needed.
         """
         pass
+
+    def end_game(self, game):
+        """This method is called when a game ends.
+        It allows the agent to clean up its state if needed.
+        """
+        pass
+
+    def get_action(self, game) -> int:
+        raise NotImplementedError("You must implement the get_action method")
 
 
 class AgentRegistry:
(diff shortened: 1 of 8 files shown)


muralx:
@alejandromarcu I removed the renderers as parameter of the play and replay method. Renderers are already part of the arena, so not sure you moved them as args of the method. Please let me know what I missed.

deep_quoridor/src/arena.py
```diff
@@ -111,8 +126,8 @@ def _replay_games(self, arena_data: dict, game_ids_to_replay: list[str]):
 
         self.plugins.end_arena(self.game, results)
 
-    def play_games(self, renderers: list[ArenaPlugin], players: list[str | Agent], times: int):
-        pygame_renderer = next((r for r in renderers if isinstance(r, PygameRenderer)), None)
+    def play_games(self, players: list[str | Agent], times: int):
+        pygame_renderer = next((r for r in self.renderers if isinstance(r, PygameRenderer)), None)
```
alejandromarcu, line 130: yeah, good point, there was no need to pass the renderer

deep_quoridor/src/agents/core/agent.py
```diff
@@ -11,15 +11,25 @@ def _friendly_name(class_name: str):
     def name(self) -> str:
         return Agent._friendly_name(self.__class__.__name__)
 
-    def get_action(self, game) -> int:
-        raise NotImplementedError("You must implement the get_action method")
+    def is_trainable(self) -> bool:
+        """Returns True if the agent is a learning agent, False otherwise."""
+        return False
+
+    def start_game(self, game, player_id):
```
alejandromarcu, line 18: Nice, actually I was working on an agent and noticed that the agent didn't know if it's player 1 or 2, so it doesn't explicitly know where to go.  I was adding it in the reset method but this works as well.  I think in the future we may want to standarize all this to use Player.ONE from quoridor.py, but we can do that separately

alejandromarcu (approved):
Very nice, great use of plug ins!

PR #136, opened 2025-03-28
Deterministic

muralx:
deterministic play
Model save and load using board and walls config


deep_quoridor/src/utils.py
```diff
@@ -0,0 +1,16 @@
...
+
+def set_deterministic(seed=42):
+    """Sets all random seeds for deterministic behavior."""
+    random.seed(seed)
+    np.random.seed(seed)
+    torch.manual_seed(seed)
+    torch.cuda.manual_seed_all(seed)
+    gymnasium.utils.seeding.np_random(seed)
+    torch.backends.cudnn.deterministic = True
+    torch.backends.cudnn.benchmark = False
```
alejandromarcu, line 16: So frustrating to have to set it everywhere 🤦

alejandromarcu (approved):

deep_quoridor/src/utils.py
muralx: There might be some better way of doing this, but for now it works. I found environment seed can also be set.

Issue #119, opened 2025-03-24
Create a heuristic agent


alejandromarcu:
A heuristic agent would raise the bar compared to the random agent and force it to respond better.
I'm thinking of the following algorithm:
- get the minimum distance to the target from the player and the opponent
- a configurable function would decide based on those 2 distances and maybe some randomness to either advance or block (e.g. the more the player is losing and the shorter the distance of the opponent to the target, the bigger the probability to block).
- If we decide to advance, we just move on one of the shortest path.
- If we decide to block, we try for each possible wall on the opponent path what would be the player and opponent shortest path, and chose the one that gives us the most advantage.  (In the initial version, we just choose the first possible block)

PR #137, opened 2025-03-29
Greedy heuristic agent

alejandromarcu:
Implements #119 
This is a first pass at a greedy agent.  I have some improvements in mind (and I wrote them as TO-DOs) but this is good enough to start and be able to challenge a bit more the training that with Random.
What it does is:
- Check the shortest path to the target from both players
- if the other player is less than 5 away (arbitrary number for now) and is closer to us, then we'll block, otherwise we advance.
- If we decided to block, we'll follow the shortest path of the opponent and try where we are allowed to put a wall (in the future I want to find the best greedy place, not the first).
- If we couldn't block or decided to advance, we just move in the direction of the shortest path.

This provides an entertaining playing between greedy agents (except for the jump, that was dumb), see below.

@jonbinney This will probably break when we switch to Board, but don't worry, I can fix it.

https://github.com/user-attachments/assets/e84efd1e-c2f8-4676-99a0-92d4ab357a55

### DIFF
--- a/deep_quoridor/src/agents/greedy.py
+++ b/deep_quoridor/src/agents/greedy.py
@@ -1,9 +1,10 @@
 from queue import Queue
 
 import numpy as np
-from quoridor import Player
 
 from agents.core import Agent
+
+type Position = tuple[int, int]
 
 
 class GreedyAgent(Agent):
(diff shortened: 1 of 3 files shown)


deep_quoridor/src/agents/greedy.py
```diff
@@ -0,0 +1,157 @@
...
+    def start_game(self, game, player_id):
+        self.player_id = player_id
+
+    def _valid_pawn_movements(self, game, action_mask: np.ndarray) -> list[Position]:
+        """Returns the coordinates of the possible movements based on the action mask.
+        This is useful for the first move, since there could be complicated jumps.  After the
+        first move, you can't easily use this because you'd need to generate a new action_mask
+        """
+        movement_mask = action_mask[: game.board_size**2]
+        valid_moves = [i for i, x in enumerate(movement_mask) if x == 1]
```
jonbinney, line 23: This is fine, but if you wanted to make it a bit faster and more numpyonic you could use `np.argwhere(movement_mask==1)`. And that's the same as `np.argwhere(movement_mask)` since 0 evaluates to false (but `x==1` is clearer so i don't have a preference.

deep_quoridor/src/agents/greedy.py
```diff
@@ -0,0 +1,157 @@
...
+        # Find the shortest path for each of the first moves and keep the shortest one.
+        for move in first_moves:
+            path = self._shortest_path_from(game, move, target_row)
+            if shortest_path is None or len(path) < len(shortest_path):
+                shortest_path = [move] + path
+
+        assert shortest_path is not None, "No path found"
+        return shortest_path
+
+    def _get_opponent_position(self, game, board: np.ndarray) -> Position:
```
jonbinney, line 100: Another good place to use `np.argwhere` to make this a bit easier to read.

deep_quoridor/src/agents/greedy.py
```diff
@@ -0,0 +1,157 @@
...
+        opponent_row = game.board_size - 1 - goal_row
+
+        my_shortest_path = self._shortest_path_from_me(game, observation, goal_row)
+        opponent_pos = self._get_opponent_position(game, observation["observation"]["board"])
+        opponent_shortest_path = self._shortest_path_from(game, opponent_pos, opponent_row)
+
+        # TODO: use a more elaborate logic to decide whether to block, could be probabilistic
+        block = False
+        if len(opponent_shortest_path) < len(my_shortest_path):
+            if len(opponent_shortest_path) < 5:
```
jonbinney, line 150: One suggestion for a slightly fancier way to do this in the future would be to compute _both_ the  best block action and compute the shortest path action, then choose between those based on which one leads to the highest `(len(opponent_shortest_path) - len(my_shortest_path)`.

jonbinney (approved):
LGTM!

deep_quoridor/src/agents/greedy.py
alejandromarcu: yeah, good idea, sometimes there may be a great opportunity for blocking that we're missing because the condition is not met

deep_quoridor/src/agents/greedy.py
alejandromarcu: that's a good one, I'm replacing it

PR #138, opened 2025-03-29
Treat edge of board as a wall for jump checking

jonbinney:
Also fixes the comparison logic that makes sure that the moves allowed by the Quoridor class are the same as the moves expected by the test. There were two problems before. When the two lists of moves were different lengths, zip() was ignoring the elements at the end of the longer array. Also the assert was on a list of truth values, and would always evaluate as true as long as the lists weren't of length zero.
  To fix this, we now convert the lists of moves to 2d numpy arrays of
size (num_points, 2) and then assert equality with np.testing.assert_equal.

### DIFF
--- a/deep_quoridor/src/quoridor.py
+++ b/deep_quoridor/src/quoridor.py
@@ -147,8 +147,20 @@ class Board:
         """
         assert is_valid_position_type(position1)
         assert is_valid_position_type(position2)
-        wall_position = position1 + position2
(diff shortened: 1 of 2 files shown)


jonbinney:
See https://boardgames.stackexchange.com/questions/57109/quoridor-jumps-at-the-edge-of-the-board for a discussion about this case. The consensus seems to be that you can jump diagonally if the edge of the board is behind the opponent.

alejandromarcu (approved):

PR #141, opened 2025-03-29
Fix logic in contiguous walls detection

muralx:
Also caches action mask and enforce it during step to prevent incorrect actions.


jonbinney (approved):
LGTM. I guess we didn't have any wall placement tests with walls that end at the edge. Might be good to add one of those, but no need to do that in this PR.

jonbinney:
oops... somehow i only saw the changes to the tests. Let me review again....

deep_quoridor/src/quoridor_env.py
```diff
@@ -267,14 +272,14 @@ def _is_wall_potential_block(self, row, col, orientation):
                 touches[0] = True
 
             # On the right border or touching another horizontal wall on the right sinde
-            if (col == self.wall_size) - 1 or (col < self.wall_size - 2 and self.walls[row, col + 2, 1] == 1):
+            if (col == self.wall_size - 1) or (col < self.wall_size - 2 and self.walls[row, col + 2, 1] == 1):
```
alejandromarcu, line 275: oops, sorry 🤦

deep_quoridor/src/quoridor_env.py
```diff
@@ -267,14 +272,14 @@ def _is_wall_potential_block(self, row, col, orientation):
...
                 touches[2] = True
 
             # Check for vertical walls touching it
             for r in range(top, bottom + 1):
                 for c in range(left, right + 1):
                     if self.walls[r, c, 0]:
-                        touches[c - left] = True
+                        touches[c - col + 1] = True
```
alejandromarcu, line 282: good catch, this was marking the wrong touches when the wall was against the wall, no?

alejandromarcu (approved):
Thanks for fixing the bugs I introduced, sorry! 🫤

jonbinney (approved):
ok I looked at all the files this time and it still LGTM :-)

deep_quoridor/src/quoridor_env.py
muralx: yes, that was the case that made the game hung in a loop for ever

PR #142, opened 2025-03-30
Fix type usage to make it compatible with python 311

muralx:


jonbinney (approved):
Works for me on python 3.12.3

jonbinney:
Curious why there are so many commits :-P

muralx:
> Curious why there are so many commits :-P

 @jonbinney They are duplicated commits from merging main from upstream into my local main fork. I am likely following the wrong flow when using the fork on github. (I have not used GitHub (neither forks) in the past, so need to read best practices.

PR #143, opened 2025-03-31
Agent log and UI visualization

alejandromarcu:
Allows the agents to log what they are planning in a structured fashion, e.g. they can log a path, a ranking of actions or text for cetain actions (it's easy to add new types).  The pygame UI takes the logs and renders them on the screen.  If we wanted, it would be easy to also render the logs as text or save them somewhere.
I also tried to do some improvements to the graphics, e.g. wall color and showing a red circle around the curent player.

BTW, pygame by default doesn't provide UI components such as buttons, that's why the buttons don't look and behave great.  In the future I'm thinking of using pygame_ui, so that this can be improved and we can use other UI components such as combo boxes

https://github.com/user-attachments/assets/1c574a40-a98c-4b60-a91e-195411ba8525

### DIFF
--- a/deep_quoridor/src/agents/__init__.py
+++ b/deep_quoridor/src/agents/__init__.py
@@ -15,9 +15,9 @@ __all__ = [
 
 from agents.core import (  # noqa: E402, F401  # noqa: E402, F401
     AbstractTrainableAgent,
+    ActionLog,
     Agent,
     AgentRegistry,
-    Log,
     ReplayBuffer,
 )
 from agents.dexp import DExpAgent, DExpPretrainedAgent  # noqa: E402
--- a/deep_quoridor/src/agents/__init__.py
+++ b/deep_quoridor/src/agents/__init__.py
@@ -17,6 +17,7 @@ from agents.core import (  # noqa: E402, F401  # noqa: E402, F401
     AbstractTrainableAgent,
     Agent,
     AgentRegistry,
+    Log,
     ReplayBuffer,
 )
 from agents.dexp import DExpAgent, DExpPretrainedAgent  # noqa: E402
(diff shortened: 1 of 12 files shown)


deep_quoridor/src/agents/__init__.py
```diff
@@ -17,6 +17,7 @@
     AbstractTrainableAgent,
     Agent,
     AgentRegistry,
+    Log,
```
muralx, line 20: nit: ActionLog

deep_quoridor/src/agents/core/agent.py
```diff
@@ -1,9 +1,86 @@
...
+
+    def action_score_ranking(self, action_scores: dict[int, float]):
+        """
+        Log a ranking of actions with their scores.
+        action_scores is a dictionary where the key is the action and the value is the score.
+        The ranking is created automatically and sorted in descending order.
+        """
+        if not self.is_enabled():
+            return
+        sorted_actions = sorted(action_scores.items(), key=lambda x: x[1], reverse=True)
```
muralx, line 56: fyi:  ss.rankdata does something like what you are doing

deep_quoridor/src/agents/core/agent.py
```diff
@@ -1,9 +1,86 @@
...
+
+
 class Agent:
     """
     Base class for all agents.
     Given a game state, the agent should return an action.
     """
 
+    def __init__(self):
+        self.log = Log()
```
muralx, line 82: nit: action_log

muralx (approved):
Thanks Cucu, looking forward to use this feature

deep_quoridor/src/agents/core/agent.py
alejandromarcu: thanks for the tip, but since we're not using scipy for now, I'm not sure if it's worth adding an extra package just for that. Numpy doesn't seem to have anything similar

PR #140, opened 2025-03-29
Figuring out Stable Baselines 3 and how to make it work with Petting Zoo

adamantivm:
This implements a new PPO agent and a separate training script but uses Stable Baselines 3 to implement and train the agent.

On initial tests with 3x3 board with no walls, training with default paramter yields pretty good results against random and greedy (NOTE that in a 3x3 board, P2 always wins if it plays right):

```
(.venv) ➜  deep_rabbit_hole git:(dq/sb3_primer) ✗ /home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/.venv/bin/python /home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/deep_quoridor/src/play.py -t 100 -N 3 -W 0 -p sb3ppo random
Progress: [--------------------------------------------------] 0.0% (0/100)Loaded model from quoridor_v0_20250407-215036.zip
/home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/deep_quoridor/src/train_sb3.py:41: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
  thobs = torch.cat((thobs, torch.tensor(v).flatten(start_dim=1)), dim=1)
Progress: [==================================================] 100.0% (100/100)
Arena stats
===========
Board size:  3
Max walls:  0
Step rewards:  False
Total number of games:  100 

+---------+--------+--------+--------+
| P1 \ P2 | random | sb3ppo | Total  |
+---------+--------+--------+--------+
|  random |   -    |   0%   |   0%   |
|  sb3ppo |  38%   |   -    |  38%   |
|  ====== | ====== | ====== | ====== |
|  Total  |  38%   |   0%   |  19%   |
+---------+--------+--------+--------+
(.venv) ➜  deep_rabbit_hole git:(dq/sb3_primer) ✗ /home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/.venv/bin/python /home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/deep_quoridor/src/play.py -t 100 -N 3 -W 0 -p sb3ppo greedy
Progress: [--------------------------------------------------] 0.0% (0/100)Loaded model from quoridor_v0_20250407-215036.zip
[...]

PR #145, opened 2025-04-01
Adding multiple new capabilities 

muralx:
Most changes were done to experiment with DExpAgent, that agent is the one that is using  most of them.
- New rotation utils with functions to rotate actions masks, actions, board and walls
- Add a way for dumping a json_config to print full args on the screen
- Encoded DExp play parameters in the file name so agent configures itself when reading the file. Hacky but good for know until we implement some level of config and tracking of training rounds.
- Move epsilon decay to end of game instead of per-transition
- Allows an extra argument after the classname during play to pass some basic info. Dexp-191 will pass 191 to the constructor of DExp. It is responsibility of the agent to interpret it. Note that agent will use those values to load the proper model.
- Allows the use of a multiplier for final results, configurable in the trainable agent
- DExp has support for: rotation, split of board for P1 and P2, inclusion of whether it is agent's turn
- DExp has a partial implementation to incorporate opponents transitions (not finished, it is only enabled by param). Hook methods are already added to trainable agent


deep_quoridor/src/quoridor_env.py
```diff
@@ -156,6 +156,7 @@ def _get_observation(self, agent_id):
         walls = self.walls.copy()
 
         return {
+            "my_turn": self.agent_selection == agent_id,
```
adamantivm, line 159: If we add this to the observation (which I support) we should also add it to he observation space. Can you add that before this specific change is merged?

deep_quoridor/src/quoridor_env.py
```diff
@@ -156,6 +156,7 @@ def _get_observation(self, agent_id):
...
             "board": board,
```
adamantivm, line 160: I don't remember where we ended up with the idea of separating the board in two one-hot grids. Should we include that here? If we do, we should also update the observation space.

deep_quoridor/src/quoridor_env.py
```diff
@@ -156,6 +156,7 @@ def _get_observation(self, agent_id):
...
             "walls": walls,
             "my_walls_remaining": self.walls_remaining[agent_id],
```
adamantivm, line 162: nit: would you mind also updating (or deleting) the large comment at the top of the file which explains the observation state format, so that it's not out of sync with the implementation? I'm afraid otherwise it will lead to confusion.

adamantivm (commented):
This is awesome stuff, thanks Diego!
I left a few small comments. The only probably mandatory one is the one about updating the state space in the environment.

I have two high-level comments:
 - Looks like some of the capabilities here have conflict / duplication with what Cucu did. I think if possible it would be better to end up with one or the other, not both, to reduce confusion on how we do things (convention over configuration)
 - I still have a feeling that much of the responsibility we're adding to the agent (e.g.: rotating the board, choosing how to use the "my turn" flag) should go in the environment or training code. The underlying reason for this is that I believe we're drifting away from good practices on how to interface an agent with an environment to solve a given problem, which includes iterating on various algorithms. It's more of a hunch, I may be wrong with this. Perhaps one idea to help us learn better about this would be to create yet another agent that we're working with and experiment with them in parallel. That will help us more clearly understand which decisions should go on an agent or not. Or at least, make sure we keep being able to train or at least play against older agents and we're not breaking things to the point where we lose our ability to benchmark as we progress. In any case, about this point, it can totally be done incrementally. Merging this and then figuring out how to generalize some capabilities out of the specific agent.

deep_quoridor/src/quoridor_env.py
muralx: I fully missed the space declaration, which I think it is there for a reason. Probably tools interact with it (and the observation space). Maybe you are right, but since multiple types of observations are also dependent on the agent and the research we are doing, I still think we should have a common API (core environment) and later maybe wrap the environment for observation space transformation. If you are OK, I will submit changes as they are so we have some baseline to experiment with RL (training) and at the same time dig into the env wrappers which I think it might be the way of getting the flexibility while sticking to the best practices.

adamantivm: I think it's OK to work on adapters later for the different space representations, but I think whatever we merge should have consistency between observation space and the format that we return in the observation, whatever that may be.

About how that is used, I think my open PR can give you a hint on how those are used by frameworks: https://github.com/jonbinney/deep_rabbit_hole/pull/140

muralx: Yes, it should not break anything. I think the changes to the space are already done in the new version of the PR. Only what is currently done by Dexp will need to be move to wrappers so they can be used in a more standard way.

muralx:
Submitting this now since Cucu will merge his way of doing args on top of this change.

deep_quoridor/src/agents/core/rotation_test.py
```diff
@@ -0,0 +1,130 @@
+import numpy as np
```
alejandromarcu, line 1: Yey for tests! but should we place them in test rather than src?

PR #146, opened 2025-04-02
Support for subargs

alejandromarcu:
I created a mini utility for sub arguments, i.e. passing arguments to an argument.  You can choose the separator (`,` by default) and the assginment (`=` by default).  So, for example, you can have a sub-argument `a=1,b=hey,c=true`.
Splitting the argument from the sub-argument is not part of the utility, so here I put it in AgentRegistry, and use `:` as the separator.  E.g. you can use `dex:a=1,b=hey,c=true`
The subargument utility takes a `@dataclass` and tries to coerce the subarguments into it, and (usually) provides a helpful error message. E.g. if I run `/play.py --players dexppretrained:include_turn=true,meh=true`. I get `utils.subargs.ParseSubargsError: Field 'meh' not in class DExpPlayParams.  Available fields are: use_rotate_board, include_turn, split_board`

I incorporated this into DExpAgent.
Other unrelated changes: fxied Deigo typos, moved utility.py to utility/misc.py

I was going to write unit tests but Diego has problems running them so I decided not to 🤣 

**Update**
In the last commit, I added p_random to greedy to take advantage of this new feature, that made it straightforward to use.
But then the problem is that if you make play greedy vs greedy with different parameters, you don't know which one is which.  To solve that, I updated the name to include the p_random parameter.  Still, this is long and ugly and can get much worse for agents with a lot of parameters, so I decided to include an extra param `nick` that you can use to specify whatever name you want to call it, that will be displayed for example in the arena results or in pygame.
E.g:

```
 %  python play.py  --players greedy:p_random=0.1,nick=greedy-ish dexppretrained:use_rotate_board=true,include_turn=true,nick=DiegoExperimental -r arenaresults2 -t 100
Loading pre-trained model from /Users/amarcu/code/deep_rabbit_hole/models/dexp_B9W10_C110_final.pt
Arena stats
===========
Board size:  9
Max walls:  10
Step rewards:  False
Total number of games:  100 

+----------------------+-------------------+------------+-------+
|        Player        | DiegoExperimental | greedy-ish | Total |
+----------------------+-------------------+------------+-------+
| P1 DiegoExperimental |         -         |     0%     |   0%  |
| P2 DiegoExperimental |         -         |     0%     |   0%  |
|    P1 greedy-ish     |        100%       |     -      |  100% |
|    P2 greedy-ish     |        100%       |     -      |  100% |
+----------------------+-------------------+------------+-------+
```


deep_quoridor/src/agents/greedy.py
```diff
@@ -1,16 +1,35 @@
...
 
 from agents.core import Agent
 
 Position: TypeAlias = tuple[int, int]
 
 
+@dataclass
+class GreedyParams(SubargsBase):
+    p_random: float = 0.0
+    nick: Optional[str] = None
```
muralx, line 17: should nick be part of the base class?

muralx (approved):
Thank you! Great to have this capability.

deep_quoridor/src/agents/greedy.py
alejandromarcu: The base class can be used for parameters other than the agent (e.g. in the future we may want to configure the renderers or something), so it would need to be an intermediate class.  I'll leave it like this for now and later we may want to create an intermediate class like AgentSubargsBase

PR #147, opened 2025-04-02
Plugins directory and other nits

alejandromarcu:
Making some changes in preparation for wandb


.vscode/settings.json
```diff
@@ -23,5 +23,9 @@
         "github.copilot",
         "github.copilot-chat"
     ],
-    "ruff.trace.server": "messages"
+    "ruff.trace.server": "messages",
+    "python.sortImports.args": ["--profile", "black"],
+    "editor.codeActionsOnSave": {
+      "source.organizeImports": true
+    }
```
alejandromarcu, line 30: this sorts the imports automatically on save

deep_quoridor/src/agents/__init__.py
```diff
@@ -8,7 +8,6 @@
     "RandomAgent",
     "ReplayAgent",
     "ReplayBuffer",
-    "SelfRegisteringAgent",
```
alejandromarcu, line 11: this class doesn't exist anymore

deep_quoridor/src/agents/dexp.py
```diff
@@ -75,9 +74,15 @@ def __init__(
             self.params = DExpPlayParams.from_str(agent_params_str)
         super().__init__(**kwargs)
 
+    def name(self):
+        return "dexp"
+
```
alejandromarcu, line 79: So we don't have to deal with the problem of extending classes having a different name, but actually I hope to be able to remove the pretrained classes and make it work automatically

deep_quoridor/src/arena.py
```diff
@@ -43,7 +44,7 @@ def _play_game(self, agent1: Agent, agent2: Agent, game_id: str) -> GameResult:
             observation, _, termination, truncation, _ = self.game.last()
             agent = agents[player_id]
             if termination or truncation:
-                if agent.is_trainable():
+                if agent.is_trainable() and isinstance(agent, AbstractTrainableAgent):
```
alejandromarcu, line 47: I added this because it was complaining about calling `handle_step_outcome`, since it doesn't know that the agent is an instance of AbstractTrainableAgent.  We could also just remove the method is_trainable and only ask if it's an instance

deep_quoridor/src/plugins/save_model.py
```diff
@@ -0,0 +1,32 @@
+import os
+
+from agents.core import AbstractTrainableAgent
+from arena_utils import ArenaPlugin
+
+
+class SaveModelEveryNEpisodesPlugin(ArenaPlugin):
```
alejandromarcu, line 7: I only moved this class

deep_quoridor/src/renderers/training_status.py
```diff
@@ -0,0 +1,37 @@
...
+    def start_game(self, game, agent1, agent2):
+        self.step = 0
+        self.player1 = agent1.name()
+        self.player2 = agent2.name()
+
+    def end_game(self, game, result):
+        if self.episode_count % self.update_every == 0:
+            for agent in self.agents:
+                agent_name = agent.name()
+                avg_reward, avg_loss = agent.compute_loss_and_reward(self.update_every)
```
alejandromarcu, line 21: I moved this class and the only change was moving this function to AbstractTrainableAgent so it can be called from other parts

.vscode/settings.json
```diff
@@ -23,5 +23,9 @@
         "github.copilot",
         "github.copilot-chat"
     ],
-    "ruff.trace.server": "messages"
+    "ruff.trace.server": "messages",
+    "python.sortImports.args": ["--profile", "black"],
+    "editor.codeActionsOnSave": {
```
muralx, line 28: weird you need this. It was already working for me with the formatOnSave that is at the top.

deep_quoridor/src/arena.py
muralx: I did not get that. I actually removed the instanceof to avoid forcing a type and used the duck typing capability of python.

I am fine using instanceof too.

muralx (approved):

PR #148, opened 2025-04-03
Move rotation test under /test folder

muralx:


adamantivm:
:thinking: is the file moved, actually?

alejandromarcu (approved):
@adamantivm , yes, it moved, it shows the old file name `->` new name

PR #149, opened 2025-04-03
Create transformation environment wrappers to allow "rotation and split board" features to be used as env wrappers

muralx:
For training purposes this might work fine. For simulation plays between different representations of the space, it will be more tricky, since env will need to be wrapped per agent.


deep_quoridor/src/environment/split_board_wrapper.py
```diff
@@ -0,0 +1,46 @@
...
+        observation = obs["observation"].copy()
+        board = observation["board"]
+        del observation["board"]
+
+        # Create one-hot representations for player and opponent
+        player_board = (board == 1).astype(np.float32)
+        opponent_board = (board == 2).astype(np.float32)
+
+        # Separate the one-hot representations
+        observation["my_board"] = player_board
```
adamantivm, line 27: NOTE: This is just a weak suggestion, or rather thinking out loud.
May I suggest, instead of using two separate dict keys, to make this into a 2-channel tensor, where we have the third dimension represent 0 for my and 1 for the opponent?
Furthermore, I'm thinking doing something that has the same shape and 2 more channels, one for the vertical walls and one for the horizontal walls may be even better.
But then way, what about e.g.: remaining walls and "it's my turn"?
Oh, and it may be even better to use that format that Jon and Cucu were talking about where it is all together represented in the same grid.
Hmm ... in conclusion then maybe there will be many wrapper variations to try different things, so it's not worth it to change this here?
Maybe just add "Dict" to the name to differentiate it from the different approach of having the split board as dimensions of a 3D tensor? Something like DictSplitBoardWrapper or so?

adamantivm:
Very clear use of wrappers, thank you!

deep_quoridor/src/environment/split_board_wrapper.py
muralx: I think we will want to experiment with different approaches (eg: the ideas you have), so we can have multiple wrappers with different transformations. We can compose them (in specific sequence since some might not work), or we can pre-create the composed version to avoid mistakes. 

I will rename this one to Dict, but I can also do the 2-channel version if you prefer. I am not planning on using it for now since I will continue playing with DExp variation so let me know which one you prefer since you will be the first user of them

adamantivm: I have a slight preference for the 3D option, but either is fine, as we can experiment later like you said

muralx: I added a 3d transformer that can be applied to the dict version

PR #150, opened 2025-04-04
Re-do: Weights & Biases (wandb) integration

alejandromarcu:
This is an improved version of https://github.com/jonbinney/deep_rabbit_hole/pull/144

This PR allows to track the training in wandb, as well as save and load model parameters and configuration.

**Usage**
If you don't want to use wandb during training (e.g. you're just doing some quick tests and care more about the code running than actually storing the results), either pass to the training `--no-wandb` or run from the command line `wand disabled` and then it won't be used until re-enabled

Otherwise, the same train command than before will log to wandb and provide a link to see the run. 

When using `play.py`:
- if you don't provide any additional information to the player (e.g. just `dexp`), then it will try to load the file as if it was saved without wandb (ending in `_final.py`).  If you run the training with wandb, you should use wandb in play as well.
- You can specify a local file to load, e.g. `dexp:model_filename=models/dexp_B7W0_mv1_C110_final.pt`
- In order to load a model from wandb, you need to pass the alias.  Each model can have more than 1 alias (but no alias can be in more than 1 model).  For example, by default you'll have the aliases v0, v1, v2...etc, increasing in each new model, and latest for the last one.  So, if you just want to use the last one, you can pass it as follows: `dexp:wandb_alias=latest`.  You could also choose a specific one, e.g. v0, or manually label something as `best` or whatever.
  - The "play parameters" are stored in wandb metadata when running the training and will be restored when loading.  E.g. if you train a model using `use_rotate_board=True` in DExpAgent, then when the model is loaded from wandb, the agent will have that configuration set.  If loading from a local file, you need to specify the parameters manually. (If useful, we could save the configuration in a local file in the future)
  - Notice that the wandb files are stored locally in the models directory to avoid re-downloading every time, using the first 5 digits of the digest in the name, e.g. `dexp_B5W5_mv1_b0c11.pt`, this way, if you're trying to load `latest` but the alias was moved to a different model, it will recognize that it's different and download it.

### DIFF
--- a/deep_quoridor/src/agents/core/trainable_agent.py
+++ b/deep_quoridor/src/agents/core/trainable_agent.py
@@ -1,15 +1,29 @@
 import os
 import random
+import tempfile
+from dataclasses import dataclass
 from pathlib import Path
-from typing import Tuple
+from typing import Optional, Tuple
 
 import numpy as np
 import torch
 import torch.nn as nn
 import torch.optim as optim
+from utils.subargs import SubargsBase
 
+import wandb
 from agents.core.agent import Agent
 from agents.core.replay_buffer import ReplayBuffer
+
+
+@dataclass
+class TrainableAgentParams(SubargsBase):
+    # Just used to display a user friendly name
+    nick: Optional[str] = None
+    # If wandb_alias is provided, the model will be fetched from wandb using the model_id and the alias
+    wandb_alias: Optional[str] = None
+    # If a filename is provided, the model will be loaded from disc
+    model_filename: Optional[str] = None
 
 
 class AbstractTrainableAgent(Agent):
@@ -26,6 +40,8 @@ class AbstractTrainableAgent(Agent):
         batch_size=64,
         update_target_every=100,
         assign_negative_reward=False,
+        training_mode=False,
+        params: TrainableAgentParams = TrainableAgentParams(),
         **kwargs,
     ):
         super().__init__()
@@ -40,9 +56,13 @@ class AbstractTrainableAgent(Agent):
         self.assign_negative_reward = assign_negative_reward
         self.final_reward_multiplier = 1
         self.action_size = self._calculate_action_size()
+        self.training_mode = training_mode
+        self.params = params
 
         # Setup device
         self.device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+
+        self.fetch_model_from_wand_and_update_params()
 
         # Initialize networks
         self.online_network = self._create_network()
@@ -56,10 +76,10 @@ class AbstractTrainableAgent(Agent):
         self.optimizer = self._create_optimizer()
         self.criterion = self._create_criterion()
         self.replay_buffer = ReplayBuffer(capacity=10000)
-        self.training_mode = False
         self.episodes_rewards = []
         self.train_call_losses = []
         self.reset_episode_related_info()
+        self.resolve_and_load_model()
 
     def reset_episode_related_info(self):
         self.current_episode_reward = 0
@@ -134,6 +154,16 @@ class AbstractTrainableAgent(Agent):
         )
 
         return avg_loss, avg_reward
+
+    def model_hyperparameters(self):
+        return {
+            "epsilon_min": self.epsilon_min,
+            "epsilon_decay": self.epsilon_decay,
+            "gamma": self.gamma,
+            "batch_size": self.batch_size,
+            "update_target_every": self.update_target_every,
+            "assign_negative_reward": self.assign_negative_reward,
+        }
 
     def _calculate_action_size(self):
         """Calculate the size of the action space."""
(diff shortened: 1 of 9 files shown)


deep_quoridor/src/agents/core/trainable_agent.py
```diff
@@ -293,10 +326,57 @@ def load_model(self, path):
...
+            filename = self.params.model_filename
         else:
-            raise FileNotFoundError(f"Model file {model_path} not found. Please provide the weights file.")
+            # If no filename is passed in training mode, assume we are not loading a model
+            if self.training_mode:
+                return
+
+            # If it's not training mode, we definitely need to load a pretrained model, so try the
+            # default path for local files
+            filename = Path(__file__).resolve().parents[4] / "models" / self.resolve_filename("final")
```
muralx, line 340: maybe make this an class argument too, we should not be guessing where to find the models.  I mean the resolve()..../models/

deep_quoridor/src/agents/core/trainable_agent.py
```diff
@@ -293,10 +326,57 @@ def load_model(self, path):
...
+        The params are updated to the artifact metadata.
+
+        """
+        alias = self.params.wandb_alias
+        if not alias:
+            return
+
+        api = wandb.Api()
+        artifact = api.artifact(f"the-lazy-learning-lair/deep_quoridor/{self.model_id()}:{alias}", type="model")
+        local_filename = Path(__file__).resolve().parents[4] / "models" / self.wandb_local_filename(artifact)
```
muralx, line 364: nit: also make configurable at least the root:
- we can have <path>/models, <path>/wandbmodels (the latter we know they are backed up)
-

deep_quoridor/src/agents/core/trainable_agent.py
```diff
@@ -293,10 +326,57 @@ def load_model(self, path):
...
+
+        with tempfile.TemporaryDirectory() as tmpdir:
+            artifact_dir = artifact.download(root=tmpdir)
+
+            path = artifact.download(root=artifact_dir)
+            tmp_filename = Path(path) / f"{self.model_id()}.pt"
+            if not os.path.exists(tmp_filename):
+                raise FileNotFoundError(f"Model file {tmp_filename} was not downloaded.  Please check the artifact")
+
+            os.rename(tmp_filename, local_filename)
```
muralx, line 380: does this work in windows if different disks are used?

deep_quoridor/src/plugins/wandb_train.py
```diff
@@ -0,0 +1,55 @@
...
+        if self.episode_count % self.update_every == 0 or self.episode_count == (self.total_episodes - 1):
+            avg_reward, avg_loss = self.agent.compute_loss_and_reward(self.update_every)
+
+            self.run.log(
+                {"loss": avg_loss, "reward": avg_reward, "epsilon": self.agent.epsilon},
+                step=self.episode_count,
+            )
+        self.episode_count += 1
+
+    def end_arena(self, game, results):
```
muralx, line 44: call wandb.finish()?

deep_quoridor/src/train.py
```diff
@@ -55,16 +56,20 @@ def train_dqn(
...
     agent2.final_reward_multiplier = 2
     # agent2.use_opponentns_actions = False
     # agent2.load_model("models/dexp_B5W0_base.pt")
     agent3 = GreedyAgent()  # noqa: F841
 
-    save_plugin = SaveModelEveryNEpisodesPlugin(
-        update_every=save_frequency, path=save_path, agents=[agent2], board_size=board_size, max_walls=max_walls
-    )
+    if use_wandb:
+        save_plugin = WandbTrainPlugin(update_every=10, path=save_path, agent=agent2, total_episodes=episodes)
```
muralx, line 67: can we use the sabemodelevery plugin and only have the Wandb plugin to save the model into wandb after it was saved by the file plugging.

I think we want to keep the save plugin to have checkpoints as we are doing, if we run a 10k training we want to have models saved every x episodes.

muralx (approved):
Thanks Cucu

deep_quoridor/src/agents/core/trainable_agent.py
alejandromarcu: Done, now you can pass `model_dir` or `wandb_dir` to specify the relative (to the root) or absolute path, and if omitted it will use "models" and "wandbmodels"

deep_quoridor/src/agents/core/trainable_agent.py
alejandromarcu: It wouldn't work, but I changed it so that the file is saved in the wandb directory so then the rename is just the name of the file

deep_quoridor/src/plugins/wandb_train.py
alejandromarcu: Good catch! now wandb is showing some more information when closing

PR #151, opened 2025-04-04
Fix some minor bug

muralx:
- Fix loss and rewards being swapped on plugins
- move all hyperparameters of dexp and trainable as constructor args
- shorter names of args to simplify command line
-

### DIFF
--- a/deep_quoridor/src/agents/core/trainable_agent.py
+++ b/deep_quoridor/src/agents/core/trainable_agent.py
@@ -9,10 +9,10 @@ import numpy as np
 import torch
 import torch.nn as nn
 import torch.optim as optim
+import wandb
 from utils.misc import resolve_path
 from utils.subargs import SubargsBase
 
-import wandb
 from agents.core.agent import Agent
 from agents.core.replay_buffer import ReplayBuffer
 
@@ -46,6 +46,7 @@ class AbstractTrainableAgent(Agent):
         update_target_every=100,
         assign_negative_reward=False,
         training_mode=False,
+        final_reward_multiplier=1,
         params: TrainableAgentParams = TrainableAgentParams(),
         **kwargs,
     ):
@@ -59,7 +60,7 @@ class AbstractTrainableAgent(Agent):
         self.batch_size = batch_size
         self.update_target_every = update_target_every
         self.assign_negative_reward = assign_negative_reward
-        self.final_reward_multiplier = 1
+        self.final_reward_multiplier = final_reward_multiplier
         self.action_size = self._calculate_action_size()
         self.training_mode = training_mode
         self.params = params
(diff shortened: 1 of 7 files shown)


deep_quoridor/src/plugins/wandb_train.py
```diff
@@ -33,7 +33,7 @@ def start_arena(self, game, total_games: int):
 
     def end_game(self, game, result):
         if self.episode_count % self.update_every == 0 or self.episode_count == (self.total_episodes - 1):
-            avg_reward, avg_loss = self.agent.compute_loss_and_reward(self.update_every)
+            avg_loss, avg_reward = self.agent.compute_loss_and_reward(self.update_every)
```
alejandromarcu, line 36: ooops, sorry 🤦

deep_quoridor/src/renderers/training_status.py
```diff
@@ -19,15 +19,16 @@ def end_game(self, game, result):
...
                 print(
-                    f"{agent_name} Episode {self.episode_count + 1}/{self.total_episodes}, Steps: {self.step}, Avg Reward: {avg_reward:.2f}, "
+                    f"{agent_name} Episode {self.episode_count + 1}/{self.total_episodes} [{'*' if won else ' '}], Steps: {self.step}, Avg Reward: {avg_reward:.2f}, "
                     f"Avg Loss: {avg_loss:.4f}, Epsilon: {agent.epsilon:.4f}"
                 )
         self.episode_count += 1
         return
 
-    def action(self, game, step, agent, action):
+    def after_action(self, game, step, agent, action):
```
alejandromarcu, line 31: No action required, but in the future we should use the @override decorator so that there's an error if the base method is renamed

deep_quoridor/src/agents/core/trainable_agent.py
```diff
@@ -203,6 +208,7 @@ def get_action(self, game):
         mask = observation["action_mask"]
 
         if random.random() < self.epsilon:
+            # print(f"rnd-mask: {mask}")
```
alejandromarcu, line 211: Maybe we should start using a log framework so that we can have lines like that and turn different levels of logging (we can do it later, no need to do anything now)

deep_quoridor/src/renderers/training_status.py
```diff
@@ -19,15 +19,16 @@ def end_game(self, game, result):
         if self.episode_count % self.update_every == 0:
             for agent in self.agents:
                 agent_name = agent.name()
-                avg_reward, avg_loss = agent.compute_loss_and_reward(self.update_every)
+                avg_loss, avg_reward = agent.compute_loss_and_reward(self.update_every)
+                won = result.winner == agent.name()
                 print(
-                    f"{agent_name} Episode {self.episode_count + 1}/{self.total_episodes}, Steps: {self.step}, Avg Reward: {avg_reward:.2f}, "
+                    f"{agent_name} Episode {self.episode_count + 1}/{self.total_episodes} [{'*' if won else ' '}], Steps: {self.step}, Avg Reward: {avg_reward:.2f}, "
```
alejandromarcu, line 25: Nice addition

alejandromarcu (approved):

PR #152, opened 2025-04-04
Moving all args into the agent config for agent related params

muralx:
Also simplifies the train.py file to instantiate player by using -p command line arguments including all settings

### DIFF
--- a/deep_quoridor/src/agents/core/trainable_agent.py
+++ b/deep_quoridor/src/agents/core/trainable_agent.py
@@ -9,10 +9,10 @@ import numpy as np
 import torch
 import torch.nn as nn
 import torch.optim as optim
-import wandb
 from utils.misc import resolve_path
 from utils.subargs import SubargsBase
 
+import wandb
 from agents.core.agent import Agent
 from agents.core.replay_buffer import ReplayBuffer
 
@@ -29,6 +29,25 @@ class TrainableAgentParams(SubargsBase):
     wandb_dir: str = "wandbmodels"
     # Directory where local models are stored
     model_dir: str = "models"
+    # Epsilon value for exploration
+    epsilon: float = 0.0
+    # Minimum epsilon value that epsilon can decay to
+    epsilon_min: float = 0.01
+    # Decay rate for epsilon
+    epsilon_decay: float = 0.995
+    # Discount factor for future rewards
+    gamma: float = 0.99
+    # Batch size for training
+    batch_size: int = 64
+    # Number of episodes between target network updates
+    # This is the number of games played, not the number of training steps
+    update_target_every: int = 100
+    # If True, the agent will receive a negative reward when game is lost
+    assign_negative_reward: bool = False
+    # If True, the agent is running in training mode
+    training_mode: bool = False
+    # If True, final reward will be multiplied by this value (positive or negative)
+    final_reward_multiplier: float = 1.0
 
 
 class AbstractTrainableAgent(Agent):
@@ -38,32 +57,24 @@ class AbstractTrainableAgent(Agent):
         self,
         board_size,
         max_walls,
-        epsilon=0,
-        epsilon_min=0.01,
-        epsilon_decay=0.995,
-        gamma=0.99,
-        batch_size=64,
-        update_target_every=100,
-        assign_negative_reward=False,
-        training_mode=False,
-        final_reward_multiplier=1,
         params: TrainableAgentParams = TrainableAgentParams(),
         **kwargs,
     ):
         super().__init__()
         self.board_size = board_size
         self.max_walls = max_walls
-        self.epsilon = epsilon
-        self.epsilon_min = epsilon_min
-        self.epsilon_decay = epsilon_decay
-        self.gamma = gamma
-        self.batch_size = batch_size
-        self.update_target_every = update_target_every
-        self.assign_negative_reward = assign_negative_reward
-        self.final_reward_multiplier = final_reward_multiplier
-        self.action_size = self._calculate_action_size()
-        self.training_mode = training_mode
+        self.initial_epsilon = params.epsilon
+        self.epsilon = params.epsilon
+        self.epsilon_min = params.epsilon_min
+        self.epsilon_decay = params.epsilon_decay
+        self.gamma = params.gamma
+        self.batch_size = params.batch_size
+        self.update_target_every = params.update_target_every
+        self.assign_negative_reward = params.assign_negative_reward
+        self.final_reward_multiplier = params.final_reward_multiplier
+        self.training_mode = params.training_mode
         self.params = params
+        self.action_size = self._calculate_action_size()
 
         # Setup device
         self.device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
@@ -166,14 +177,7 @@ class AbstractTrainableAgent(Agent):
         return avg_loss, avg_reward
 
     def model_hyperparameters(self):
-        return {
-            "epsilon_min": self.epsilon_min,
-            "epsilon_decay": self.epsilon_decay,
-            "gamma": self.gamma,
-            "batch_size": self.batch_size,
-            "update_target_every": self.update_target_every,
-            "assign_negative_reward": self.assign_negative_reward,
-        }
+        return {}
 
     def _calculate_action_size(self):
         """Calculate the size of the action space."""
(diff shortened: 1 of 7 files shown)


deep_quoridor/src/agents/core/trainable_agent.py
```diff
@@ -29,6 +29,25 @@ class TrainableAgentParams(SubargsBase):
...
+    batch_size: int = 64
+    # Number of episodes between target network updates
+    # This is the number of games played, not the number of training steps
+    update_target_every: int = 100
+    # If True, the agent will receive a negative reward when game is lost
+    assign_negative_reward: bool = False
+    # If True, the agent is running in training mode
+    training_mode: bool = False
+    # If True, final reward will be multiplied by this value (positive or negative)
+    final_reward_multiplier: float = 1.0
```
alejandromarcu, line 50: The only thing is that I don't think that we use most of them in play mode, but I think we can just ignore them so we don't have to make the code more complicated

deep_quoridor/src/agents/dexp.py
```diff
@@ -46,20 +46,21 @@ def forward(self, x):
...
-            split_board=bool(int(agent_params_str[2])),
-        )
-        return p
+    # Whether oppoments actions are used for training
+    # This is used for training only, not for playing
+    use_opponents_actions = False
 
+    # Parameters used for training which are required to be used with the same set of values during training
+    #  and playing are used to generate a 'key' to identify the model.
+    # This is used to load the model from disk or wandb
```
alejandromarcu, line 63: Unless you changed it and I missed it, we don't use it to load the model from wandb.  In wandb, we just load a specific model based on the alias and then we figure out the parameters from the metadata

deep_quoridor/src/plugins/wandb_train.py
```diff
@@ -31,6 +31,19 @@ def start_arena(self, game, total_games: int):
...
+        if (self.agent is not None) and (self.agent != agent1) and (self.agent != agent2):
+            raise ValueError("WandbTrainPlugin being used for an agent, but another agent is being trained")
+        if self.agent is not None:
+            return
+        if isinstance(agent1, AbstractTrainableAgent) and agent1.training_mode:
+            self.agent = agent1
+        elif isinstance(agent2, AbstractTrainableAgent) and agent2.training_mode:
+            self.agent = agent2
+        else:
+            raise ValueError("WandbTrainPlugin can only be used with a training agent, both agents are not training")
```
alejandromarcu, line 44: I'm not sure why you're making this change.  My idea was that the plugin is for one agent, and you want to train 2 agents at the same time, you can have 2 instances of the plugin.  That way the runs are separated.

alejandromarcu: Ok, after reading the rest of the code I see why. I think it's ok for now and then if we need it we can figure out a way of spawning one plugin per trainable agent

deep_quoridor/src/train.py
```diff
@@ -135,32 +72,35 @@ def train_dqn(
         default=True,
         help="Disable Weights & Biases logging",
     )
+    parser.add_argument(
+        "-p",
+        "--players",
+        nargs="+",
+        type=player_with_params,
+        help=f"List of players to compete against each other. Can include parameters in parentheses. Allowed types {AgentRegistry.names()}",
+    )
```
alejandromarcu, line 81: Maybe it's worth adding also an argument for the renderers, since I see you manually switched to pygame?

deep_quoridor/src/train.py
```diff
@@ -135,32 +72,35 @@ def train_dqn(
         default=True,
         help="Disable Weights & Biases logging",
     )
+    parser.add_argument(
+        "-p",
+        "--players",
+        nargs="+",
+        type=player_with_params,
+        help=f"List of players to compete against each other. Can include parameters in parentheses. Allowed types {AgentRegistry.names()}",
```
alejandromarcu, line 80: So can you now train an agent against a few different opponents?

alejandromarcu (approved):

alejandromarcu:
> @alejandromarcu What do you think? I did remove the hyperparameters since all player params contain that info. Let me know what you think
> 
> Example output: https://wandb.ai/the-lazy-learning-lair/deep_quoridor/runs/g7tagozh/overview

Looks good to me, I just wanted it saved somewhere

alejandromarcu:
Actually, I have an idea for the plugins like WandbTrainPlugin and TrainingStatusRenderer that should have one instance per agent.

I asked chatgpt to code it with this prompt:
```
Write a plugin named PerAgentPlugin that will be a plugin that forwards each call to some (may be 0) agents. It could be instantiated for example as follows: PerArenaPlugin(creator=lambda agent: MyPlugin(agent)) The plugin will keep a dict pointing from agent instances to plug in instances. The collection will be initially empty. When calling start_game, it will check for each agents if they are already in the dict, and if not, it would call the creator lambda to get an instance. Notice that the creator can return None, in which case we will skip the agent because it means that it doesn't require the plugin. If it returns an insttance, it will be saved in the internal collection. Then, still in start_game, the plugins for this game will be stored in an instance variable, and cleared in end_game. For each one of the calls (except start_arena), we will call the same method in the plug ins corresponding to the agents for the game
```

It had some issues so I asked it:
```
Include a type for creator, which will take an Agent and return an ArenaPlugin instance. Then, we don't need to check if the plugin has each one of the methods. Additionally, we should forward end_arena, for that we will need to not clear the current plugins in end_game, that's ok
```

And this is what I got, which seems reasonable but I haven't tried it. Probably I'm missing something and would en up being more complicated, but the idea would be to keep the implementation of WandbTrainPlugin to still receive 1 agent (and we could make TrainingStatusRenderer get just 1 agent), and then we would use a plug in like: `PerAgentPlugin(labmda agent: WandbTrainPlugin(update_every, episodes, agent) if isinstance(Agent, AbstractTrainableAgent) else None`

```
class PerAgentPlugin(ArenaPlugin):
    """
    A plugin that forwards calls (except start_arena) to agent-specific plugins.
    
    The plugin is instantiated with a creator function that takes an Agent and returns an ArenaPlugin
    instance (or None if the agent does not require a plugin). The instance keeps a dictionary to map agents
    to their plugin instances. For each game (start_game), a mapping for the current game is also built.
    
    All calls except start_arena will be forwarded to the agent-specific plugin. Notably, end_arena is forwarded,
    so the current game mapping is only cleared after end_arena.
    
    Example instantiation:
        per_agent_plugin = PerAgentPlugin(creator=lambda agent: MyPlugin(agent))
    """
    def __init__(self, creator: Callable[[Agent], Optional[ArenaPlugin]]):
        self.creator = creator
        self._agent_plugins: Dict[Agent, ArenaPlugin] = {}
        self._current_game_plugins: Dict[Agent, ArenaPlugin] = {}

    def start_game(self, game, agent1: Agent, agent2: Agent):
        # Ensure each agent has a plugin instance.
        for agent in (agent1, agent2):
            if agent not in self._agent_plugins:
                plugin_instance = self.creator(agent)
                if plugin_instance is not None:
                    self._agent_plugins[agent] = plugin_instance
        # Build the mapping for the current game (only include agents with a plugin)
        self._current_game_plugins = {}
        for agent in (agent1, agent2):
            if agent in self._agent_plugins:
                self._current_game_plugins[agent] = self._agent_plugins[agent]
                self._agent_plugins[agent].start_game(game, agent1, agent2)

    def end_game(self, game, result: GameResult):
        for plugin in self._current_game_plugins.values():
            plugin.end_game(game, result)
        # Do not clear _current_game_plugins here so that end_arena can be forwarded.

    def before_action(self, game, agent: Agent):
        if agent in self._current_game_plugins:
            self._current_game_plugins[agent].before_action(game, agent)

    def after_action(self, game, step, agent: Agent, action):
        if agent in self._current_game_plugins:
            self._current_game_plugins[agent].after_action(game, step, agent, action)

    def start_arena(self, game, total_games: int):
        # start_arena is not forwarded to agent-specific plugins.
        pass

    def end_arena(self, game, results: List[GameResult]):
        for plugin in self._current_game_plugins.values():
            plugin.end_arena(game, results)
        # Clear the current game mapping.
        self._current_game_plugins.clear()
```

deep_quoridor/src/agents/core/trainable_agent.py
muralx: Agree, I would have liked to have play args + train args, but actually they were all already there in the constructor so it does not really change things. Also I think that for wanddb all these args matter.

deep_quoridor/src/plugins/wandb_train.py
muralx: Yes, it is ugly and also don't like it. I think the problem is we are using arena plugins as agent plugins, since we don't have support for them. I am fine throwing this away.

alejandromarcu:
I think it's ok to merge and then we can iterate on how to specify the training and the agent plugins

PR #153, opened 2025-04-06
Fix passing of renderers

muralx:


deep_quoridor/src/arena.py
```diff
@@ -27,7 +28,7 @@ def __init__(
         self.game = env(board_size=board_size, max_walls=max_walls, step_rewards=step_rewards)
 
         self.renderers = renderers
-        self.plugins = CompositeArenaPlugin([p for p in plugins + renderers + [saver] if p is not None])
+        self.plugins = CompositeArenaPlugin([p for p in np.concatenate([plugins, renderers, [saver]]) if p is not None])
```
alejandromarcu, line 31: I don't understand what's going on, concatenating lists should work with `+`,  and np.concatenate shouldn't, e.g.:

```
>> a = ["a", "b"]
>>> a + ["c"]
['a', 'b', 'c']
>>> np.concatenate(["a"], ["b"])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'list' object cannot be interpreted as an integer
```

alejandromarcu: Actually I see, you're using np.concatenate like this:

```
>>> np.concatenate([a, ["c"]])
array(['a', 'b', 'c'], dtype='<U1')
```

But why is the `+` not working for lists?

muralx: Not sure, I wanted to get it fixed quickly since it broke the train.py script.

deep_quoridor/src/train.py
```diff
@@ -42,7 +43,7 @@ def train_dqn(
         board_size=board_size,
         max_walls=max_walls,
         step_rewards=step_rewards,
-        renderers=np.concatenate([print_plugin], renderers),
+        renderers=np.concatenate([[print_plugin], renderers]),
```
alejandromarcu, line 46: I tried the train script and I was getting an error, but I fixed it using `        renderers=[print_plugin] + renderers,
`

deep_quoridor/src/arena.py
alejandromarcu: What was the error? I tried running train.py with the last parameters I saw in your run and I got just the error in arena.py, not here

muralx: First step was to fix the train to actually pass [[print_plugin], renderers], but after that there is an error:

  File "/home/mural/lazy/deep_rabbit_hole/deep_quoridor/src/arena.py", line 30, in __init__
    self.plugins = CompositeArenaPlugin([p for p in plugins + renderers + [saver] if p is not None])
                                                    ~~~~~~~~^~~~~~~~~~~
TypeError: unsupported operand type(s) for +: 'SaveModelEveryNEpisodesPlugin' and 'TrainingStatusRenderer'

I am sure there is something else that could be done, but concatenate fixed it too.

deep_quoridor/src/train.py
muralx: right, but see the error above

deep_quoridor/src/arena.py
muralx: Changed to python list style

alejandromarcu: Based on your error, it looks that both plugins and renderers are objects, which is not what I'm expecting based on the code and on the names, I'm expecting both to be lists of objects, and then they can be concatened with +

alejandromarcu (approved):
what happened with the other thing in arena?

PR #154, opened 2025-04-07
Using explicit naming for training sessions

muralx:
Using name generated as {run_prefix)-{run_suffix} both allowed from commandline. Suggested to use your name/nick as prefix and let the suffix use the default YYYYMMDD-HHMMSS
- Using the run name in wandb as id and name
- using run name to save files in local storage to avoid overwriting past files


alejandromarcu (approved):
LGTM.

PR #155, opened 2025-04-08
Integrating SB3/PPO training and agent with WandB

adamantivm:
Integrates WandB into PPO so now it's possible to save new models when training as well as play with pretrained models from wandb.

### DIFF
--- a/deep_quoridor/src/train_sb3.py
+++ b/deep_quoridor/src/train_sb3.py
@@ -14,6 +14,7 @@ TODO:
 import glob
 import os
 import time
+from dataclasses import asdict
 
 import pettingzoo.utils
 import quoridor_env
@@ -121,6 +122,11 @@ def mask_fn(env):
     return env.action_mask()
 
 
+def make_model_id(env_kwargs):
+    # HACK: This is takenfrom trainable_agent model_id()
+    return f"sb3ppo_B{env_kwargs['board_size']}W{env_kwargs['max_walls']}_mv0"
+
+
 def train_action_mask(env_fn, steps=10_000, seed=0, **env_kwargs):
     """Train a single model to play as each agent in a zero-sum game environment using invalid action masking."""
     env = env_fn.env(**env_kwargs)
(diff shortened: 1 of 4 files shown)


deep_quoridor/src/train_sb3.py
```diff
@@ -213,15 +227,15 @@ def eval_action_mask(env_fn, num_games=100, render_mode=None, **env_kwargs):
...
+    env_kwargs = {"board_size": 5, "max_walls": 3}
 
     # Evaluation/training hyperparameter notes:
     # 10k steps: Winrate:  0.76, loss order of 1e-03
     # 20k steps: Winrate:  0.86, loss order of 1e-04
     # 40k steps: Winrate:  0.86, loss order of 7e-06
 
     # Train a model against itself (takes ~20 seconds on a laptop CPU)
-    train_action_mask(env_fn, steps=20_480, seed=0, **env_kwargs)
+    train_action_mask(env_fn, steps=10_480, seed=0, **env_kwargs)
```
alejandromarcu, line 238: just curious, any reason for using 10_480 vs a round number like 10_000?

alejandromarcu (approved):

deep_quoridor/src/train_sb3.py
adamantivm: no, just copied it from the example. I suppose it's becasue it runs in batches of 2048, so it's 10 batches? And then I wanted to run faster to do tests and I made it "kind of 10k".

deep_quoridor/src/train_sb3.py
```diff
@@ -121,6 +122,11 @@ def mask_fn(env):
     return env.action_mask()
 
 
+def make_model_id(env_kwargs):
+    # HACK: This is takenfrom trainable_agent model_id()
+    return f"sb3ppo_B{env_kwargs['board_size']}W{env_kwargs['max_walls']}_mv0"
```
alejandromarcu, line 127: If I understand correctly this is because we don't have an instance of the model here, no? my worry is that it may get out of sync (e.g. you hardcoded the model version to 0 here).  Maybe we can just instantiate the model class without a filename and wandb so that nothing is loaded (should be pretty light) just to get the model id?
Otherwise don't worry, we can leave it like this

alejandromarcu (approved):

deep_quoridor/src/train_sb3.py
adamantivm: ok, I went ahead and started using an instance to get the name, so that it's all in a single place and the verison name is not hardcoded.
I had to move more stuff around than I wanted to do it, but I think it's a small step in the right direction.

PR #157, opened 2025-04-08
WIP: Fix ruff errors and add ruff to github actions

alejandromarcu:


.vscode/settings.json
```diff
@@ -25,7 +25,4 @@
     ],
     "ruff.trace.server": "messages",
     "python.sortImports.args": ["--profile", "black"],
-    "editor.codeActionsOnSave": {
```
adamantivm, line 28: Do we really want to remove organize imports on save? I liked that

muralx: This is odd. Before originals cucu's change (the one adding python.SortImports, and editor.codeAction..),  imports were being organized after I added (editor.formatOnSave) under python section. After his change to add a new import organizer, there is a discrepancy in the way imports are done between his setup and mine.

Also, when I added a base config for ruff, I explicitly disabled some warnings, since I thought they might be too intrusive.

PR #158, opened 2025-04-09
Use Quoridor class for game logic

jonbinney:
... in the petting zoo environment.

Good things:
- Tests pass! `quoridor_test.py` now tests both the QuoridorEnv and the Quoridor class at the same time.
- Using the classes and enums from quoridor.py should be a bit more readable than using numbers
- We now have the combined wall and player positions grid from the Quoridor class that might be a good representation to give to a neural net (@muralx maybe interesting to try with a conv net that can use the 2d structure?)

Meh things:
- Currently `profile_observe.py` takes about 4x as long as the old implementation on my computer. I think I can close that gap and make the new implementation faster in the next couple of days, but I didn't want to spend more time optimizing before getting this PR merged. For now we're using pretty much the same DFS code, which isn't yet optimized for the representation used in the `Board ` class.
- Some (all?) of the agents are broken. Especially the greedy agent since it made heavy use of the environment's API. I Started fixing this; will continue tomorrow. Unless @alejandromarcu has some time :-P . `QuoridorEnv._get_observation()` tries to provide the same arrays (walls, a grid of player positions, etc) so that the agents are easy to keep working.


deep_quoridor/src/quoridor.py
```diff
@@ -155,14 +191,17 @@ def is_wall_between(self, position1: Position, position2: Position) -> bool:
...
-            return self._grid[*wall_position] == Board.PLAYER1_WALL
+            wall_position = position1 + position2 + 2
+            return self._grid[*wall_position] == Board.WALL
         else:
             # By convention we treat the border as a "wall". This is makes checking jumps more convenient, since
             # players are allowed to jump diagonally if they are adjacent to another player and the border of the
             # board is on the other side of that player.
             return True
 
+    def get_old_style_walls(self):
```
jonbinney, line 202: For now the Board class keeps track of the walls in both the new combined grid as well as the `(board_size-1) x (board_size -1) x 2` grid so that agents can use whichever they want.

jonbinney:
I'll debug the agents so that `play.py` works again in the next couple of days. It'll be good for me to catch up on that part of the code.

deep_quoridor/src/quoridor_env.py
```diff
@@ -386,69 +268,42 @@ def observation_space(self, agent):
...
-        elif action_type == 1:  # Placing a vertical wall
-            return self.board_size**2 + row * self.wall_size + col
-        elif action_type == 2:  # Placing a horizontal wall
-            return self.board_size**2 + self.wall_size**2 + row * self.wall_size + col
+        if isinstance(action, MoveAction):
+            return action.destination[0] * self.board_size + action.destination[1]
+        elif isinstance(action, WallAction) and action.orientation == WallOrientation.VERTICAL:
+            return self.board_size**2 + action.position[0] * self.wall_size + action.position[1]
+        elif isinstance(action, WallAction) and action.orientation == WallOrientation.HORIZONTAL:
+            return self.board_size**2 + self.wall_size**2 + action.position[0] * self.wall_size + action.position[1]
```
alejandromarcu, line 280: I think it would be cleaner to define an index method in `Action` (with NotImplementedError) and override it in each subclass, something like this:

```
class MoveAction(Action):
 # etc
   @override
   def index(self, board_size, wall_size)
      return self.destination[0] * board_size + self.destination[1] 
```


Then, action_params_to_index would just return `action.index(self.board_size, self.wall_size)` (or you can even not declare action_params_to_index and use that every time)

deep_quoridor/src/quoridor.py
```diff
@@ -49,35 +50,49 @@ def is_valid_position_type(position: Position) -> bool:
...
-    """
-    Grid which has both walls and possible pawn positions.
-
-    Pawn positions are on even rows and columns, walls are on odd rows and columns. We use negative
-    values to represent free cells and walls, 0 for player 1, and 1 for player 2.
-    integers.
-    """
-
     # Possible values for each cell in the grid.
     OUT_OF_BOUNDS = -2  # Returned if a move is off of the board.
```
alejandromarcu, line 54: Should be removed, no?

deep_quoridor/src/quoridor.py
```diff
@@ -49,35 +50,49 @@ def is_valid_position_type(position: Position) -> bool:
...
+
+        # We also keep track of the walls in a (row x col x orientation) array so that we can give that
+        # representation to agents that expect it.
+        self._old_style_walls = np.zeros((self.board_size - 1, self.board_size - 1, 2), dtype=np.int8)
+
         self._players = [Player.ONE, Player.TWO]
         self._player_positions = [np.array([0, board_size // 2]), np.array([board_size - 1, board_size // 2])]
         for player, position in zip(self._players, self._player_positions):
-            self._grid[*position * 2] = player
+            self._grid[*(position * 2 + 2)] = player
```
alejandromarcu, line 80: I'm not a fan of position being an np array.  I know it lets you do some math, but I'd prefer to use a simple int,int tupe and use helper functions when needed, e.g. here it could be `self._set_player_cell(position, player)`.

deep_quoridor/src/quoridor.py
```diff
@@ -49,35 +50,49 @@ def is_valid_position_type(position: Position) -> bool:
...
         self._walls_remaining = [self.max_walls, self.max_walls]
 
+        self._potential_wall_neighbors = {
```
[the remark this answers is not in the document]
alejandromarcu: Also, why do you need np.array and not just lists? if you're just iterating a list would be enough

alejandromarcu (commented):
I still want to look a bit more, but I left some comments.  One thing that I don't like much is to use np.array for Position, I think a simple tuple would be nicer.
Also, I can help you to migrate the agents.
Does HumanAgent work? I think playing human vs human can be a good sanity test (you can do it by running `python src/play.py --players human human -r pygame`)

deep_quoridor/src/quoridor_env.py
jonbinney: The design now is that `quoridor.py` implements the game logic and then `quoridor_env.py` wraps it and knows about things like action masks and indexes into them. It seems icky to have `MoveAction` include a function which is specific to the way the Environment creates indices.

deep_quoridor/src/quoridor.py
jonbinney: I agree - since it would involve lots of little changes across `quoridor.py` I was avoiding including that change in this PR, but it's probably worth it. Do you like a tuple more than a `Position` class with row and col members?

deep_quoridor/src/quoridor_env.py
jonbinney: I've got an idea on how to do this nicely..... trying it out now.

jonbinney: ok, here's what I'm thinking:

I think the Action class should be tied to a specific action->index encoding - that seems too tightly coupled. But agents shouldn't need to use a member function of the environment to do the translation between actions and indices either - that also seems like too tight of a coupling. How about having an ActionEncoder class, which takes the board size as an argument at initialization time, and has two functions: `action_to_index(self, action)` and `index_to_action(self, action)`. Each agent would instantiate one and use it. 
   The environment would also use this class out of convenience. But really, the interface between agents and the environment should be limited to an shared agreement on what the observation arrays mean and what each action index does - the fact that they both use the Quoridor class and the ActionEncoder is an implementation detail, but an agent could choose to use some other implementation of the game, so long as it had the same idea of what each action index does.
  The tight coupling, where agents are passed the entire environment and can call its functions, has made this PR more difficult, so I think it would be good to remove it.  I'll add a way to construct a Quoridor object from an observation, rather than passing the entire Quoridor object in the observation, or having the agent access it through the environment. Then get_action would only need to take the observation, instead of taking the environment and calling `.last()` on it. I'm not sure whether the agents need to know about termination and truncation.... right now they just return `None` in those cases and do nothing... but that might not always be the case?

What do you think?

deep_quoridor/src/quoridor.py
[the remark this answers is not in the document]
jonbinney: Ah yes I'll add a comment. They're numpy arrays so that I can do slicing and math with them in the `_is_wall_potential_block` function. In this case it doesn't matter so much, but generally doing numpy operations on multiple values at once is much faster than iterating in python, since it all gets done in compiled C code in numpy.

deep_quoridor/src/quoridor.py
```diff
@@ -109,7 +124,14 @@ def get_player_cell(self, position: Position) -> int:
...
             raise ValueError(f"Position {position} is out of bounds")
 
-        return self._grid[*position * 2]
+        return self._grid[*position * 2 + 2]
+
+    def get_walls_remaining(self, player: Player) -> int:
+        """
+        Get the number of walls remaining for the player.
+        """
+        assert isinstance(player, Player)
```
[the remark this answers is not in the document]
jonbinney: Does the hint checker give some feedback in vscode? I might not have things configured correctly in my settings.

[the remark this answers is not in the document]
alejandromarcu: Yeah, it should give you, for example if I type

```
def meh(x : int):
    pass

meh("hey")
```
it underlines "hey" and says it's the wrong type. If it doesn't, try installing the pylance extension: https://marketplace.visualstudio.com/items/?itemName=ms-python.vscode-pylance

deep_quoridor/src/quoridor.py
alejandromarcu: I think the easiest would be just a simple tuple (r,c) (i.e `tuple[int, int]`).

deep_quoridor/src/quoridor_env.py
alejandromarcu: I like your idea of the ActionEncoder class, it will be elegant for decoupling things.  It's been frustrating that the agent receives an action_mask but it needs to understand what it means, or that it knows what move it wants to make but needs to encode it in an index. 
Also creating a Quoridor object from an observation would be great if it's not too complex to implement or too inefficient

alejandromarcu: BTW, I'm not sure about returning None, I'd think that since in the Arena we check for that case already, we won't encounter it in the agents, I don't know why we are using it TBH, maybe we got it from an example that was doing things differently? or I can be missing something

deep_quoridor/src/quoridor.py
jonbinney: A couple downsides with that:
- MoveAction, WallAction, Player are all classes. feels a bit odd for position to be treated differently
- can't do vector math, e.g. `position + offset`
- can't modify the `__str__(self)` function to pretty-print positions

But I'm not against it. I do think that change is outside of the scope of this PR though.

deep_quoridor/src/quoridor_env.py
jonbinney: Yeah i think you're right that the agents don't need to worry about termination/truncation and that can be handled int the arena class. I've pushed changes that do that with the Greedy agent, and `play.py` works fine.

deep_quoridor/src/quoridor.py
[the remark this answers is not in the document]
jonbinney: I do have pylance, but i've never seen that warning. Maybe i don't make that mistake? :-P
How would it catch more complicated problems though? Like:
```
x = None
if ...:
    x = right_type
else:
   x = wrong_type
foo(x)
```

jonbinney:
I've pushed my implementation of `ActionEncoder` and a few other things:
- agents take observation and action_mask in get_action, instead of taking the entire environment
- Quoridor class can now construct itself from an observation, so that agents that want to try out actions and see how they affect the state can (Like Greedy)
-The (row, col, action_type) representation is gone - instead we use either "action index" (the index into the aciton_mask) or the MoveAction and WallAction classes from quoridor.py
- the mapping between action index and action classes is now its own "ActionEncoder" class in quoridor.py, so that agents can convert back and forth without needing access to the entire QuoridorEnv class

The easiest way to see the changes is to look at the greedy agent's `get_action()` function in this PR. I've got it running again with `play.py` but there may still be subtle bugs hiding.

jonbinney:
Let me know what you think about those changes. Now I'm going to update the other agents.

deep_quoridor/src/quoridor.py
alejandromarcu: The thing with using a class is that instead of just use for example `(1,2)` you need to use `Position(1,2)`. But it's nice to be able to have functions for offset or __str__.  We can discuss this more after this PR

alejandromarcu:
I like the changes, it's cleaner using a Quoridor object than the mess of having the whole environment, thanks!

alejandromarcu:
It's looking good and tests are passing, yey! I guess dexp and ppo are missing, anything else? we probably should be running tests for them in CI, the hard part is making sure that the alias that we use is always working. e.g. we could use 'best', but if we bump the model version, we need to make sure that we create the tag immediately.  Or we could use `latest`, but there's the possibility that something fails and it would break the tests.

jonbinney:
Yeah, I need to update dexp and ppo, and make sure that I didn't break incremental rewards. Working on that now.

deep_quoridor/src/quoridor.py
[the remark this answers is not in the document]
jonbinney: I went ahead and removed the type checking asserts. I'll have to figure out why pylance doesn't catch passing arguments with the wrong type.

deep_quoridor/src/quoridor.py
jonbinney: ok i actually did go ahead and change from array to tuple. We can still use numpy functions, but have to be a bit careful because they change the type. For example `np.subtract(x, y)` returns an ndarray even if x and y are tuples.

jonbinney:
Almost there! Tests pass, agents run. Two problems:
- when running `python deep_quoridor/src/play.py -t 2 -yp` the human agent hangs. This might be because the agent its playing against hangs; i haven't dug into it yet.
- I get slightly different results when running the tournament, so something has changed.... a bit...

Here are the results of `python deep_quoridor/src/play.py -yi benchmark_B5W3 ` on this branch:
```
+------------+------------+--------+------------+--------+--------+
|  P1 \ P2   | dexp (111) | greedy | greedy-ish | sb3ppo | Total  |
+------------+------------+--------+------------+--------+--------+
| dexp (111) |     -      |  82%   |    78%     |  98%   |  86%   |
|   greedy   |     0%     |   -    |    98%     |  100%  |  66%   |
| greedy-ish |    22%     |  66%   |     -      |  100%  |  63%   |
|   sb3ppo   |     0%     |   0%   |     8%     |   -    |   3%   |
|   ======   |   ======   | ====== |   ======   | ====== | ====== |
|   Total    |     7%     |  49%   |    61%     |  99%   |  54%   |
+------------+------------+--------+------------+--------+--------+
```
And on `origin/main`:
```
+------------+------------+--------+------------+--------+--------+
|  P1 \ P2   | dexp (111) | greedy | greedy-ish | sb3ppo | Total  |
+------------+------------+--------+------------+--------+--------+
| dexp (111) |     -      |  94%   |    72%     |  98%   |  88%   |
|   greedy   |     4%     |   -    |    100%    |  100%  |  68%   |
| greedy-ish |    26%     |  74%   |     -      |  98%   |  66%   |
|   sb3ppo   |     2%     |   0%   |     4%     |   -    |   2%   |
|   ======   |   ======   | ====== |   ======   | ====== | ====== |
|   Total    |    11%     |  56%   |    59%     |  99%   |  56%   |
+------------+------------+--------+------------+--------+--------+
```

Anyone see something obvious I'm missing? If not I'll continue debugging tomorrow.

deep_quoridor/src/quoridor_env.py
```diff
@@ -27,20 +27,37 @@
...
 import random
-from typing import Optional, Tuple
+from typing import Optional
 
 import numpy as np
 from gymnasium import spaces
 from pettingzoo import AECEnv
 from pettingzoo.utils import wrappers
 
+from deep_quoridor.src.quoridor import Action, Board, MoveAction, Player, Quoridor, WallAction, WallOrientation
```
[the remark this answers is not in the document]
jonbinney: Fixed

deep_quoridor/src/quoridor.py
jonbinney: Yep, removed now.

deep_quoridor/src/agents/core/trainable_agent.py
```diff
@@ -94,6 +95,7 @@ def __init__(
         self.training_mode = params.training_mode
         self.params = params
         self.action_size = self._calculate_action_size()
```
muralx, line 97: Just a comment in this. The encoder really depends on the observations, env wrapper being used. It should not really depend on the internal implementation of the game.

deep_quoridor/src/agents/core/trainable_agent.py
```diff
@@ -94,6 +95,7 @@ def __init__(
...
+        self.action_encoder = ActionEncoder(self.board_size)
```
muralx, line 98: Just a comment in this. The encoder really depends on the observations, env wrapper being used. It should not really depend on the internal implementation of the game.

jonbinney: True - it only depends on the rest of `quoridor.py` because it uses the MoveAction and WallAction classes. I'd like to keep it out of `quoridor_env.py` because it really shouldn't depend on anything in that file either, but maybe it should get its own file?

jonbinney:
@adamantivm @muralx @alejandromarcu It's past time that I stop tweaking this trying to get everything to work perfectly. Here are the highlights:


- We now pass the observations in to each agents `get_action` function, rather than passing in the entire environment and letting them do whatever they want with it
- Same for the `handle_step_*` functions in the trainable agent
- there's a function `construct_game_from_observation()` in quoridor.py that agents can use to construct a Quoridor game object from an observation. This allows us to keep the agent interfaces clean while still having a convenient way for agents like Greedy to try out actions
- Tests pass, profiling script works, play and train work (at least with the various argument combos I've been trying.

And the two things that have been holding me back from merging:
- Judging by the decreased win percentages, I've broken dexp in some way while changing the API. I've mucked about with it but I really need @muralx 's help fixing it - sorry for the trouble!
- Using classes for the actions makes things clearer, but also adds overhead for python to create the objects. This is about 3x slower, which matters most for agents like greedy that look at the actions instead of just the action mask. I think we should go back to tuples at some point, but I think we'd be better off getting all the other changes in first.

deep_quoridor/src/arena.py
```diff
@@ -50,32 +50,61 @@ def _play_game(self, agent1: Agent, agent2: Agent, game_id: str) -> GameResult:
...
+                        observation_after_action=observation,
+                        reward=self.game.rewards[agent_id],
+                        action=None,
+                        agent_id=agent_id,
+                        done=True,
+                    )
                 break
 
-            action = int(agent.get_action(self.game))
+            assert (observation["action_mask"] == self.game.last_action_mask[agent_id]).all()
```
muralx, line 74: I am not sure this would work with env wrappers. It is probably better to add a method in the wrapper and the env and force wrappers that modify the action mask to store the last action mask after transforming it.

deep_quoridor/src/arena.py
```diff
@@ -104,7 +133,14 @@ def _play_games(self, players: list[str | Agent], times: int, mode: PlayMode):
...
             else:
                 agents.append(
-                    AgentRegistry.create_from_encoded_name(p, board_size=self.board_size, max_walls=self.max_walls)
+                    AgentRegistry.create_from_encoded_name(
+                        p,
+                        board_size=self.board_size,
+                        max_walls=self.max_walls,
+                        action_space=self.game.action_space(
+                            None
+                        ),  # Action space doesn't depend on the player so we just pass None
```
muralx, line 142: It might too, for instance. If I want to represent movement using (1,0) (-1,0) ... to indicate the direction of the movement, action space could be different. I think for now we don't have agent wrappers, so it won't break anything.

deep_quoridor/src/arena.py
```diff
@@ -50,32 +50,61 @@ def _play_game(self, agent1: Agent, agent2: Agent, game_id: str) -> GameResult:
...
                 break
 
-            action = int(agent.get_action(self.game))
+            assert (observation["action_mask"] == self.game.last_action_mask[agent_id]).all()
+
+            action = int(agent.get_action(observation["observation"], observation["action_mask"]))
 
             self.plugins.before_action(self.game, agent)
             self.game.step(action)
+            observations.append((agent_id, self.game.observe(agent_id), action, self.game.rewards[agent_id]))
```
muralx, line 80: I am not sure this is correct, in general we want to have (s_n, action, s_n+1, rewards), but in here we have (s_n+1, action, rewards).

deep_quoridor/src/arena.py
```diff
@@ -50,32 +50,61 @@ def _play_game(self, agent1: Agent, agent2: Agent, game_id: str) -> GameResult:
...
 
             if agent.is_trainable() and isinstance(agent, AbstractTrainableAgent):
-                agent.handle_step_outcome(observation, action, self.game)
+                agent.handle_step_outcome(
+                    observation_before_action=observations[-1][1],
```
muralx, line 84: Since last item added to the observations is s_n+1, this is not the obsertation_before_action, but after action ?

deep_quoridor/src/arena.py
```diff
@@ -50,32 +50,61 @@ def _play_game(self, agent1: Agent, agent2: Agent, game_id: str) -> GameResult:
...
+                    opponent_observation=self.game.observe(opponent_agent_id),
```
muralx, line 85: maybe make this name explicit as opponent_observation_after_action ?

deep_quoridor/src/arena.py
```diff
@@ -50,32 +50,61 @@ def _play_game(self, agent1: Agent, agent2: Agent, game_id: str) -> GameResult:
...
+                    agent_id=agent_id,
+                    done=self.game.is_done(),
+                )
 
-            opponent_agent = agents[self.game.agent_selection]
-            if opponent_agent.is_trainable() and isinstance(opponent_agent, AbstractTrainableAgent):
-                opponent_agent.handle_opponent_step_outcome(observation, action, self.game)
+            if opponent_agent.is_trainable() and isinstance(opponent_agent_id, AbstractTrainableAgent):
+                opponent_agent.handle_opponent_step_outcome(
+                    observation_before_action=observations[-1][1],
```
muralx, line 95: naming here will be confusing. Since we are calling the agent, we should name this from the POV of the agent being called.  So this should be opponent_observation_before_action, similar for the rest.

deep_quoridor/src/arena.py
```diff
@@ -50,32 +50,61 @@ def _play_game(self, agent1: Agent, agent2: Agent, game_id: str) -> GameResult:
...
+                    opponent_observation=self.game.observe(opponent_agent_id),
```
muralx, line 96: This could be my_observation_after_opponents_actions?

deep_quoridor/src/arena.py
```diff
@@ -50,32 +50,61 @@ def _play_game(self, agent1: Agent, agent2: Agent, game_id: str) -> GameResult:
         self.plugins.start_game(self.game, agent1, agent2)
         start_time = time.time()
         step = 0
-        for player_id in self.game.agent_iter():
+        observations = []  # list of tuples of (agent_id, observation, action)
+        for agent_id in self.game.agent_iter():
```
muralx, line 54: I don't think you need this. It is harder to see what's going on and in general you only need the current states and not past. See my other comment for game termination.

deep_quoridor/src/arena.py
```diff
@@ -50,32 +50,61 @@ def _play_game(self, agent1: Agent, agent2: Agent, game_id: str) -> GameResult:
...
-            agent = agents[player_id]
+            agent = agents[agent_id]
+            opponent_agent_id = self.game.get_opponent(agent_id)
+            opponent_agent = agents[agent_id]
+
             if termination or truncation:
                 if agent.is_trainable() and isinstance(agent, AbstractTrainableAgent):
                     # Handle end of game (in case winner was not this agent)
-                    agent.handle_step_outcome(observation, None, self.game)
+                    agent.handle_step_outcome(
```
muralx, line 63: For this case you can pass none for observation after_action, etc, since there is no Action in this  case. Maybe we can later refactor this to be a separate method that says, handle_opponent_win which only receives the info it needs.

adamantivm:
Just want to say thanks for the hard work. I realize I'm too out of the loop on the evolution of all our abstractions, including the agent superclasses, the env and the arena, so I'm not sure how much I'd be able to contribute with a review. 
I'll support whatever you all decide and don't mind if a merge breaks things, I'll deal with it.

deep_quoridor/src/arena.py
jonbinney: Yes i was wondering about this! Ok that makes sense.

deep_quoridor/src/arena.py
jonbinney: You mean the list of past observations? I added that to make it easier to figure out what went wrong when a game gets into a bad state and you catch it in the debugger. I'm ok with removing it though, and I think it repeats some information that is stored in other places.

jonbinney:
@muralx Thank you for your review! I completely forgot to consider the case of wrapped environments, so I'll dig into that tomorrow. My first (probably wrong) thought is: shouldn't it just work? If the environment is wrapped, then when `Arena._play_game()` calls `.observe()` on the wrapped environment, it will get the observation with any transformations applied by the environment wrappers. Then it will pass that observation into the agent's `.get_action()` function, and the when the agent constructs a Quoridor object from it, it will be of the transformed game. Or do the transformations do things that lead to observations that don't actually represent a game anymore?

muralx:
@jonbinney I think the observation could be any format, or orientation, so it would be hard for Quoridor game to construct itself from it. Current agents that are using the Quoridor might anyway depends in a particular observation, so things might still work. But any other agent depending on a different observation style, would need some transformer to convert the observation back to the quoridor game accepted format. Maybe not important for now, but once I add the agent adapters it will become more important if game interaction is needed from the agent.

deep_quoridor/src/agents/core/trainable_agent.py
```diff
@@ -155,23 +157,58 @@ def get_opponent_player_id(self, player_id):
...
+    def adjust_reward(self, r, done):
+        if done:
             r *= self.final_reward_multiplier
         return r
 
-    def handle_step_outcome(self, observation_before_action, action, game):
+    def handle_step_outcome(
+        self,
+        observation_before_action,
+        opponent_observation,
```
muralx, line 179: maybe also name it as opponent_observation_after_action

deep_quoridor/src/agents/core/trainable_agent.py
```diff
@@ -155,23 +157,58 @@ def get_opponent_player_id(self, player_id):
...
+        observation_after_action,
+        reward,
+        action,
+        agent_id,
```
muralx, line 183: This should not be passed ideally. Since this is the outcome self, self.player_id should be use internally.

deep_quoridor/src/agents/core/trainable_agent.py
```diff
@@ -185,24 +222,24 @@ def handle_step_outcome_all(self, observation_before_action, action, game, playe
                     return reward
             return 0
 
-        state_before_action = self.observation_to_tensor(observation_before_action, player_id)
-        state_after_action = self.observation_to_tensor(game.observe(player_id), player_id)
-        opponent_state = game.observe(self.get_opponent_player_id(player_id))
+        state_before_action = self.observation_to_tensor(observation_before_action["observation"], agent_id)
+        state_after_action = self.observation_to_tensor(observation_after_action["observation"], agent_id)
+        # opponent_state = game.observe(self.get_opponent_player_id(player_id))
+        opponent_state = opponent_observation
```
muralx, line 228: nit, just use opponent_observation.

deep_quoridor/src/agents/core/trainable_agent.py
muralx, line 183: nit: note this is the player_id for the game and not the id of the agent. If we prefer using agent id, maybe it will be better to rename all other variables and methods to agent_id too. Though I think an id is kind of fixed, and should not change from match to match.

deep_quoridor/src/agents/core/trainable_agent.py
```diff
@@ -259,19 +296,13 @@ def update_target_network(self):
         """Copy parameters from online network to target network."""
         self.target_network.load_state_dict(self.online_network.state_dict())
 
-    def get_action(self, game):
+    def get_action(self, observation, action_mask):
```
muralx, line 299: is this action mask also in the observation? Note that the use of observation above includes the action mask. Not sure how to name the two version of observations so we can understand what we have in each one..

deep_quoridor/src/agents/core/trainable_agent.py
muralx: I guess my point was similar to the other comments. If observation is changed, which currently should be handled OK by the agent (assuming wrapper has the action encoder function), it won't no longer work after this change since action encoder is different depending on the transformations being used.

muralx (commented):
Thanks @jonbinney, I left some additional comments. Many of them are around the transformations, but you can add TODOs in there so we cover those when we add adapter or find a way of using per agent env wrappers.

I think there are a few things about naming that would be good to address, and some actual potential bug that I highlighted in the previous review (I might be missing something though).

I did not review greedy or simple (I will let @alejandromarcu to check those since he is more familiar with the code)

deep_quoridor/src/arena.py
muralx: I thought you were using it for keeping state in between steps. 

It does add some readability challenges into what is being passed to the methods, so maybe if you want to keep it like this, you could extract the values to variables so it is easier to see the intention of the code and detect potential future bugs.

deep_quoridor/src/agents/core/trainable_agent.py
jonbinney: This observations is `observation["observation"]`, which doesn't include the action mask. I don't like that pettingzoo allows environments to put the action mask in either the observation or the info dict. Agree that `observation["observation"]` is a confusing naming convention.

deep_quoridor/src/agents/core/trainable_agent.py
jonbinney: Ah, yes that's a problem.

deep_quoridor/src/agents/core/trainable_agent.py
jonbinney: Makes sense. `handle_step_outcome_all` does need to have the agent_id passed in though, right?

Side note: i've started trying to be more consistent about using `agent_id` for the string id of the agent, and `player` terminology within the Quoridor class.

deep_quoridor/src/arena.py
jonbinney: In that case is`s_n+1` the state after this action, or the state after this action and the opponent's action?

jonbinney:
@muralx I understand what you mean now about wrappers that modify the observation breaking my approach. I think the best way forward is for me to keep the part of this PR that uses the Quoridor class within the environment, but to revert the changes to the `Agent.get_action()` and `Agent.handle_step_outcome*` functions (go back to passing the environment in. Does that sound reasonable?

deep_quoridor/src/agents/core/trainable_agent.py
muralx: yes, since it is called from two different pov. Fine with any naming convention, lets just keep them consistent.

muralx:
@jonbinney I think nothing will break for now, and I will be adding per agent adapter or environment, so I will need to change that code anyway, so for me it is OK for you to submit and we can refactor things after when we add support for use of wrappers (only one using wrapper now is sb3 for training purposes).

deep_quoridor/src/agents/greedy.py
```diff
@@ -175,35 +180,36 @@ def _log_action(
...
                 self.action_log.clear()
                 # TO DO: when we have the functionality in the log to output a message,
                 # use it to say that the move will be random.
-            return game.action_space(game.agent_selection).sample(observation["action_mask"])
+            return self.action_space.sample(action_mask)
 
-        goal_row = game.board_size - 1 if self.player_id == "player_0" else 0
-        opponent_row = game.board_size - 1 - goal_row
+        goal_row = self.board_size - 1 if self.player_id == "player_0" else 0
+        opponent_goal_row = self.board_size - 1 - goal_row
```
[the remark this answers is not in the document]
jonbinney: Fixed now

deep_quoridor/src/arena.py
jonbinney: True. I've updated the comment to remind us to fix this if we add agent wrappers.

deep_quoridor/src/agents/core/trainable_agent.py
jonbinney: As discussed below I'll leave this for now and we can change it when we start using wrappers.

deep_quoridor/src/arena.py
jonbinney: As discussed below, I'll leave this in this PR and we can change it when we use wrappers.

deep_quoridor/src/agents/core/trainable_agent.py
jonbinney: Done

deep_quoridor/src/arena.py
jonbinney: ok, i understand. I've changed this to store `(s_n, action, s_n+1, rewards),`

jonbinney: Actually I'm removing this `observations` list - you're right that it makes things less readable.

deep_quoridor/src/arena.py
jonbinney: I think you're right. I think I've fixed this during my switch back to individual variables instead of a big list.

deep_quoridor/src/arena.py
jonbinney: Good idea; done.

deep_quoridor/src/arena.py
jonbinney: Done

deep_quoridor/src/arena.py
jonbinney: Agreed about the readability problems; i've switched to individual variables instead of a list.

deep_quoridor/src/arena.py
jonbinney: Done

deep_quoridor/src/arena.py
jonbinney: I ended up using individual variables instead of a list.

deep_quoridor/src/agents/core/trainable_agent.py
jonbinney: Done

deep_quoridor/src/agents/core/trainable_agent.py
jonbinney: Makes sense. I'm leaving it alone in this PR, but we will need to think a bit about what convention to use.

jonbinney:
@muralx I made the naming changes you suggested and replaced the observations list with individual variables for clarity. It looks like I've still broken something in dexp, since its win rate has gone way down. If you're ok with merging and fixing in another PR, I think that might make things simpler.

alejandromarcu:
I'll take a look today or tomorrow, thanks guys!

alejandromarcu:
I just tried training dexp with the same parameters than I did in main (and I merged main here), and the metrics were significantly different. Are we ok with this?


image


Training cmd:
```
/Users/amarcu/code/deep_rabbit_hole/.venv/bin/python /Users/amarcu/code/deep_rabbit_hole/deep_quoridor/src/train.py  -N 5 -W 3 -e 600 -i 43 -p dexp:rotate=true,split=true,turn=false,epsilon=1,epsilon_decay=0.9985,epsilon_min=0.1,final_reward_multiplier=20,training_mode=true,use_opponents_actions=true,update_target_every=100,gamma=0.95,target_as_source_for_opponent=true,use_negative_qvalue_function=true,batch_size=128,mask_targetq=true,nick=dexp_train greedy:p_random=0.3 greedy:p_random=0.1 simple -rp cucu -s
```

muralx:
If only dexp is the problem, I will take care of it once it is merged.
Thanks for double checking

deep_quoridor/src/agents/human.py
```diff
@@ -25,21 +27,22 @@ def name(self):
...
+                    valid_moves.add((valid_action.destination[0], valid_action.destination[1], 0))
+                elif isinstance(valid_action, WallAction):
+                    valid_moves.add(
+                        (valid_action.position[0], valid_action.position[1], valid_action.orientation.value - 1)
+                    )
 
         result = PygameQuoridor.instance().get_human_input(valid_moves)
 
-        return None if result is None else game.action_params_to_index(*result)
+        return None if result is None else self.action_encoder.action_to_index(*result)
```
alejandromarcu, line 48: When running the UI with human this fails:

```
(.venv) (base) amarcu@MacBookPro deep_rabbit_hole % /Users/amarcu/code/deep_rabbit_hole/.venv/bin/python /Users/amarcu/code/deep_rabbit_hole/deep_quoridor/src/play.py -yi human_vs_greedy                                                                                              
====< Running human_vs_greedy >====
Exception in thread Thread-1 (_play_games):
Traceback (most recent call last):
  File "/opt/homebrew/Cellar/python@3.12/3.12.9/Frameworks/Python.framework/Versions/3.12/lib/python3.12/threading.py", line 1075, in _bootstrap_inner
    self.run()
  File "/opt/homebrew/Cellar/python@3.12/3.12.9/Frameworks/Python.framework/Versions/3.12/lib/python3.12/threading.py", line 1012, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/amarcu/code/deep_rabbit_hole/deep_quoridor/src/arena.py", line 160, in _play_games
    result = self._play_game(agent_1, agent_2, f"game_{match_id:04d}")
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/amarcu/code/deep_rabbit_hole/deep_quoridor/src/arena.py", line 75, in _play_game
    agent.get_action(observation_before_action["observation"], observation_before_action["action_mask"])
  File "/Users/amarcu/code/deep_rabbit_hole/deep_quoridor/src/agents/human.py", line 48, in get_action
    return None if result is None else self.action_encoder.action_to_index(*result)
                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: ActionEncoder.action_to_index() takes 2 positional arguments but 4 were given
```

The problem is that result is a tuple of 3 elements (row, col, type) and we're expecting an Action.  Is there a way to convert those 3 elements into an Action or we don't use them anymore?  Otherwise, in pygame.py we should change the `handle_hover` method to create an Action (I hope this would be straightforward), and change the type of hover_action and click_action to be Action

alejandromarcu (approved):
Looks good! If you want just merge and I'll send a fix for the human player.

jonbinney:
Thanks guys! Merging!

PR #159, opened 2025-04-09
Training against multiple agents, fix action for rotation

muralx:
Uses the list of agents to do [0] vs [1..n-1] randomly.

### DIFF
--- a/deep_quoridor/src/arena.py
+++ b/deep_quoridor/src/arena.py
@@ -1,4 +1,6 @@
+import random
 import time
+from enum import Enum
 from threading import Thread
 from typing import Optional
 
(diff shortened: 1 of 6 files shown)


deep_quoridor/src/arena.py
```diff
@@ -95,11 +100,37 @@ def _play_games(self, players: list[str | Agent], times: int):
...
+                        result = self._play_game(agent_1, agent_2, f"game_{match_id:04d}")
+                        results.append(result)
+                        match_id += 1
+        else:  # FIRST_VS_RANDOM mode
+            first_agent = agents[0]
+            remaining_agents = agents[1:]
+
+            for _ in range(times):
+                for _ in range(len(remaining_agents)):
+                    opponent = random.choice(remaining_agents)
```
alejandromarcu, line 129: if the players are A B C and D, then A will play with 3 random opponents, e.g. it may play with B 2 times and once with C in a round.  Is that what you wanted? or if you just wanted to use the 3 agents in a different order you could use random.shuffle

alejandromarcu (approved):

deep_quoridor/src/arena.py
muralx: I wanted to just randomly play with different agents, since times usually is in the 1000s this works fine. If we also want to use this for other purposes, we can do what you suggest (with times = 1 for instance) we can change it

PR #161, opened 2025-04-10
Fix representation of boards and turns

muralx:
Code was using my_turn to select player 1 or 2 from the board, same for using my walls vs opponent walls. New code uses always 1  from the board observation since observations are always generated for the player's move we are processing.

Also remove the inversion of remaining walls, which is likely incorrect.


alejandromarcu (approved):

muralx:
Submitting this basic fix. The new one will go in a new Agent since it is too big and I want to keep this one.

PR #162, opened 2025-04-10
Some tweaks to PPO

adamantivm:
Nothing that changes things much, but it all seems to still work out.

Most importantlly, incorporates two tricks developed by @muralx in the form of Wrappers
 - Split boards (as dict)
 - Rotated board

It also includes some unrelated refactoring:
 - Pushed more of the agent details to the PPO agent (instead of the train file)
 - Parameterized the training file
 - Made the environment wrapping a method so that it's in a single placed and used in both agent play and training

I think the evaluation part of train_sb3 is wrong in the case of the rotated board (how does it not crash??) but that seems not too critical.

### DIFF
--- a/deep_quoridor/src/environment/dict_split_board_wrapper.py
+++ b/deep_quoridor/src/environment/dict_split_board_wrapper.py
@@ -1,4 +1,5 @@
 import numpy as np
+from gymnasium import spaces
 from pettingzoo.utils.env import AgentID, ObsType
 from pettingzoo.utils.wrappers import BaseWrapper
 
@@ -28,18 +29,24 @@ class DictSplitBoardWrapper(BaseWrapper):
 
         return {"observation": observation, "action_mask": obs["action_mask"]}
 
-    def observation_space(self):
+    def observation_space(self, agent):
         """Define the observation space for the transformed observations."""
-        original_space = self.env.observation_space
+        original_space = self.env.observation_space(agent)
         board_shape = (self.board_size, self.board_size)  # Shape for each board (player and opponent)
         return {
-            "observation": {
-                "my_turn": original_space["observation"]["my_turn"],
-                "my_board": original_space["observation"]["board"].__class__(shape=board_shape, dtype=np.float32),
-                "opponent_board": original_space["observation"]["board"].__class__(shape=board_shape, dtype=np.float32),
-                "walls": original_space["observation"]["walls"],
-                "my_walls_remaining": original_space["observation"]["my_walls_remaining"],
-                "opponent_walls_remaining": original_space["observation"]["opponent_walls_remaining"],
-            },
+            "observation": spaces.Dict(
+                {
+                    "my_turn": original_space["observation"]["my_turn"],
+                    "my_board": original_space["observation"]["board"].__class__(
+                        0, 1, shape=board_shape, dtype=np.float32
+                    ),
+                    "opponent_board": original_space["observation"]["board"].__class__(
+                        0, 1, shape=board_shape, dtype=np.float32
+                    ),
+                    "walls": original_space["observation"]["walls"],
+                    "my_walls_remaining": original_space["observation"]["my_walls_remaining"],
+                    "opponent_walls_remaining": original_space["observation"]["opponent_walls_remaining"],
+                }
+            ),
             "action_mask": original_space["action_mask"],
         }
--- a/deep_quoridor/src/train_sb3.py
+++ b/deep_quoridor/src/train_sb3.py
@@ -18,35 +18,15 @@ import os
 import time
 
 import quoridor_env
-import torch
-from agents.sb3_ppo import SB3ActionMaskWrapper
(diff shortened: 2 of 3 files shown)


deep_quoridor/src/environment/dict_split_board_wrapper.py
```diff
@@ -28,18 +29,24 @@ def observe(self, agent: AgentID) -> ObsType:
...
-                "opponent_board": original_space["observation"]["board"].__class__(shape=board_shape, dtype=np.float32),
-                "walls": original_space["observation"]["walls"],
-                "my_walls_remaining": original_space["observation"]["my_walls_remaining"],
-                "opponent_walls_remaining": original_space["observation"]["opponent_walls_remaining"],
-            },
+            "observation": spaces.Dict(
+                {
+                    "my_turn": original_space["observation"]["my_turn"],
+                    "my_board": original_space["observation"]["board"].__class__(
+                        0, 1, shape=board_shape, dtype=np.float32
```
muralx, line 41: nice addition

deep_quoridor/src/train_sb3.py
```diff
@@ -82,10 +62,12 @@ def train_action_mask(env_fn, steps=10_000, seed=0, **env_kwargs):
...
     model.save(local_filename)
-    artifact = wandb.Artifact(f"{model_id}", type="model")
-    artifact.add_file(local_path=local_filename)
-    artifact.save()
-    wandb.finish()
+
+    if upload_to_wandb:
+        artifact = wandb.Artifact(f"{model_id}", type="model")
+        artifact.add_file(local_path=local_filename)
+        artifact.save()
```
muralx, line 69: I think that using wandb.log_artifact(artifact).wait() allows you to get the version from the artifact as artifact.version(), so you can print it, which is helpful to load the model for play later on.

(I have that change from trainable agent, not yet in a PR)

deep_quoridor/src/train_sb3.py
```diff
@@ -158,20 +139,52 @@ def eval_action_mask(env_fn, num_games=100, render_mode=None, **env_kwargs):
...
+    parser = argparse.ArgumentParser(
+        description="Train and evaluate SB3 agents in Quoridor using invalid action masking"
+    )
+    parser.add_argument("-N", "--board_size", type=int, default=3, help="Board Size")
+    parser.add_argument("-W", "--max_walls", type=int, default=0, help="Max walls per player")
+    parser.add_argument("-e", "--steps", type=int, default=20_480, help="Number of steps to train for")
+    parser.add_argument("-g", "--num_games", type=int, default=100, help="Number of games for evaluation")
+    parser.add_argument("-i", "--seed", type=int, default=0, help="Random seed for training and evaluation")
+    parser.add_argument("--no-train", action="store_true", default=False, help="Skip training and only run evaluation")
+    parser.add_argument("--no-upload", action="store_true", default=False, help="Skip training and only run evaluation")
```
muralx, line 151: fix help

deep_quoridor/src/train_sb3.py
```diff
@@ -158,20 +139,52 @@ def eval_action_mask(env_fn, num_games=100, render_mode=None, **env_kwargs):
...
+    parser.add_argument("--no-eval", action="store_true", default=False, help="Skip evaluation and only run training")
```
muralx, line 152: same here

deep_quoridor/src/train_sb3.py
```diff
@@ -94,12 +76,11 @@ def train_action_mask(env_fn, steps=10_000, seed=0, **env_kwargs):
     env.close()
 
 
-def eval_action_mask(env_fn, num_games=100, render_mode=None, **env_kwargs):
+def eval_action_mask(env_fn, num_games=100, render_mode=None, player=0, **env_kwargs):
```
muralx, line 79: why not use the arena for this? I imagine eval could be easily done by running games in the arena. If there was some data/feedback missing from the arena, it might be good to add it there.

muralx (approved):
LGTM

deep_quoridor/src/train_sb3.py
adamantivm: Yeah, it's just because it was auto-generated originally and it comes handy.
In a future pass this should totally be replaced by an Arena call.
TODO for a next PR.

deep_quoridor/src/train_sb3.py
```diff
@@ -136,7 +117,7 @@ def eval_action_mask(env_fn, num_games=100, render_mode=None, **env_kwargs):
                 round_rewards.append(env.rewards)
                 break
             else:
-                if agent == env.possible_agents[0]:
+                if agent == env.possible_agents[1 - player]:
```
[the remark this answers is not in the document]
adamantivm: I didn't go into deep detail to understand why but, for some reason, if I do that then this crashes with "invalid action, forbidden by action mask". 
I can't understand (without debugging) why ...

PR #163, opened 2025-04-10
Adding new options to DExp

muralx:
- Use substraction for the target qfunction (option use_negative_qvalue_function, default false)
- Supporting target_as_source_for_opponent, to make target states to match next step source step. This make the target state of the player 0/1's s2  (eg: (r1, s1, s2, a1)) to match the source state of player 1/0 (r2, s2, s3, a2) for all states. This can be combined with target_as_source_for_opponent and use_opponents_actions.

### DIFF
--- a/deep_quoridor/src/agents/dexp.py
+++ b/deep_quoridor/src/agents/dexp.py
@@ -71,12 +71,29 @@ class DExpAgentParams(TrainableAgentParams):
 class DExpAgent(AbstractTrainableAgent):
     """Diego experimental Agent using DRL."""
 
+    def check_congiguration(self):
+        """
+        Check the configuration of the agent.
+        This is used to check if the agent is configured correctly.
+        """
+        if self.params.use_opponents_actions and self.params.assign_negative_rewards:
+            raise ValueError(
+                "use_opponents_actions and assign_negative_rewards cannot be used together. "
+                "use_opponents_actions is used for training only, not for playing."
+            )
+        if self.params.target_as_source_for_opponent and self.params.assign_negative_rewards:
+            raise ValueError(
+                "target_as_source_for_opponent and assign_negative_rewards cannot be used together. "
+                "target_as_source_for_opponent is used for training only, not for playing."
+            )
+
     def __init__(
         self,
         params=DExpAgentParams(),
         **kwargs,
     ):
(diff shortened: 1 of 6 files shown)


deep_quoridor/src/plugins/wandb_train.py
```diff
@@ -64,8 +64,9 @@ def end_arena(self, game, results):
         self.agent.save_model(save_file)
 
         artifact = wandb.Artifact(f"{self.agent.model_id()}", type="model", metadata=asdict(self.agent.params))
-        artifact.add_file(local_path=str(save_file))
```
alejandromarcu, line 67: why are you removing the model file?

deep_quoridor/test/agents/dexp_test.py
```diff
@@ -0,0 +1,105 @@
...
+            "my_turn": True,
+        }
+    }
+
+
+@pytest.fixture
+def base_agent():
+    params = DExpAgentParams(rotate=False, turn=False, split=False, training_mode=True)
+    agent = DExpAgent(params=params, board_size=3, max_walls=5)
+    agent.device = torch.device("cpu")
```
alejandromarcu, line 24: Why are you doing this? it should be detecting the best device automatically

deep_quoridor/src/agents/dexp.py
```diff
@@ -56,8 +55,13 @@ class DExpAgentParams(TrainableAgentParams):
 
     # Whether oppoments actions are used for training
     # This is used for training only, not for playing
+    # This should not be combined with assign_negative_rewards
```
alejandromarcu, line 58: If it's easy, can you put an assert somewhere (E.g. in the constructor of DexpAgent) to make sure it doesn't happen?

alejandromarcu (approved):
Please look at my comments before merging, you may be removing the model upload to wandb if I understand correctly

deep_quoridor/src/plugins/wandb_train.py
muralx: good catch, not sure how that happened.

deep_quoridor/test/agents/dexp_test.py
muralx: It does not work if it detects cuda. It works with device CPU.

PR #164, opened 2025-04-11
yargs: a command line augmentation using yaml

alejandromarcu:
Passing all the arguments to play.py or train.py is getting out of control.
This allows us to store configurations in yaml file, and then we can run them just by referencing the ids or the file name.
We could easily incorporate this in train.py.
We can also extend it in the future, e.g.:
- adding tags, so you can run all the entries with a certain tag
- allowing to override parameters from the command line

For example, with the current setup, you could play human vs greedy just by doing:
```
 python deep_quoridor/src/play.py -yi human_vs_greedy
```

I also started collecting our "benchmarks" (we can later log them in wandb) in the `agent_benchmark.yaml` file, so now you can run them all as follows:

```
deep_rabbit_hole % python deep_quoridor/src/play.py -yp agent_benchmark.yaml
====< Running benchmark_B5W3 >====
Loading pre-trained model from /Users/amarcu/code/deep_rabbit_hole/wandbmodels/dexp_B5W3_mv1_6c57b.pt
Progress: [========------------------------------------------] 16.7% (100/600)Loaded model from /Users/amarcu/code/deep_rabbit_hole/wandbmodels/sb3ppo_B5W3_mv0_cd5f0.zip
/home/julian/aaae/deep-rabbit-hole/code/deep_rabbit_hole/deep_quoridor/src/train_sb3.py:45: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
Progress: [==================================================] 100.0% (600/600)
Arena stats
===========
Board size:  5
Max walls:  3
Step rewards:  False
Total number of games:  600

+------------+------------+--------+------------+--------+--------+
|  P1 \ P2   | dexp (111) | greedy | greedy-ish | sb3ppo | Total  |
+------------+------------+--------+------------+--------+--------+
| dexp (111) |     -      |  86%   |    74%     |  100%  |  87%   |
|   greedy   |     8%     |   -    |    100%    |  100%  |  69%   |
| greedy-ish |    20%     |  74%   |     -      |  100%  |  65%   |
|   sb3ppo   |     2%     |   0%   |     0%     |   -    |   1%   |
|   ======   |   ======   | ====== |   ======   | ====== | ====== |
|   Total    |    10%     |  53%   |    58%     |  100%  |  55%   |
+------------+------------+--------+------------+--------+--------+
====< Running benchmark_B9W10 >====
Loading pre-trained model from /Users/amarcu/code/deep_rabbit_hole/wandbmodels/dexp_B9W10_mv1_377a4.pt
Progress: [==================================================] 100.0% (300/300)
Arena stats
===========
Board size:  9
Max walls:  10
Step rewards:  False
Total number of games:  300

+------------+------------+--------+------------+--------+
|  P1 \ P2   | dexp (000) | greedy | greedy-ish | Total  |
+------------+------------+--------+------------+--------+
| dexp (000) |     -      |   0%   |     0%     |   0%   |
|   greedy   |    100%    |   -    |    94%     |  97%   |
| greedy-ish |    100%    |  22%   |     -      |  61%   |
|   ======   |   ======   | ====== |   ======   | ====== |
|   Total    |    100%    |  11%   |    47%     |  53%   |
+------------+------------+--------+------------+--------+
```

### DIFF
--- a/deep_quoridor/src/utils/misc.py
+++ b/deep_quoridor/src/utils/misc.py
@@ -1,10 +1,13 @@
+import argparse
 import random
+from glob import glob
 from pathlib import Path
 from typing import Optional
 
 import gymnasium.utils.seeding
 import numpy as np
 import torch
+import yaml
 
 
 def resolve_path(dir: str, filename: Optional[str] = None) -> Path:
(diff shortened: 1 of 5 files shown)


deep_quoridor/src/utils/misc.py
```diff
@@ -35,3 +38,81 @@ def my_device():
...
+
+    args = parser.parse_args()
+    if args.yargs_path is None and args.yargs_ids is None:
+        # Not using yargs
+        return {"": args}
+
+    if args.yargs_path is None:
+        path = Path(default_path)
+    else:
+        path = Path(default_path) / args.yargs_path
```
jonbinney, line 103: pathlib! nice!

deep_quoridor/src/utils/misc.py
```diff
@@ -1,10 +1,13 @@
+import argparse
 import random
+from glob import glob
 from pathlib import Path
 from typing import Optional
 
 import gymnasium.utils.seeding
 import numpy as np
 import torch
+import yaml
```
jonbinney, line 10: Thank you for using YAML! It is better than json in subtle but very nice ways..... I've worked with too many frontend devs that think they're the same.

deep_quoridor/src/utils/misc.py
```diff
@@ -35,3 +38,81 @@ def my_device():
...
+        # Execute all the entries in the file {default_path}/agent_benchmark.yaml
+        python myapp.py -yp agent_benchmark.yaml
+
+        # Execute the entries benchmark_B5W3 benchmark_B9W10 (should be found in the files in the default_path)
+        python myapp.py -yi benchmark_B5W3 benchmark_B9W10
+
+    """
+    parser.add_argument(
+        "-yp",
+        "--yargs_path",
```
jonbinney, line 81: nit: posix style for long args is to use dashes instead of underscores, so this would be `--yargs-path`: https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html

argparse is smart enough to convert these so that the resulting argument is `args.yargs_path` in python, so that the python variable follows python conventions and the command line arg follows posix conventions.

yargs/play/agent_benchmark.yaml
```diff
@@ -0,0 +1,5 @@
+benchmark_B5W3:
+  args: -t 100 -N 5 -W 3 --players greedy greedy:p_random=0.1,nick=greedy-ish sb3ppo:wandb_alias=latest dexp:wandb_alias=best
```
jonbinney, line 2: Looks good. Do you think we'll keep the `agent:variable=val,...` syntax now that we have yaml files? Or could those be more like:
```
args:
    agents:
           agent_1_nick:
                - agent_type: greedy
                - p_random: 0.1
                - ....
```
Or something similar? (not in this pr, though)
```

jonbinney (approved):

yargs/play/agent_benchmark.yaml
alejandromarcu: For the command line I'm planning to leave it like it is.  For the yaml file, I actually started doing something like the above but it got tricky because sometimes you have lists, dicts, single values, etc, so I aborted thinking that we may not need it, but I'd be happy to implement it, and still support the current format in yaml files, which has the advantage that you can just copy paste it from the command line

PR #165, opened 2025-04-15
Multiple minor fixes

muralx:
- Avoids loading training params when loading a model. During play those should not be used, and if training, new params should be passed.
- Fix negative rewards that was a TODO and was broken from previous PR. 
- Fix import that is using src.deep_quoridor to match current style being used.
- Fix training status renderer to skip agents if they are not in training mode.

### DIFF
--- a/deep_quoridor/src/agents/core/trainable_agent.py
+++ b/deep_quoridor/src/agents/core/trainable_agent.py
@@ -52,6 +52,7 @@ class TrainableAgentParams(SubargsBase):
 
     def training_only_params(cls) -> set[str]:
         """Returns a set of parameter names that are only used during training."""
+        # TODO: we should ideally have two set of params: one for training and one for inference
         return {
             "epsilon",
             "epsilon_min",
@@ -159,12 +160,11 @@ class AbstractTrainableAgent(Agent):
             ## TODO: Revisit this since it won't work for the case in which
             ## opponents actions are used
             if self.assign_negative_reward:
-                if self.assign_negative_reward:
-                    if len(self.replay_buffer) > 0:
-                        last = self.replay_buffer.get_last()
-                        last[2] = reward  # update final reward
-                        last[4] = 1.0  # mark as done
-                        return reward
+                if len(self.replay_buffer) > 0:
+                    last = self.replay_buffer.get_last()
+                    last[2] = reward  # update final reward
+                    last[4] = 1.0  # mark as done
+                    return reward
             return 0
 
         state_before_action = self.observation_to_tensor(observation_before_action, player_id)
--- a/deep_quoridor/src/agents/core/trainable_agent.py
+++ b/deep_quoridor/src/agents/core/trainable_agent.py
@@ -50,6 +50,23 @@ class TrainableAgentParams(SubargsBase):
     # If True, the target q-value function will substract maxq_a'(s', a') instead of adding it
     use_negative_qvalue_function: bool = False
 
+    def training_only_params(cls) -> set[str]:
+        """Returns a set of parameter names that are only used during training."""
+        return {
+            "epsilon",
+            "epsilon_min",
+            "epsilon_decay",
+            "gamma",
+            "batch_size",
+            "update_target_every",
+            "assign_negative_reward",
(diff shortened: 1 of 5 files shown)


deep_quoridor/src/agents/core/trainable_agent.py
```diff
@@ -141,12 +158,14 @@ def handle_step_outcome_all(self, observation_before_action, action, game, playe
         if action is None:
             ## TODO: Revisit this since it won't work for the case in which
             ## opponents actions are used
-            if self.assign_negative_reward and len(self.replay_buffer) > 0:
-                last = self.replay_buffer.get_last()
-                last[2] = reward  # update final reward
-                last[4] = 1.0  # mark as done
-                self.current_episode_reward += reward
-            return reward
+            if self.assign_negative_reward:
```
jonbinney, line 161: Does this change fix the above TODO? Should that be removed now?

deep_quoridor/src/agents/core/trainable_agent.py
```diff
@@ -50,6 +50,23 @@ class TrainableAgentParams(SubargsBase):
...
+            "gamma",
+            "batch_size",
+            "update_target_every",
+            "assign_negative_reward",
+            "training_mode",
+            "final_reward_multiplier",
+            "use_negative_qvalue_function",
+            "wandb_dir",
+            "model_dir",
+        }
```
jonbinney, line 68: Whitelisting individual parameters seems brittle - would it be better to have all traning parameters be of the form `training.foo` so we can ignore them en masse?

deep_quoridor/src/agents/core/trainable_agent.py
adamantivm: also, is this an LLM hallucination? The if looks repeated to me

deep_quoridor/src/agents/core/trainable_agent.py
adamantivm: +1 to this comment, but then again it's OK IMHO to leave this as a TODO

adamantivm:
IMHO the only mandatory fix is the repeated if, then it's OK to merge.

deep_quoridor/src/agents/core/trainable_agent.py
jonbinney: Agreed

deep_quoridor/src/agents/core/trainable_agent.py
muralx: @jonbinney  TODO is not fixed yet, but I am restricting using negative_rewards and  use opponent's actions in DExp, I don't think it makes sense to apply negative rewards if we are discounting the positive reward of the opponent.

@adamantivm That is my bad merge skill :)

deep_quoridor/src/agents/core/trainable_agent.py
muralx: We could do that, but it requires storing and passing to param set, which is a larger change. I am adding TODO since we talked with Cucu about it.

muralx:
Done, thanks for the review.

deep_quoridor/src/agents/core/trainable_agent.py
alejandromarcu: I wanted to do something like:

```
   @training_only
   epsilon: float
```
But chatgpt told me is not possible :'(

@jonbinney, not sure wht you mean by `training.foo`, since we can't use dot, would it be renaming the parameters to start with `training` or something along those lines?
In any case, I agree that we can merge this and take a look later.

jonbinney: Yeah i was thinking about prefixing the name with `training.`; as you point out that the `.` is illegal so maybe `traning_`? But not for this PR

muralx: @alejandromarcu  I suggested that option to copilot (if feasible) but it went the set definition.
@jonbinney I understood you wanted two different params classes, so we can add them to the agent as play_params, training_params. Using prefix just for the name will make the command line for training pretty long.

PR #166, opened 2025-04-17
Use a larger NN for testing purposes on DExp.

muralx:
Also
- adds additional params to be ignored when loading from wandb
- Add a config for DExp benchmarking for testing purposes.

### DIFF
--- a/deep_quoridor/src/agents/dexp.py
+++ b/deep_quoridor/src/agents/dexp.py
@@ -30,13 +30,15 @@ class DExpNetwork(nn.Module):
 
         # Define network architecture
         self.model = nn.Sequential(
-            nn.Linear(flat_input_size, 1024),
+            nn.Linear(flat_input_size, 2048),
             nn.ReLU(),
-            nn.Linear(1024, 1024),
+            nn.Linear(2048, 4096),
             nn.ReLU(),
-            nn.Linear(1024, 512),
+            nn.Linear(4096, 4096),
             nn.ReLU(),
-            nn.Linear(512, action_size),
+            nn.Linear(4096, 2048),
+            nn.ReLU(),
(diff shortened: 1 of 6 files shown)


deep_quoridor/src/arena.py
```diff
@@ -27,10 +27,12 @@ def __init__(
         saver: Optional[ArenaPlugin] = None,
         plugins: list[ArenaPlugin] = [],
         swap_players: bool = True,
+        max_steps: int = 10000,
```
adamantivm, line 30: this is great!

deep_quoridor/src/agents/dexp.py
```diff
@@ -30,13 +30,15 @@ def __init__(self, board_size, action_size, split_board, include_turn):
...
-            nn.Linear(1024, 1024),
+            nn.Linear(2048, 4096),
             nn.ReLU(),
-            nn.Linear(1024, 512),
+            nn.Linear(4096, 4096),
             nn.ReLU(),
-            nn.Linear(512, action_size),
+            nn.Linear(4096, 2048),
+            nn.ReLU(),
+            nn.Linear(2048, action_size),
```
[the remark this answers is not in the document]
muralx: Yes, I need to see how to properly size the model. I did notice that the small model was not working well after the latest changes, not sure why, so I went with a larger, but I think this is likely too big.

deep_quoridor/src/arena.py
muralx: I implemented it in the arena, but I am adding a TODO to push this to the environment and to implement proper truncation. (I will wait for new changes from @jonbinney are done for that)

muralx:
Thanks, submitting

PR #167, opened 2025-04-17
Ignore qvalues for actions that are not allowed when computing qvalue

muralx:
Results from the change are not great. It is an optional setting. Exp1 does not use the new setting, Exp2 uses it.

```
+--------------+-----------+-----------+-----------+-----------+-----------+--------+--------+-------+
|    Player    | dexp_best | dexp_exp1 | dexp_exp2 | greedy-01 | greedy-03 | sb3ppo | simple | Total |
+--------------+-----------+-----------+-----------+-----------+-----------+--------+--------+-------+
| P1 dexp_best |     -     |    100%   |     0%    |    84%    |    74%    |  100%  |  94%   |  75%  |
| P2 dexp_best |     -     |     0%    |    100%   |    88%    |    76%    |   0%   |  84%   |  58%  |
| P1 dexp_exp1 |    100%   |     -     |    100%   |    92%    |    62%    |   0%   |  46%   |  67%  |
| P2 dexp_exp1 |     0%    |     -     |     0%    |     6%    |    28%    |  100%  |  50%   |  31%  |
| P1 dexp_exp2 |     0%    |    100%   |     -     |    14%    |    50%    |  100%  |  90%   |  59%  |
| P2 dexp_exp2 |    100%   |     0%    |     -     |    82%    |    64%    |  100%  |  82%   |  71%  |
| P1 greedy-01 |    12%    |    94%    |    18%    |     -     |    90%    |  100%  |  100%  |  69%  |
| P2 greedy-01 |    16%    |     8%    |    86%    |     -     |    52%    |  98%   |  98%   |  60%  |
| P1 greedy-03 |    24%    |    72%    |    36%    |    48%    |     -     |  92%   |  96%   |  61%  |
| P2 greedy-03 |    26%    |    38%    |    50%    |    10%    |     -     |  74%   |  94%   |  49%  |
|  P1 sb3ppo   |    100%   |     0%    |     0%    |     2%    |    26%    |   -    |  30%   |  26%  |
|  P2 sb3ppo   |     0%    |    100%   |     0%    |     0%    |     8%    |   -    |  62%   |  28%  |
|  P1 simple   |    16%    |    50%    |    18%    |     2%    |     6%    |  38%   |   -    |  22%  |
|  P2 simple   |     6%    |    54%    |    10%    |     0%    |     4%    |  70%   |   -    |  24%  |
+--------------+-----------+-----------+-----------+-----------+-----------+--------+--------+-------+

```

### DIFF
--- a/deep_quoridor/src/agents/core/trainable_agent.py
+++ b/deep_quoridor/src/agents/core/trainable_agent.py
@@ -174,6 +174,7 @@ class AbstractTrainableAgent(Agent):
         state_before_action = self.observation_to_tensor(observation_before_action, player_id)
         state_after_action = self.observation_to_tensor(game.observe(player_id), player_id)
         opponent_state = game.observe(self.get_opponent_player_id(player_id))
+        # next action mask is stored with the same rotation of the next state.
         next_state_mask = self.convert_action_mask_to_tensor_for_player(
             opponent_state["action_mask"], self.get_opponent_player_id(player_id)
         )
@@ -261,7 +262,7 @@ class AbstractTrainableAgent(Agent):
         return np.random.choice(valid_actions)
 
     def convert_action_mask_to_tensor(self, mask):
-        return self.convert_action_mask_to_tensor_for_player(mask.self.player_id)
+        return self.convert_action_mask_to_tensor_for_player(mask, self.player_id)
 
     def convert_action_mask_to_tensor_for_player(self, mask, player_id):
         return torch.tensor(mask, dtype=torch.float32, device=self.device)
--- a/deep_quoridor/src/agents/core/trainable_agent.py
+++ b/deep_quoridor/src/agents/core/trainable_agent.py
@@ -137,6 +137,10 @@ class AbstractTrainableAgent(Agent):
         if (self.games_count % self.update_target_every) == 0:
             self.update_target_network()
 
+    def get_opponent_player_id(self, player_id):
+        """Get the opponent player ID."""
+        return "player_1" if player_id == "player_0" else "player_0"
+
     def handle_opponent_step_outcome(self, observation_before_action, action, game):
         pass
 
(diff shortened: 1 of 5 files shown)


deep_quoridor/src/agents/core/trainable_agent.py
```diff
@@ -182,6 +195,7 @@ def handle_step_outcome_all(self, observation_before_action, action, game, playe
             if state_after_action is not None
             else np.zeros_like(state_before_action.cpu().numpy()),
             float(done),
+            next_state_mask.cpu().numpy(),
```
adamantivm, line 198: shouldn't all the next_state_mask logic be done only if the mask_targetq flag is true?

deep_quoridor/src/agents/core/trainable_agent.py
```diff
@@ -305,32 +322,47 @@ def train(self, batch_size):
...
     def _compute_current_q_values(self, states, actions):
         """Compute current Q-values."""
         return self.online_network(states).gather(1, actions).squeeze()
 
-    def _compute_target_q_values(self, rewards, next_states, dones):
-        """Compute target Q-values."""
+    def _compute_target_q_values(self, rewards, next_states, dones, next_state_masks):
+        """Compute target Q-values considering action masks."""
+
+        if self.params.mask_targetq:
```
adamantivm, line 353: nit: this way both branches of the if look very different  (differently formatted, return from the middle of one if branch with no else, etc.

I think it could use some improved readability making both branches somewhat consistent.

It's just a nit, feel free to ignore

adamantivm (approved):
I suggest making all the processing of the target masks also dependent of the flag, so that they are not calculated in vein. 
The last reuqest: It would be nice to include in the PR description also the command you ran in order to get those results.
Feel free to ignore all my comments. Great job! Super interesting.

deep_quoridor/src/agents/core/trainable_agent.py
muralx: Done, I cleaned up. I tried to keep the old one unmodified, but it did not make sense.

muralx:
@adamantivm Thanks for the review. I addressed all comments. I don't have the command line, but it was mainly based on DEXP yaml file. Also tags on the models as exp1, exp2, last, best are not static so it will be hard to make it reproducible. For now on, I will use v1, vn for models and automatically adjust the nick for that version, so it will be easier to repro things.

I would really like to store runs results from evals in wandb, but I need to look if that is possible.

PR #168, opened 2025-04-19
Mutiple changes that were in my testing workspace

muralx:
Trainable agent
- Add the capability  to inspect actions for the opponent, computing qvalues  for the an opponent's state
- Allows using softmax action selection during explotation. Not ideal, requires more tweaking, off by default
- Increases replay buffer to 400k capacity (According to lecture large replay buffer is better)

Environment (@jonbinney in case something needs to change also in the new impl)
- Fix shortest path implementation that was not actually returning the shortest path  (@alejandromarcu fyi)
- Encapsulate step rewards computation, and uses a better computation that relies on whether situation improved with the action instead of giving negative rewards if situation improved but it is still bad. (Requires more tweaking)


deep_quoridor/src/quoridor_env.py
```diff
@@ -547,3 +592,38 @@ def env(**kwargs):
         return wrappers.CaptureStdoutWrapper(QuoridorEnv(**kwargs))
     else:
         return QuoridorEnv(**kwargs)
+
+
+class StepRewardCalculator:
```
jonbinney, line 597: I like having the step rewards as a separate class! Should make it easier to have various different reward functions in the future.

deep_quoridor/src/quoridor_env.py
```diff
@@ -547,3 +592,38 @@ def env(**kwargs):
...
+        agent_distance, opponent_distance = self._get_distances()
+
+        # Calculate the reward based on the distance to the goal
+        reward = 0
+        distance_change = (self.orig_agent_distance - agent_distance) - (
+            self.orig_opponent_distance - opponent_distance
+        )
+        if distance_change > 0:
+            reward = 1
+        elif distance_change < 0:
```
jonbinney, line 627: @muralx @adamantivm @alejandromarcu have any of you tried making the step reward larger depending on the distance change? For instance adding a wall might increase the opponent distance to goal by a lot.

jonbinney (approved):
Environment changes LGTM. I trust you on the dexp changes :-)

deep_quoridor/src/quoridor_env.py
```diff
@@ -470,7 +469,49 @@ def get_opponent(self, agent):
     def get_goal_row(self, agent):
         return 0 if agent == "player_1" else self.board_size - 1
 
-    def _dfs(self, row, col, target_row, visited, any_path=True):
+    def _bfs(self, row, col, target_row, visited):
+        """
+        Performs a breadth-first search to find whether the pawn can reach the target row.
```
alejandromarcu, line 474:
```suggestion
        Performs a breadth-first search to find the shortest path to the target row.
```

deep_quoridor/src/quoridor_env.py
```diff
@@ -521,7 +559,10 @@ def distance_to_target(self, row, col, target_row, any_path=False):
         Otherwise, the shortest path will be returned (potentially slower)
         """
         visited = np.zeros((self.board_size, self.board_size), dtype="bool")
-        return self._dfs(row, col, target_row, visited, any_path)
+        if any_path:
```
alejandromarcu, line 562: I think the API for this is getting confusing.  I'd keep `can_reach` that can directly call _dfs, and in this method remove `any_path` and call _dfs, so that we can get the minimum distance (I feel like when you say distance it's usually the minimum)

alejandromarcu (approved):
The _dfs function was returning the Julianan distance when using any_path 😝 .  I think it will be clearer if when we talk about distance it's the minimum, and when we just want to know if we can reach we use dfs that should be faster in general and we don't care if it's the minimum.

PR #169, opened 2025-04-22
Elo computation and renderer

alejandromarcu:
Defines a method to compute the elo rating from the results, and uses it in a renderer.
In a future PR I'll use this to get the score of the agent that we're training and log it in wandb.  For now, I think it's interesting to see the scores.  E.g.:

```
% python deep_quoridor/src/play.py -yi benchmark_B5W3
.....
+-------------+-------------+--------+-----------+-----------+--------+--------+--------+--------+
|   P1 \ P2   | dexp (1011) | greedy | greedy-01 | greedy-03 | random | sb3ppo | simple | Total  |
+-------------+-------------+--------+-----------+-----------+--------+--------+--------+--------+
| dexp (1011) |      -      |  100%  |    90%    |    74%    |  82%   |  100%  |  82%   |  88%   |
|    greedy   |      0%     |   -    |    100%   |    98%    |  100%  |  100%  |  100%  |  83%   |
|  greedy-01  |     14%     |  78%   |     -     |    94%    |  98%   |  100%  |  100%  |  81%   |
|  greedy-03  |     22%     |  30%   |    34%    |     -     |  100%  |  92%   |  100%  |  63%   |
|    random   |     18%     |   0%   |     0%    |     2%    |   -    |  44%   |  22%   |  14%   |
|    sb3ppo   |      0%     |   0%   |     2%    |    20%    |  52%   |   -    |  32%   |  18%   |
|    simple   |     22%     |   0%   |     4%    |     2%    |  58%   |  34%   |   -    |  20%   |
|    ======   |    ======   | ====== |   ======  |   ======  | ====== | ====== | ====== | ====== |
|    Total    |     13%     |  35%   |    38%    |    48%    |  82%   |  78%   |  73%   |  52%   |
+-------------+-------------+--------+-----------+-----------+--------+--------+--------+--------+
+-------------+------------+
|    Player   | Elo Rating |
+-------------+------------+
| dexp (1011) |    1962    |
|    greedy   |    1758    |
|  greedy-01  |    1756    |
|  greedy-03  |    1527    |
|    simple   |    1250    |
|    sb3ppo   |    1191    |
|    random   |    1054    |
+-------------+------------+
```

### DIFF
--- a/deep_quoridor/src/utils/misc.py
+++ b/deep_quoridor/src/utils/misc.py
@@ -116,3 +116,29 @@ def yargs(parser: argparse.ArgumentParser, default_path: str):
                     args_dict[k] = parser.parse_args(v["args"].split(" "))
 
     return args_dict
+
+
+def compute_elo(results: list["GameResult"], k=32, initial_rating=1500) -> dict[str, float]:
+    ratings = {}
+    random.shuffle(results)
+
+    for result in results:
+        if result.player1 not in ratings:
+            ratings[result.player1] = initial_rating
+
+        if result.player2 not in ratings:
+            ratings[result.player2] = initial_rating
+
+        delta = (ratings[result.player1] - ratings[result.player2]) / 400.0
+        e1 = 1 / (1 + 10**-delta)
+        e2 = 1 / (1 + 10**delta)
+
+        # If the game was truncated because it was too long, they both will count as losing
+        # so their rating decreases
+        p1_won = 1 if result.winner == result.player1 else 0
+        p2_won = 1 if result.winner == result.player2 else 0
+
+        ratings[result.player1] += k * (p1_won - e1)
+        ratings[result.player2] += k * (p2_won - e2)
+
+    return ratings
--- a/deep_quoridor/src/renderers/__init__.py
+++ b/deep_quoridor/src/renderers/__init__.py
@@ -25,6 +25,7 @@ __all__ = [
     "ArenaResults2Renderer",
     "ArenaResultsRenderer",
     "CursesBoardRenderer",
+    "EloResultsRenderer",
     "MatchResultsRenderer",
     "NoneRenderer",
     "ProgressBarRenderer",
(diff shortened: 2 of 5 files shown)


muralx (approved):
Nice!

muralx:
We should upload the results table and the elo table too

deep_quoridor/src/utils/misc.py
```diff
@@ -116,3 +116,29 @@ def yargs(parser: argparse.ArgumentParser, default_path: str):
                     args_dict[k] = parser.parse_args(v["args"].split(" "))
 
     return args_dict
+
+
+def compute_elo(results: list["GameResult"], k=32, initial_rating=1500) -> dict[str, float]:
```
jonbinney, line 121: Is `"GameResult"` a string in the type hint so that you don't need to import that class in this file?

deep_quoridor/src/utils/misc.py
```diff
@@ -116,3 +116,29 @@ def yargs(parser: argparse.ArgumentParser, default_path: str):
...
+
+        if result.player2 not in ratings:
+            ratings[result.player2] = initial_rating
+
+        delta = (ratings[result.player1] - ratings[result.player2]) / 400.0
+        e1 = 1 / (1 + 10**-delta)
+        e2 = 1 / (1 + 10**delta)
+
+        # If the game was truncated because it was too long, they both will count as losing
+        # so their rating decreases
```
jonbinney, line 137: Ouch! But i guess this makes sense..... if either player is smart, there shouldn't be any way for the game to go on forever. Do we do game truncation anywhere in the arena code yet?

jonbinney (approved):
LGTM!

deep_quoridor/src/utils/misc.py
alejandromarcu: yes, because if I import the class I get a circular reference.  Using a string works anyway for the linter, so it's easier than having to re-organize things in the files to be able to import the lcass

deep_quoridor/src/utils/misc.py
alejandromarcu: Diego implemented game truncation in the Arena class, is currently set to 10,000, so if any of the agents is not completely horrible it should converge much faster anyway.

PR #170, opened 2025-04-24
Integrate elo ratings in wandb

alejandromarcu:
When training an agent, send the elo rating and the win percentage to wandb, so we can have a single number (or two) of how it is performing.
The whole elo table is sent to wandb every interval, but the tables are clunky and I still haven't figured out how they work.
It also sends what I call the "relative elo" score, which is the score the agent got minus the score of the best opponent.  The rationale behind this is the following: we want to have information on how good the model is performing with just 1 number, but if we just see our agent's elo rating, it's hard to draw conclusions.  Is an elo rating of 1800 good? well, it depends, if a few other agents are above 1800 it's not that great, but if everyone is below 1800 it looks much better.
So, if the relative elo is positive, it means that the agent is in the first position.  A relative elo score of 200 would indicate for example that our agent has about 75% chances of beating the second one, and a score of 400 would indicate 91% chances.
When the score is negative, we know the agent is not the best.

I'm also logging the win percentage, that has the same shape than the relative elo, so we may end up just using the win percentage.  The main difference is that the elo takes into account how good the adversaries are, e.g. you get your elo increased more if you beat an adversary with higher elo, so winning against greedy is more significant than winning against random.

The win_perc has the advantage that if it reaches 100%, you know that you topped the metric, so you won't be trying to improve it without devising a harder metric first.  With elo score, you may think you still have room for improvement because there's no ceiling.

I'm currently running this training (copied from the command line from `dexp_B5W3_mv2:v15`, but added epsilon_min and reduced the number of episodes):

```
 /Users/amarcu/code/deep_rabbit_hole/.venv/bin/python /Users/amarcu/code/deep_rabbit_hole/deep_quoridor/src/train.py  -N 5 -W 3 -e 500 -i 43 -p dexp:rotate=true,split=true,turn=false,epsilon=1,epsilon_decay=0.9985,epsilon_min=0.1,final_reward_multiplier=20,training_mode=true,use_opponents_actions=true,update_target_every=100,gamma=0.95,target_as_source_for_opponent=true,use_negative_qvalue_function=true,batch_size=128,mask_targetq=true,nick=dexp_train greedy:p_random=0.3 greedy:p_random=0.1 simple -rp cucu -s
```

And this is what I got so far ([link to the run](https://wandb.ai/the-lazy-learning-lair/deep_quoridor/runs/cucu-20250423-164800)):
image

### DIFF
--- a/deep_quoridor/src/metrics.py
+++ b/deep_quoridor/src/metrics.py
@@ -1,0 +1,95 @@
+from agents import AbstractTrainableAgent, Agent
+from agents.core.agent import AgentRegistry
+from arena import Arena, PlayMode
+from arena_utils import GameResult
+from utils.misc import compute_elo
+
+
+class Metrics:
+    """
+    Computes metrics for an agent by making it play against other agents.
+    """
+
+    def __init__(self, board_size: int, max_walls: int):
+        self.board_size = board_size
+        self.max_walls = max_walls
+        self.stored_elos = {}
+
+    def _win_perc(self, results: list[GameResult], agent_name: str):
+        played = 0
+        won = 0
+        for result in results:
+            if result.player1 == agent_name or result.player2 == agent_name:
+                played += 1
+
+            if result.winner == agent_name:
+                won += 1
+
+        return 100.0 * won / played if played > 1 else 0.0
+
+    def _compute_relative_elo(self, elo_table: dict[str, float], agent_name: str) -> int:
+        agent_rating = elo_table[agent_name]
+        best_opponent = 0
+        for name, elo in elo_table.items():
+            if name != agent_name:
+                best_opponent = max(best_opponent, elo)
+
+        return int(agent_rating - best_opponent)
+
+    def compute(self, agent: str | Agent) -> tuple[int, dict[str, float], int, float]:
+        """
+        Evaluates the performance of a given agent by running it against a set of predefined opponents and computing its Elo rating and win percentage.
+
+        Args:
+            agent (str | Agent): The agent to evaluate, specified either as a string identifier or an Agent instance.
+
+        Returns:
+            tuple[int, dict[str, float], int, float]: A tuple containing:
+                - VERSION (int): The version number of the scoring method.
+                - elo_table (dict[str, float]): A dictionary mapping agent names to their computed Elo ratings.
+                - relative_elo (int): The evaluated agent's Elo rating minus the elo for the best opponent.
+                - win_perc (float): The win percentage of the evaluated agent against the opponents.
+
+        Notes:
+            - The method disables training mode for trainable agents during evaluation and restores it afterward.
+            - Opponent Elo ratings are cached to avoid redundant computations.
+            - The evaluation is performed using a fixed set of baseline agents and a specified number of games.
+        """
+        # Bump if there's any change in the scoring
+        VERSION = 1
+        times = 50
+
+        players: list[str | Agent] = [
+            "greedy",
+            "greedy:p_random=0.1,nick=greedy-01",
+            "greedy:p_random=0.3,nick=greedy-03",
+            "greedy:p_random=0.5,nick=greedy-05",
+        ]
+
+        if isinstance(agent, str):
+            agent = AgentRegistry.create_from_encoded_name(agent, board_size=self.board_size, max_walls=self.max_walls)
+
+        # Make sure the agent is not in training mode to do the evaluation
+        if isinstance(agent, AbstractTrainableAgent):
+            prev_training_mode = agent.training_mode
+            agent.training_mode = False
+
+        arena = Arena(self.board_size, self.max_walls)
+
+        # We store the elos of the opponents playing against each other so we don't have to play those matches
+        # every time
+        if not self.stored_elos:
+            results = arena._play_games(players, times, PlayMode.ALL_VS_ALL)
+            self.stored_elos = compute_elo(results)
+
+        results = arena._play_games([agent] + players, times, PlayMode.FIRST_VS_RANDOM)
+
+        if isinstance(agent, AbstractTrainableAgent):
+            agent.training_mode = prev_training_mode
+
+        elo_table = compute_elo(results, initial_elos=self.stored_elos.copy())
+        relative_elo = self._compute_relative_elo(elo_table, agent.name())
+
+        win_perc = self._win_perc(results, agent.name())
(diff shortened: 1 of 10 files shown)


deep_quoridor/src/train.py
```diff
@@ -24,7 +24,7 @@ def train_dqn(
     plugins = []
 
     if use_wandb:
-        plugins.append(WandbTrainPlugin(update_every=10, total_episodes=episodes, run_id=run_id))
+        plugins.append(WandbTrainPlugin(update_every=100, total_episodes=episodes, run_id=run_id))
```
muralx, line 27: I think we should upload the elo and results after running a tournament with the training complete, similar to the eval that Julian's training process is doing. Uploading the results from training (in relation to elo and % of winning is not really meaningful)

adamantivm: If I understand correctly, this PR is doing the same thing: running "play" against all players, but doing it many times instead of only one time at the end.

I think that's going to be pretty onerous. I think at the minimum we need to have the option of: 
 - Not running this
 - Running it just once at the end (default?)
 - Running it once every `update_every` (current impl) as an option

alejandromarcu: Yeah, this runs the "tournament" every N episodes and reports the results.  I think it's pretty useful because you can see how the training is progressing.  But I'll make a separate configuration for the frequency, so you can choose to run it it less often or not run it.   Doing it every 100 episodes in my case didn't significantly improve the time, since it was taking a bit over 1 minute to do 100 episodes and then 4-5 seconds to do the tournament.

alejandromarcu:
## UPDATE 
cc: @adamantivm , @muralx 

With the last commit, we can pass a tournament frequency, by default I put 500 episodes, I think in general the overhead of this should be very small, but it can be set to a lower frequency from the command line.

e.g. run with command:
```
/Users/amarcu/code/deep_rabbit_hole/.venv/bin/python /Users/amarcu/code/deep_rabbit_hole/deep_quoridor/src/train.py  -N 5 -W 3 -e 600 -i 43 -p dexp:rotate=true,split=true,turn=false,epsilon=1,epsilon_decay=0.9985,epsilon_min=0.1,final_reward_multiplier=20,training_mode=true,use_opponents_actions=true,update_target_every=100,gamma=0.95,target_as_source_for_opponent=true,use_negative_qvalue_function=true,batch_size=128,mask_targetq=true,nick=dexp_train greedy:p_random=0.3 greedy:p_random=0.1 simple -rp cucu -s
```
Here's the [run]( https://wandb.ai/the-lazy-learning-lair/deep_quoridor/runs/cucu-20250424-134353?nw=nwuseramarcu
) in wandb

image

muralx:
I think my comment was not submitted and it was pending. I've just sent it, but I will review the new version, but I still think metrics generation for an agent that happens to be computed while playing, just be as decoupled as possible from the wand training plugin.

alejandromarcu:
> I think my comment was not submitted and it was pending. I've just sent it, but I will review the new version. I still think metrics generation for an agent that happens to be computed while playing, should be as decoupled as possible from the wand training plugin.

Is this because of risks of affecting the training? E.g. I already had an issue that the epsilon was discounting when playing to measure the ELO and I fixed it, but there may be other sides effects.  We could save the model and instantiate a completely new agent to remove the risk.

I think running the metrics at a certain frequency during training provides valuable information.  E.g. if you look at the graph I posted here, you can see that the win perc and elo was increasing during training, so this is a good signal, and if I was trying to make it better I'd increase the episodes in this case.  But in another case you may see that the ELO reaches a max and it's not improving, so this helps you know that more episodes won't help, probably something else needs to be changed to make it improve.

deep_quoridor/src/metrics.py
```diff
@@ -0,0 +1,95 @@
...
+            "greedy:p_random=0.3,nick=greedy-03",
+            "greedy:p_random=0.5,nick=greedy-05",
+        ]
+
+        if isinstance(agent, str):
+            agent = AgentRegistry.create_from_encoded_name(agent, board_size=self.board_size, max_walls=self.max_walls)
+
+        # Make sure the agent is not in training mode to do the evaluation
+        if isinstance(agent, AbstractTrainableAgent):
+            prev_training_mode = agent.training_mode
```
muralx, line 74: Mhh, this might not be enough. Some methods might not be checking this. In particular I think end_game is updating target network to match online network (it should not), but it will impact training (slightly).  (Just as an example). We should review methods are properly using this.

muralx: Also, epsilon should ideally be zero during play when not training (no exploration)

alejandromarcu: I already changed end_game in this PR to return early if it's not in training mode.
As I mentioned, one thing we could do to make sure we don't interfere would be to save the model to disk and evaluate a new instance of the agent loading the model from the disk, so we don't risk any interference. I think I could just remove the`training_only_params` from the parameters that were passed for training.
Another advantage is that if it's saved to disk, the evaluation could be run in parallel in another thread (but of course we incur in the cost of saving the model, which is probably not a big deal for now).
What do you think?

BTW, another thing I'm wondering if it may be worth doing in the future is to have separate classes for training and playing, probably extending the same class, so things are separate, and the playing agent is much easier than the training one anyway,

muralx: Since we are already saving snapshots of the model, you could use that one. That being said if you have already verified methods are handling the training_mode setting properly, just go ahead and submit. We can always improve things later.

alejandromarcu:
# Update 2

With this new changes, the model is evaluated in a tournament after saving it.  The model is re-loaded from disk and a new instance of the agent is created, this way we eliminate the risk of interference between evaluation and training.
E.g. the previous version was evaluating the agent with the current value of the epsilon, which should be 0 during play.  Now it shows how it starts with better results, because we're actually using epsilon=0 during play.

image

I tried to run the evaluation in a separate thread, but the wall time ended being a bit longer (not sure why), and it makes the code more messy and the output unclear, so I scratched the idea.  Each save adds about 5s in my computer with the current setup, and by default we save every 500 episodes.  So, it shouldn't have a lot of impact on run time, but we can always pass a parameter to run it less often.  E.g. if we were training for many thousands episodes, it would make sense anyway to save and evaluate less often

muralx (approved):
Thanks Cucu!

PR #176, opened 2025-04-29
More PPO updates

adamantivm:
- Publishes loss, value_loss and other useful training params to wandb:
 - Adds custom (bigger) network params (alas with no great reasult so far)
 - Adds (and uses) the option to drop the "it's my turn" flag from the observation space, via a new wrapper param
 - Points the N3W0 benchmark to a new best sb3ppo model published using the "best" tag
 - Confirmed for N3W0 that rewards multiplier wasn't necessary, making the loss functions converge more clearly


deep_quoridor/src/agents/sb3_ppo.py
```diff
@@ -109,7 +110,7 @@ def model_name(self):
     @staticmethod
     def version():
         """Bump this version when compatibility with saved models is broken"""
-        return 2
+        return 3
```
muralx, line 113: I wish I remember doing this before running training with modified version of the agent

muralx (approved):

adamantivm:
I forgot to merge this earlier. I think it shouldn't conflict with the ongoing work. Let me know otherwise.

PR #178, opened 2025-04-30
Fix human player

alejandromarcu:
We were passing tuples of col, row, type but we needed Action objects, so using the human agent was throwing an error.
Used:
```
 /Users/amarcu/code/deep_rabbit_hole/.venv/bin/python /Users/amarcu/code/deep_rabbit_hole/deep_quoridor/src/play.py -yi human_vs_greedy
```


jonbinney (approved):
Thanks for fixing this! Looks good to me. I suspected there was some kind of bug like this, but i didn't see any error messages or exceptions in the console.

PR #179, opened 2025-04-30
Fix greedy and action logs

alejandromarcu:
Fixes the greedy agent issue that @muralx reported (see https://github.com/jonbinney/deep_rabbit_hole/pull/177/files).  The implementation is a bit different but it does the same: add the curent position to the shortest path (I forgot now what was the problem, but it was leading to a bug when the opponent next move is the target).
I also fixed action log that was broken with the new Quoridor class, and I added some logs for greedy: show the distance or RAND if the next movement is random

### DIFF
--- a/deep_quoridor/src/agents/greedy.py
+++ b/deep_quoridor/src/agents/greedy.py
@@ -4,7 +4,15 @@ from queue import Queue
 from typing import Optional, TypeAlias
 
 import numpy as np
-from quoridor import ActionEncoder, Quoridor, MoveAction, Player, WallAction, WallOrientation, construct_game_from_observation
+from quoridor import (
+    ActionEncoder,
+    MoveAction,
+    Player,
+    Quoridor,
+    WallAction,
+    WallOrientation,
+    construct_game_from_observation,
+)
 from utils import SubargsBase
 
 from agents.core import Agent
@@ -66,7 +74,7 @@ class GreedyAgent(Agent):
         Given a position in the board, it returns the shortest path to get to the target_row.
         """
         if pos[0] == target_row:
-            return []
+            return [pos]
 
         # How many steps are needed to get to that position, starting from pos (None means unreachable)
         distances = [[None] * game.board.board_size for _ in range(game.board.board_size)]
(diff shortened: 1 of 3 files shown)


deep_quoridor/src/agents/greedy.py
```diff
@@ -66,7 +74,7 @@ def _shortest_path_from(self, game: Quoridor, pos: Position, target_row: int) ->
         Given a position in the board, it returns the shortest path to get to the target_row.
         """
         if pos[0] == target_row:
-            return []
+            return [pos]
```
alejandromarcu, line 77: If we want the shortest path and start from the target row, we need to include this cell

deep_quoridor/src/agents/greedy.py
```diff
@@ -97,18 +105,20 @@ def _shortest_path_from(self, game: Quoridor, pos: Position, target_row: int) ->
...
         # Trace back the path
         path = [dest]
-        while True:
+        prev_pos = None
+        while prev_pos != pos:
             last_pos = path[-1]
             prev_pos = coming_from[last_pos]
-            if prev_pos == pos:
-                break
-
```
alejandromarcu, line 105: The change here does what @muralx  did of adding the initial position to the path

deep_quoridor/src/agents/greedy.py
```diff
@@ -117,7 +127,7 @@ def _shortest_path_from_me(self, game: Quoridor, action_mask: np.ndarray, target
         for move in first_moves:
             path = self._shortest_path_from(game, move, target_row)
             if shortest_path is None or len(path) < len(shortest_path):
-                shortest_path = [move] + path
+                shortest_path = [my_position] + path
```
alejandromarcu, line 130: now tha path includes the initial position, which is "move", so instead we need to include the current pawn position

deep_quoridor/src/agents/greedy.py
```diff
@@ -172,18 +182,16 @@ def _log_action(
...
-            # self.action_log.action_text(self.action_encoder.index_to_action(int(action)), "")
-            pass
+            self.action_log.action_text(self.action_encoder.index_to_action(int(action)), "")
 
-        my_coords = np.argwhere(observation["board"] == 1)
-        path = my_shortest_path[:]
-        path.insert(0, (int(my_coords[0][0]), int(my_coords[0][1])))
-        # self.action_log.path(path)
+        self.action_log.path(my_shortest_path)
+        self.action_log.path(opponent_shortest_path)
```
alejandromarcu, line 189: Now that the path includes the coordinates of the pawn we don't need to manually add them!

muralx (approved):

muralx:
For the record, the issue was that in case there was only one step to reach the goal, the shortest path would only include the target cell, and not the previous one, so the code to find the place to put a wall would not work since it required at least 2 entries in the list.

PR #180, opened 2025-05-01
Db fixdexp after big merge

muralx:
Also, fix some nits.

Trainable params include modelfile name, which breaks ELO computation, so I added a hacky fix to fix it.

### DIFF
--- a/deep_quoridor/src/agents/core/agent.py
+++ b/deep_quoridor/src/agents/core/agent.py
@@ -121,7 +121,9 @@ class AgentRegistry:
         return AgentRegistry.agents[friendly_name](**kwargs)
 
     @staticmethod
-    def create_from_encoded_name(encoded_name: str, remove_training_args=False, **kwargs) -> Agent:
+    def create_from_encoded_name(
+        encoded_name: str, remove_training_args=False, keep_args: set[str] = {}, **kwargs
+    ) -> Agent:
         parts = encoded_name.split(":")
         agent_type = parts[0]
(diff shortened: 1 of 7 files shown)


deep_quoridor/src/agents/core/agent.py
```diff
@@ -121,7 +121,9 @@ def create(friendly_name: str, **kwargs) -> Agent:
         return AgentRegistry.agents[friendly_name](**kwargs)
 
     @staticmethod
-    def create_from_encoded_name(encoded_name: str, remove_training_args=False, **kwargs) -> Agent:
+    def create_from_encoded_name(
```
muralx, line 124: Note that the compute code that calls this might actually break if we are training a model that happens to be initialized with a model file name. (This is a note for @alejandromarcu, and it is not related to this change)

deep_quoridor/src/agents/core/agent.py
```diff
@@ -109,7 +109,7 @@ def end_game(self, game):
         """
         pass
 
-    def get_action(self, game) -> int:
+    def get_action(self, observation, action_mask) -> int:
```
alejandromarcu, line 112: also replay.py needs to be updated to use this signature, can you add it to this PR?

deep_quoridor/src/agents/core/agent.py
alejandromarcu: Thanks, will take a look

PR #182, opened 2025-05-05
Use minimax algorithm in simple agent

jonbinney:
By default it tries a max of 20 different actions at each step and looks 2 steps ahead (e.g. if I do X, then opponent could do Y). Beyond that horizon it maximizes the usual heuristic of `(opponent_distance_to_goal - my_distance_to_goal)`. I've tried having it look 3 steps ahead, but to keep computation time down, I had to reduce the number of actions considered per turn to 8, which resulted in worse results.

This agent takes longer than the others; a couple tenths of a second per turn instead of a couple milliseconds per turn. It can take much much longer if the branching factor or max depth are increased. Should give us some interesting options for training other agents - we could start with a dumb simple agent and then later in the training process crank up its intelligence as the trained agent gets better.

There are a few parameters for the agent:
```sh
    # How many moves to look ahead in the minimax algorithm.
    max_depth: int = 2

    # How many actions to consider at each minimax stage.
    branching_factor: int = 20

    # We use a Guassian distribution centered around each player's position to sample wall actions.
    # Sigma is the standard deviation of this distribution; a smaller sigma means the agent is more
    # likely to place walls near themselves or their opponent, rather than somewhere else on the board.
    wall_sigma: float = 0.5

    # Future rewards are worth less. This makes the agent try to win quickly. Without this,
    # once the agent knows it can win, it may oscillate between two moves instead of just winning.
    discount_factor: float = 0.99
```

I wasn't able to test against the trained agents because there's not a model with the "best" tag for the current version, but it does pretty well against greedy:
```
Board size:  5
Max walls:  3
Step rewards:  False
Total number of games:  100 

+-----------+--------+-----------+-----------+--------+---------+--------+
|  P1 \ P2  | greedy | greedy-01 | greedy-03 | random | simple  | Total  |
+-----------+--------+-----------+-----------+--------+---------+--------+
|   greedy  |   -    |    100%   |    100%   |  100%  |    0%   |  75%   |
| greedy-01 |  80%   |     -     |    100%   |  100%  |   20%   |  75%   |
| greedy-03 |  60%   |    40%    |     -     |  100%  |    0%   |  50%   |
|   random  |   0%   |     0%    |     0%    |   -    |    0%   |   0%   |
|  simple   |  100%  |    80%    |    100%   |  100%  |    -    |  95%   |
|   ======  | ====== |   ======  |   ======  | ====== |  ====== | ====== |
|   Total   |  60%   |    55%    |    75%    |  100%  |    5%   |  59%   |
+-----------+--------+-----------+-----------+--------+---------+--------+
+-----------+------------+
|   Player  | Elo Rating |
+-----------+------------+
|  simple   |    1766    |
| greedy-01 |    1603    |
|   greedy  |    1548    |
| greedy-03 |    1384    |
|   random  |    1196    |
```

### DIFF
--- a/deep_quoridor/src/agents/simple.py
+++ b/deep_quoridor/src/agents/simple.py
@@ -83,7 +83,7 @@ def choose_action(
     branching_factor: int,
     wall_sigma: float | None,
     discount_factor: float,
-) -> float:
+) -> tuple[Optional[Action], float]:
     """
     Minimax algorithm to choose the best action for the current player.
 
--- a/deep_quoridor/src/agents/simple.py
+++ b/deep_quoridor/src/agents/simple.py
@@ -1,14 +1,24 @@
 import copy
 import random
 
-from quoridor import ActionEncoder, Player, Quoridor, construct_game_from_observation
+import numpy as np
+from quoridor import Action, ActionEncoder, Player, Quoridor, WallAction, construct_game_from_observation
 
 from agents.core import Agent
 
 
-def sample_valid_actions(game: Quoridor, n: int):
+def compute_wall_weight(wall: WallAction, position_1: np.ndarray, position_2: np.ndarray, sigma) -> float:
+    wall_position = np.array(wall.position)
+    d1 = np.linalg.norm(position_1 - wall_position)
+    d2 = np.linalg.norm(position_2 - wall_position)
+    min_distance = min(d1, d2)
+    weight = np.exp(-0.5 * (min_distance / sigma) ** 2)
+    return weight
+
+
+def sample_actions(game: Quoridor, n: int, wall_sigma: float | None = None) -> list[Action]:
     """
-    Choose a random sample of valid actions for the current player.
+    Choose a sample of valid actions for the current player.
 
     Starts by including all valid move actions, then adds a random sample of valid wall actions. May return less than
     n actions if there are not enough valid actions.
@@ -21,14 +31,32 @@ def sample_valid_actions(game: Quoridor, n: int):
         num_wall_actions = n - len(actions)
         wall_actions = game.get_valid_wall_actions()
         if len(wall_actions) > num_wall_actions:
-            wall_actions = random.sample(wall_actions, num_wall_actions)
+            if wall_sigma is None:
+                wall_actions = random.sample(wall_actions, num_wall_actions)
+            else:
+                position_1 = np.array(game.board.get_player_position(Player.ONE))
+                position_2 = np.array(game.board.get_player_position(Player.TWO))
+                wall_weights = [compute_wall_weight(wall, position_1, position_2, wall_sigma) for wall in wall_actions]
+                wall_actions = np.random.choice(
+                    wall_actions,
+                    size=num_wall_actions,
+                    replace=False,
+                    p=wall_weights / np.sum(wall_weights),
+                )
 
-        actions += wall_actions
+        actions.extend(wall_actions)
 
     return actions
 
 
-def choose_action(game: Quoridor, player: Player, opponent: Player, max_depth: int, branching_factor: int) -> float:
+def choose_action(
+    game: Quoridor,
+    player: Player,
+    opponent: Player,
+    max_depth: int,
+    branching_factor: int,
+    wall_sigma: float | None = None,
+) -> float:
     """
     Minimax algorithm to choose the best action for the current player.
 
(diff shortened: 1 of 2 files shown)


deep_quoridor/src/agents/simple.py
```diff
@@ -1,61 +1,175 @@
...
 
+def choose_action(
+    game: Quoridor,
+    player: Player,
+    opponent: Player,
+    max_depth: int,
+    branching_factor: int,
+    wall_sigma: float | None,
+    discount_factor: float,
+) -> float:
```
alejandromarcu, line 86: I think this should be:

```suggestion
) -> tuple[Optional[Action], float]:
```

deep_quoridor/src/agents/simple.py
```diff
@@ -1,61 +1,175 @@
...
+    # How many moves to look ahead in the minimax algorithm.
+    max_depth: int = 2
+
+    # How many actions to consider at each minimax stage.
+    branching_factor: int = 20
+
+    # We use a Guassian distribution centered around each player's position to sample wall actions.
+    # Sigma is the standard deviation of this distribution; a smaller sigma means the agent is more
+    # likely to place walls near themselves or their opponent, rather than somewhere else on the board.
+    wall_sigma: float = 0.5
```
alejandromarcu, line 29: I think that usually you want to put walls around the opponent, except at the beginning when you're marching towards each other and you don't want to block your own path.  Maybe using 2 different wall sigmas so you look less on your side could trim unnecessary actions?

deep_quoridor/src/agents/simple.py
```diff
@@ -1,61 +1,175 @@
...
+
+    # Future rewards are worth less. This makes the agent try to win quickly. Without this,
+    # once the agent knows it can win, it may oscillate between two moves instead of just winning.
+    discount_factor: float = 0.99
+
+
+def compute_wall_weight(wall: WallAction, position_1: np.ndarray, position_2: np.ndarray, sigma) -> float:
```
alejandromarcu, line 36: If in the profiler this is taking a significant amount of time, it may be worth vectorizing it, i.e. passing all the wall actions at once, since all those math operations would be much more efficient

deep_quoridor/src/agents/simple.py
```diff
@@ -1,61 +1,175 @@
...
+    Minimax algorithm to choose the best action for the current player.
+
+    The other agent tries to minimize the reward of the current player.
+    """
+    if game.check_win(player):
+        return None, WINNING_REWARD
+    elif game.check_win(opponent):
+        return None, -WINNING_REWARD
+    elif max_depth == 0:
+        return None, game.player_distance_to_target(opponent) - game.player_distance_to_target(player)
```
alejandromarcu, line 97: What do you think of including the number of walls remaining as a small fraction, so that it can help break ties?

deep_quoridor/src/agents/simple.py
alejandromarcu: Also, I just tried doing:
```
    print(min_distance, weight)
```

and got for example:
```
1.0 0.1353352832366127
1.4142135623730951 0.018315638888734165
2.23606797749979 4.539992976248477e-05
```
So, this decreases very quickly, the 3rd row is a difference in 1 vertical and 2 horizontal and it's already 3 orders of magnitude less likely to be chosen than one that is just 1 in difference, so almost never would be chosen.

Additionally, I realized the following: the coordinate of the wall is the top left, so for example a wall sitting right below the pawn would have different distances if it's extending to the right or to the left.  I think we should consider the distance to the center of the wall

deep_quoridor/src/agents/simple.py
```diff
@@ -1,61 +1,175 @@
...
+    actions = game.get_valid_move_actions()
+    if len(actions) > n:
+        actions = random.sample(actions, n)
+
+    if len(actions) < n:
+        num_wall_actions = n - len(actions)
+        wall_actions = game.get_valid_wall_actions()
+        if len(wall_actions) > num_wall_actions:
+            if wall_sigma is None:
+                wall_actions = random.sample(wall_actions, num_wall_actions)
```
alejandromarcu, line 61: I think just picking up a random wall would never be a decent strategy, any reason to keep this?

deep_quoridor/src/agents/simple.py
```diff
@@ -1,61 +1,175 @@
...
+            else:
+                position_1 = np.array(game.board.get_player_position(Player.ONE))
+                position_2 = np.array(game.board.get_player_position(Player.TWO))
```
alejandromarcu, line 64: nit: I'd just pass the positions without the array and make compute_wall_weight convert them, since it's more of an implementation thing

alejandromarcu (commented):
Very nice!

deep_quoridor/src/agents/simple.py
jonbinney: I think you may be right; I'd like to explore that idea in a future PR. Having different standard deviations around each player could work, but it might be be better to have a scaling factor on the value instead (since decreasing or increasing the standard devation of a gaussian doesn't actually change the total probability of it)

deep_quoridor/src/agents/simple.py
jonbinney: Yes on both points. The value decreasing quickly was (mostly) intentional since I wanted to really see the effect. Since it chooses wall actions without replacement, this results in almost always choosing the N closest walls. 

The distance being wrong is a bigger bug; I'll take a look at fixing that today.

deep_quoridor/src/agents/simple.py
jonbinney: It's a good idea; I'd like to leave that to a future PR. I want this one to get the overall minimax algorithm in, then I'll try a bunch of things out.

deep_quoridor/src/agents/simple.py
jonbinney: It's still sliiiightly useful so I'd like to keep it. I like having it as a comparison against the "choose walls near players" strategy. It definitely only works at all when the branching factor (and therefor num_wall_actions) is large. Increasing wall sigma to a very large value has a similar effect, but doesn't quite create a true uniform distribution.

deep_quoridor/src/agents/simple.py
jonbinney: Good call; fixed.

deep_quoridor/src/agents/simple.py
jonbinney: Updated to use the distance from the center of the wall to the player position, and performance is better now!

I did try increasing wall_sigma from 0.5 to 1, but performance slightly decreased, so I'll leave it at 0.5 for now.

jonbinney:
@alejandromarcu I've addressed your comments; ready for a another look.

alejandromarcu (approved):

alejandromarcu:
BTW, another thing that could be interesting for future PRs would be to use ActionLog to show the values that are assigned to the different actions in the UI.

PR #184, opened 2025-05-05
New version of DExp using adapter

muralx:
Extra base class for trainnable agent (interface like), with not logic
Create a base adapter for trainable agents
Create all needed adapters to transform observations based on what Dexp was using internally
Creates one agent that is preconfigured with all the adapters

Also changes get action (all agents) to just receive observation since already has action mask, otherwise we have two kind of observations around that is confusing and error prone.


deep_quoridor/src/agents/pagent.py
```diff
@@ -0,0 +1,180 @@
...
+        max_walls,
+        observation_space,
+        action_space,
+        params: PAgentParams = PAgentParams(),
+        **kwargs,
+    ):
+        new_action_space = action_space
+        new_observation_space = DictSplitBoardAdapter.get_observation_space(
+            RemoveTurnAdapter.get_observation_space(observation_space)
+        )
```
alejandromarcu, line 166: This is a bit cumbersome and fragile (e.g. if you change the wrappers below you need to remember to change this and you need to know which ones to use).  

Would it be possible to just chain those methods automatically, like we're doing with the rest of the adapter methods?  so each adapter would adapt (by default do nothing) the observation parameter of its agent, and the last agent needs to return somehow the space of the env

deep_quoridor/src/agents/pagent.py
```diff
@@ -0,0 +1,180 @@
...
+                flat_obs.extend(value.flatten())
+            else:
+                flat_obs.append(value)
+        return torch.FloatTensor(flat_obs).to(self.device)
+
+    def _convert_action_mask_to_tensor_for_player(self, mask, player_id):
+        return torch.tensor(mask, dtype=torch.float32, device=self.device)
+
+
+class NewDexpAgent(BaseTrainableAgentAdapter):
```
alejandromarcu, line 149: I find it a bit confusing that some stuff is in `PAgentAgent`, e.g. the name and model name (so this will appear as "pagent", not as "newdexp").  Also, for example let's say that you change the wrappers, then I think you need to update the version, but the version is in PAgentAgent, and if someone else was using it, their version would be incremented unnecessarily.  Not a big deal, not sure what a good solution would be anyway.

alejandromarcu: One possibility to avoid having 2 classes would be that AgentRegistry instead of calling the constructor, calls a static method `create` or something, that would give the opportunity to the agents to be wrapped (and by default it can just call the constructor)

deep_quoridor/src/agents/adapters/dict_split_board_adapter.py
```diff
@@ -0,0 +1,97 @@
...
+                space[key] = value
+            else:
+                obs_spc = {}
+                for key, value in original_space["observation"].items():
+                    if key == "board":
+                        obs_spc["my_board"] = original_space["observation"]["board"].__class__(
+                            0, 1, shape=board_shape, dtype=np.float32
+                        )
+                        obs_spc["opponent_board"] = original_space["observation"]["board"].__class__(
+                            0, 1, shape=board_shape, dtype=np.float32
```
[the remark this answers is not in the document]
muralx: No, i think it was like that in some other wrapper, in here we would use the same space (whatever it is), but we need to change the expected values to be 0,1 instead of 0,1 and 2

deep_quoridor/src/agents/adapters/base.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def save_model(self, path: str) -> None:
+        self._agent().save_model(path)
+
+    def load_model(self, path: str) -> None:
+        self._agent().load_model(path)
+
+    def get_opponent_player_id(self, player_id):
+        """Get the opponent player ID."""
+        return "player_1" if self.player_id == "player_0" else "player_0"
```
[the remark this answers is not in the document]
muralx: This was moved to a library and forgot to delete it.

deep_quoridor/src/agents/pagent.py
muralx: The problem is the agent needs the observation space created by the adapters since it will need to process it, but adapters need the agent in order to be created, so not ideal.

We could delay init in the agent, and have an init_with_space(..) or something, but also I don't like it. I will think of something, but I agree how it is right now it is not good. 

I don't like the adapters to be honest, but otherwise I will need to start copying and pasting for doing CNN, since I will need a completely different observation.

alejandromarcu: ok, we can move forward and if it causes problems in the future we'll figure something out.  Move fast and break things :p

alejandromarcu (approved):
ship it!

deep_quoridor/src/agents/pagent.py
```diff
@@ -0,0 +1,180 @@
...
+from agents.adapters.base import BaseTrainableAgentAdapter
+from agents.adapters.dict_split_board_adapter import DictSplitBoardAdapter
+from agents.adapters.remove_turn_adapter import RemoveTurnAdapter
+from agents.adapters.rotate_adapter import RotateAdapter
+from agents.adapters.use_opponent_after_action_obs import UseOpponentsAfterActionObsAdapter
+from agents.core import AbstractTrainableAgent
+from agents.core.trainable_agent import TrainableAgentParams
+
+
+class PAgentNetwork(nn.Module):
```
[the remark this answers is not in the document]
muralx: Renamed to use a better name.

deep_quoridor/src/agents/pagent.py
muralx: I have created a factory helper method that makes sure things are done in order, also re-arrange things based on renames, etc.

Note that I think the class NDexp is not needed, but not using it will bring additional issues that need to be solved and I prefer to move with other things.

deep_quoridor/src/agents/pagent.py
muralx: Yes, I think we can do better. Maybe I will refactor it later once I have the CNN version, I will have more than one agent to think about how to generalize and make it cleaner. For now all the mess has been done for a single agent :(

muralx:
Thanks @alejandromarcu, please have a look at the latest changes and I will merge after.

deep_quoridor/src/agents/core/agent.py
```diff
@@ -107,7 +107,7 @@ def end_game(self, game):
         """
         pass
 
-    def get_action(self, observation, action_mask) -> int:
```
jonbinney, line 110: I'm fine with this change, but here are the things I considered when choosing to pass observation and action_mask separately:
- having an observation dict  inside the observation dict the way petting zoo does is awkward and confusing
- petting zoo also allows environments to put the action_mask inside the "info" dict instead of the "observation" dict, so this isn't a convention that is always followed (although we don't need to care about what other environments do)
- we only really _need_ to deal with the nested observation dictionary in the environment itself, so we could choose to isolate it there and pass observation and action_mask separately everywhere else (arena, agents, helper functions).

But also arguing the other way:
- the action_mask is kind of a part of the observation, since it is something you learn about your environment. although in that case "action_mask" should probably be alongside all the other keys in the observation instead of having a nested "observation" key that contains all the other stuff

adamantivm: (out of context comment)
For the record, this decision is really a matter of convention.
In fact if you look at the sb3_ppo.py implementation, a big part of the need for a wrapper is to adapt between the petting zoo convetion (action mask as part of observation) to the gymnasium convention (action mask goes separate to the observation).

muralx: I am fine with either, but we had both ways of using the observation and that was causing problems since we needed to handle observation differently depending on the method being invoked. (get action: obs w/o mask, handle outcome(s): full observation). It was easier to change that one than to change everything else.

muralx: (I also considered action mask is part of the state, hence the observation, but that is a grey area)

deep_quoridor/src/agents/__init__.py
```diff
@@ -26,6 +27,7 @@
 from agents.flat_dqn import FlatDQNAgent  # noqa: E402
 from agents.greedy import GreedyAgent  # noqa: E402, F401
 from agents.human import HumanAgent  # noqa: E402, F401
+from agents.pagent import NewDexpAgent
```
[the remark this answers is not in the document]
muralx: I think this should be fixed now. (I did not see this comment before)

deep_quoridor/src/agents/core/agent.py
jonbinney: Yeah, it does feel like part of the observation. Works for me!

PR #185, opened 2025-05-05
Adds support for different version of the NN

muralx:
- New NN using nn.Dropouts achieving better performance (version 4)
- Keeping previous version as V3
- Making TrainingStatusRenderer single step and formatting output better to see more details of each episode

### DIFF
--- a/deep_quoridor/src/agents/dexp.py
+++ b/deep_quoridor/src/agents/dexp.py
@@ -15,7 +15,7 @@ class DExpNetwork(nn.Module):
     Takes observation from the Quoridor game and outputs Q-values for each action.
     """
 
-    def __init__(self, board_size, action_size, split_board, include_turn):
+    def __init__(self, board_size, action_size, split_board, include_turn, nn_version):
         super(DExpNetwork, self).__init__()
 
         # Calculate input dimensions based on observation space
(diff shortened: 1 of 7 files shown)


deep_quoridor/src/agents/dexp.py
```diff
@@ -124,7 +140,13 @@ def _calculate_action_size(self):
 
     def _create_network(self):
         """Create the neural network model."""
-        return DExpNetwork(self.board_size, self.action_size, self.params.split, self.params.turn)
+        return DExpNetwork(
+            self.board_size,
+            self.action_size,
+            self.params.split,
+            self.params.turn,
+            self.params.nn_version if self.params.nn_version is not None else "4",
```
alejandromarcu, line 148: I think you could just pass `self.params.nn_version` here and in the args initialize it with a default of 4: ` nn_version: str = "4"`

alejandromarcu (approved):
When you use play, does it get the correct NN version?  In theory it should store it in the artifact when training and then use it when playing

deep_quoridor/src/agents/dexp.py
muralx: Donve

muralx: It actually does not work, since it seems the value is for when nothing is specified, which is not the case if the value passed is none.

PR #186, opened 2025-05-05
Improvements for using wandb in training

alejandromarcu:
This will give us more flexibility when using wandb when running train.py
- By default, wandb is not going to be used (as opposed to now that is used unless --no-wandb is passed)
- It can be turned on just by passing `-w`, and the user name in the os will be used as the prefix
- Multiple parameters can be passed to the plugin, e.g `-w prefix=cucu,project=testing_xyz,notes="this time using abc",upload_model=False`
  - You can now pass a different project (by default is deep_quoridor, the one we're using now), so you can use a new project (will be created automatically if it doesn't exist) to try out whatever, and avoid poluting the main one.  Additionally, it's very easy to just delete the whole project once you're done.
  - You can pass notes that will be visible in the overview of the run, so you can easily remember wtf you were trying to do
  - You can skip uploading the model, which is useful when you're debugging or trying out stuff

### DIFF
--- a/deep_quoridor/src/plugins/wandb_train.py
+++ b/deep_quoridor/src/plugins/wandb_train.py
@@ -1,21 +1,47 @@
+import datetime
+import getpass
 import os
-from dataclasses import asdict
+from dataclasses import asdict, dataclass
 
 from agents.core.trainable_agent import AbstractTrainableAgent
 from arena_utils import ArenaPlugin
 from metrics import Metrics
 from utils import resolve_path
+from utils.subargs import SubargsBase
 
 import wandb
 
 
+@dataclass
+class WandbParams(SubargsBase):
+    # Prefix for this run. This is used to create a unique run id for naming, and tagging artifacts and files
+    prefix: str = getpass.getuser()
+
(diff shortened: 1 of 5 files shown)


deep_quoridor/src/plugins/wandb_train.py
```diff
@@ -31,12 +57,13 @@ def _initialize(self, game):
         self.metrics = Metrics(game.board_size, game.max_walls)
 
         self.run = wandb.init(
-            project="deep_quoridor",
+            project=self.params.project,
```
muralx, line 60: don't need it now, but we should consider training based on an existing model, which could lead to 2 different projects to be used.

alejandromarcu: you could pass 2 different projects now, because one is passed in the plug in and the other in the agent, e.g.
`--players dexp:wandb_project=a,wandb_alias=best -w project=b`

This would load the model from the project a, alias best, and save it to project b, with the alias being the latest version.  We could consider adding an option to pass additional aliases, but I feel like usually you don't know if you want to tag it as 'best' or something until after you ran it.  

Having said that, I'm not sure if this works, I'll need to test it

PR #187, opened 2025-05-05
Upgrade Greedy to use features of the new env

alejandromarcu:
Some of the things I was doing manually from the action_mask now can be done much easier using the Quoridor object.  Thanks Jon! Functionallity shouldn't have changed.

### DIFF
--- a/deep_quoridor/src/agents/greedy.py
+++ b/deep_quoridor/src/agents/greedy.py
@@ -47,15 +47,6 @@ class GreedyAgent(Agent):
     def start_game(self, game, player_id):
         self.player_id = player_id
 
-    def _valid_pawn_movements(self, action_mask: np.ndarray) -> list[Position]:
-        """Returns the coordinates of the possible movements based on the action mask.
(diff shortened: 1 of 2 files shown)


jonbinney (approved):
LGTM. If the shortest path function the greedy agent uses in its `_shortest_path_from()` function is more efficient than `Quoridor._bfs` we could replace `Quoridor._bfs` with it so we're using the best shortest path algorithm we have in all cases. (in some future PR)

alejandromarcu:
> LGTM. If the shortest path function the greedy agent uses in its `_shortest_path_from()` function is more efficient than `Quoridor._bfs` we could replace `Quoridor._bfs` with it so we're using the best shortest path algorithm we have in all cases. (in some future PR)

The shortest path function also uses bfs and it's similar, but in Greedy it gives the whole path (not just the minimun distance), so I'd expect it to be a bit less efficient.  In the future we can try to unify them in a way that keeps the efficiency.

PR #188, opened 2025-05-06
Fixes PPO for opponent play in new env

adamantivm:
A small fix I missed when I merged with main. It otherwise seems to work.

### DIFF
--- a/deep_quoridor/src/agents/sb3_ppo.py
+++ b/deep_quoridor/src/agents/sb3_ppo.py
@@ -75,7 +75,8 @@ class SB3ActionMaskWrapper(BaseWrapper):
             unwrapped_env = self.unwrapped
 
             # Get opponent action - the opponent handles the raw environment
-            raw_opponent_action = self.opponent.get_action(unwrapped_env)
+            raw_action_mask = self.unwrapped.observe(self.agent_selection)["action_mask"]
+            raw_opponent_action = self.opponent.get_action(unwrapped_env, raw_action_mask)
 
             # Apply rotation if needed (since we're going to use super().step which doesn't handle rotation)
             # Check if the opponent is player_1 who needs rotation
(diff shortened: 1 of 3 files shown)


deep_quoridor/src/agents/sb3_ppo.py
```diff
@@ -75,7 +75,8 @@ def step(self, action):
             unwrapped_env = self.unwrapped
 
             # Get opponent action - the opponent handles the raw environment
-            raw_opponent_action = self.opponent.get_action(unwrapped_env)
+            raw_action_mask = self.unwrapped.observe(self.agent_selection)["action_mask"]
```
muralx, line 78: General comment: If the opponent should operate in an unwrapped environment, shouldn't all operations related to the environment and used by the opponent be done on the unwrapped environment (In particular step)?

deep_quoridor/src/agents/sb3_ppo.py
muralx, line 1: It does not allow me to write comments where there are no edits! I think the next_state that should be stored, is the next state after opponent has moved. So the p(s', s,a) is calculated properly.

deep_quoridor/src/agents/sb3_ppo.py
muralx: You would not need to rotate either.

muralx (commented):
Some comments from quickly looking at the current state of the implementations

deep_quoridor/src/agents/sb3_ppo.py
adamantivm: to which pice of the code are you referring to? I can't undrestand this comment.

deep_quoridor/src/agents/sb3_ppo.py
adamantivm: I need to first remember why did I have to use the unwrawpped environment in the first place :thinking:

deep_quoridor/src/agents/sb3_ppo.py
adamantivm: You're right. I actually don't know very well how it worked before. There's no noticeable difference in performance after the fix. It's really a mystery to me.

PR #189, opened 2025-05-06
Renderer plugin for showing agent computation times

jonbinney:
Keeps track of how long each agent takes to choose an action, and prints out stats at the end. I've added it to the default benchmarks. Example output:
```
+-------------+-------------------+---------------+---------------+
|    Player   | Average Time (ms) | Min Time (ms) | Max Time (ms) |
+-------------+-------------------+---------------+---------------+
|    sb3ppo   |       2.577       |     1.453     |     6.769     |
|   simple    |       63.338      |     0.923     |    196.253    |
|    greedy   |       1.042       |     0.743     |     1.307     |
|  greedy-03  |       0.886       |     0.288     |     1.801     |
| dexp (1011) |       0.360       |     0.327     |     0.446     |
|  greedy-01  |       1.258       |     0.871     |     1.580     |
|    random   |       0.039       |     0.038     |     0.041     |
+-------------+-------------------+---------------+---------------+
```

### DIFF
--- a/deep_quoridor/src/arena_utils.py
+++ b/deep_quoridor/src/arena_utils.py
@@ -4,6 +4,13 @@ from agents import Agent
 
 
 @dataclass
+class MoveInfo:
+    player: str
+    action: int
+    computation_time: float  # in seconds
+
+
+@dataclass
 class GameResult:
     player1: str
     player2: str
@@ -11,6 +18,7 @@ class GameResult:
     steps: int
     time_ms: int
     game_id: str
+    moves: list[MoveInfo]
 
 
 class ArenaPlugin:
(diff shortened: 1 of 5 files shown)


deep_quoridor/src/arena_utils.py
```diff
@@ -3,6 +3,13 @@
 from agents import Agent
 
 
+@dataclass
+class MoveInfo:
```
muralx, line 7: Do you need to store every single move ? Can you just compute the sum instead of adding the move ?

muralx (approved):

deep_quoridor/src/arena_utils.py
jonbinney: The sum wouldn't be enough to get the min and max, right? It seems cheap enough to store them all and this also allows adding other stats in the renderer without changing the GameResult data structure.

PR #192, opened 2025-05-07
Allow DExp to train against the current model being trained

muralx:
- Not the cleanest approach but achieves its goal
- It improved training reaching a new ELO high


adamantivm:
I suppose you're reading about self-play best practices of continuing to swap the opponent with newer versions of itself over time?

LGTM

alejandromarcu (approved):
An idea for the future to allow to do things like this in a cleaner way: the AgentRegistry, instead of directly instantiating the agents, would call a create method on the agent class, which could return more than one agent.  This way you would just pass an extra argument to dexp to indicate that you want to run with a mimic, and the create method would take care of creating the 2 instances

PR #194, opened 2025-05-07
Logging abs elo for charting purposes

muralx:


muralx:
@alejandromarcu Does it make sense to have it ?

alejandromarcu (approved):
Sure, if you think it's useful let's add it, and in the future we can see which metrics are more useful and maybe remove the others.

PR #196, opened 2025-05-08
Train against a list of opponents

adamantivm:
NOTE: This one builds on top of #195 please review and merge that one first.

This PR adds the ability to train against a predefined list of opponents (for now specified in code).
It also starts tracking winrate win wandb which comes handy to understand what's going on, as well as to confirm that the training agent is getting some wins to learn from as the opponents get tougher.

### DIFF
--- a/deep_quoridor/src/train_sb3.py
+++ b/deep_quoridor/src/train_sb3.py
@@ -28,6 +28,7 @@ from sb3_contrib import MaskablePPO
 from sb3_contrib.common.maskable.policies import MaskableActorCriticPolicy
 from sb3_contrib.common.wrappers import ActionMasker
 from stable_baselines3.common.callbacks import BaseCallback, CallbackList
+from stable_baselines3.common.monitor import Monitor
 from utils import resolve_path, set_deterministic
 
 import wandb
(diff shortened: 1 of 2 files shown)


deep_quoridor/src/train_sb3.py
```diff
@@ -58,6 +78,29 @@ def _on_rollout_end(self):
         else:
             self.current_player = "player_0"
         self.env.set_player(self.current_player)
+        self.current_opponent_games_left -= 1
```
muralx, line 81: Is each rollout equal to 1 game? Also, do you expect to run opponents in a sequence ? We are doing random on the other trainings.

deep_quoridor/src/train_sb3.py
```diff
@@ -80,7 +123,7 @@ def train_action_mask(env_fn, steps=10_000, seed=0, upload_to_wandb=False, train
         sync_tensorboard=True,
     )
 
-    masked_env = ActionMasker(env, mask_fn)  # Wrap to enable masking (SB3 function)
+    masked_env = Monitor(ActionMasker(env, mask_fn))  # Wrap to enable masking (SB3 function)
```
muralx, line 126: Maybe add a comment for the rest of the audience, so it is easy to see why it is needed

muralx (approved):

deep_quoridor/src/train_sb3.py
adamantivm: A rollout is a fixed number of steps, which is typically multiple games. I renamed the variable to make it more consistent, thanks for the comment.

In this PR, I'm on purpose trying to train sequentially. This is part of the exploration I'm doing, trying to understand what's the best strategy to choose opponents for playing.
I think this merits and interesting discussion and is probably worth reading about.

PR #199, opened 2025-05-14
Fix file problems when downloading models from wandb

alejandromarcu:
Trying to move the file fails if the temporary dir is in another device, so just copy it.
Also, it was failing if the `wandbmodels` directory didn't exist


adamantivm:
LGTM

adamantivm:
FYI @shoemakerlevy9 you can try this again now

PR #200, opened 2025-05-14
Fix GPU memory leak

alejandromarcu:
In this PR: https://github.com/jonbinney/deep_rabbit_hole/pull/174 I removed the call to `.item()` on the loss because it forced a sync from the GPU (so it had to wait for operations to complete) and made it slower.
But now I was trying to train a model with a bigger batch and got out of GPU memory.  The cause seems to be this, because since we're storing the tensor and not dettaching it, the whole computation graph is kept in memory.  @adamantivm had some crashes that may have been related to this!
To fix this, in this PR I collect the losses for the episode without using `item()`, and at the end of the game I compute the mean loss for the episode and free the memory by clearing the list with the tensors.
Additionally, this fixes another bug: we were storing in the `train_call_losses` all the losses, but we have many per episode (I think one per move), and then when we average we take the last `length`, that is supposed to be in number of episodes, so we're not actually using the loss of the episode, just of the last `length` moves in the episode.  Now the losses are smoother, e.g. orange was before this fix and green with the fix.

image

### DIFF
--- a/deep_quoridor/src/agents/core/trainable_agent.py
+++ b/deep_quoridor/src/agents/core/trainable_agent.py
@@ -67,7 +67,7 @@ class TrainableAgentParams(SubargsBase):
     # Learning rate to use for the optimizer
     learning_rate: float = 0.001
     # Specifies the NN version to use
-    nn_version: str = None
+    nn_version: Optional[str] = None
 
     @classmethod
     def training_only_params(cls) -> set[str]:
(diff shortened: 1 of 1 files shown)


deep_quoridor/src/agents/core/trainable_agent.py
```diff
@@ -312,7 +320,7 @@ def _handle_step_outcome_all(
         if len(self.replay_buffer) > self.batch_size:
             loss = self._train(self.batch_size)
             if loss is not None:
-                self.train_call_losses.append(loss)
+                self.current_episode_losses.append(loss)
```
muralx, line 323: Losses and episodes are not really related, so not sure it makes sense to name it like that.

muralx (approved):

deep_quoridor/src/agents/core/trainable_agent.py
alejandromarcu: Good point, I'll rename it.  I could go back to the previous name, but why "call"?

muralx: Maybe just use the following names , suggestion in case you can't find better names
avg_training_losses
training_losses_batch?

PR #201, opened 2025-05-15
Use numba to speedup simple agent

jonbinney:
The Quoridor game logic that was taking up the most computation time is re-implemented using numba, which just-in-time compiles it.

Good stuff:
- On my computer, the average running time of the Simple agent with default parameters goes from 24 ms to about 5 ms
- More speedup is possible. I think another 10x or so if the minimax implementation in the simple agent itself is numba-fied, and parallel processing is enabled

Meh stuff:
- To get much benefit from numba, we now do more batch processing in free functions. This blurs some of the abstraction provided by the Quoridor class (for example the numba grid functions in `qgrid.py` do action encoding/decoding themselves instead of using the ActionEncoder class)
- Compilation by numba takes a couple seconds each time `play.py` is run. This doesn't seem to get cached on disk and reused the next time the python interpreter is restarted.
- Numba works well for what it does, but has a lot of magic, doesn't feel particularly actively used or developed, and has limitations that would make we not want to use it in a more "serious" project.

A couple more notes:
- I've added tie-breaking of the minimax reward function using the number of walls remaining (suggested by @alejandromarcu when reviewing that last PR). This gives a nice bump to the results of the simple agent without slowing it down.
- We'd probably be better off removing the `Board` class and just calling `qgrid.py` functions directly from the Quoridor class. Three levels of abstraction feels like too much, but I didn't want to change too much in this PR so I left it for now.
- For the moment I'm using numba-fied BFS for both shortest path and for "does this wall block a player's path to goal". For blocking checks, we can use DFS instead, but it wasn't a bottleneck so I didn't do that yet.

### DIFF
--- a/deep_quoridor/src/agents/simple.py
+++ b/deep_quoridor/src/agents/simple.py
@@ -33,20 +33,16 @@ class SimpleParams(SubargsBase):
     discount_factor: float = 0.99
 
 
-def adjust_wall_position(position: tuple[int, int]) -> tuple[float, float]:
-    """
-    Compute a wall position that reflects where its center is relative to player positions.
-    """
-    return position[0] + 0.5, position[1] + 0.5
-
-
-def compute_wall_weight(wall: WallAction, position_1: tuple[int, int], position_2: tuple[int, int], sigma) -> float:
-    wall_position = np.array(adjust_wall_position(wall.position))
-    d1 = np.linalg.norm(np.array(position_1) - wall_position)
-    d2 = np.linalg.norm(np.array(position_2) - wall_position)
-    min_distance = min(d1, d2)
-    weight = np.exp(-0.5 * (min_distance / sigma) ** 2)
-    return weight
+def compute_wall_weights(
+    wall_actions: list[WallAction], position_1: tuple[int, int], position_2: tuple[int, int], sigma
+) -> float:
+    walls_array = np.array([wall.position for wall in wall_actions], dtype=np.float32)
+    walls_array += (0.5, 0.5)  # Adjust wall positions to reflect where their center is in player coordinates.
+    squared_distances_to_position_1 = ((walls_array - position_1) ** 2).sum(axis=1)
+    squared_distances_to_position_2 = ((walls_array - position_2) ** 2).sum(axis=1)
+    min_squared_distances = np.minimum(squared_distances_to_position_1, squared_distances_to_position_2)
+    weights = np.exp(-0.5 * min_squared_distances / sigma**2)
+    return weights
 
 
 def sample_actions(game: Quoridor, n: int, wall_sigma: float | None = None) -> list[Action]:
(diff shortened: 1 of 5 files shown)


alejandromarcu:
Wow, very nice!
In my computer I saw Simple's average time coming down from 85 to 7.7, so more than 10x!
One thing that I noticed is that the functions that you numba-fied don't appear anymore in the profiler, so it's hard to continue profiling.   One easy solution if you're in Python 3.12 is that when you want to profile those functions, you do `export NUMBA_ENABLE_SYS_MONITORING=1` and then they appear.

One thing that I noticed when profiling is that `index_to_action` appears in the 4th position, which seemed to much for such a simple function:
```
    20901    0.611    0.000    0.846    0.000 ffi.py:195(__call__)
625363/104832    0.337    0.000    0.530    0.000 ir.py:321(_rec_list_vars)
3982182/3982176    0.307    0.000    0.371    0.000 {built-in method builtins.isinstance}
   490365    0.254    0.000    0.431    0.000 quoridor.py:57(index_to_action)
82811/283    0.198    0.000    3.542    0.013 simple.py:83(choose_action)
```
The thing is that every time you call it, it is instantiating an object, which is probably the slow part.  Notice that the objects are immutable and there's a limited quantity of possible objects, so a very easy solution that I tried is to add `@cache` to index_to_action (you need to do `from functools import cache`), and that gave me a 10% bump in speed for simple (now at 7ms). With this we'll just create about 200 objects rather than 490k (it was in just 10 games), which not sure if will have any impact in garbage collection or memory but won't hurt. 

> * Compilation by numba takes a couple seconds each time `play.py` is run. This doesn't seem to get cached on disk and reused the next time the python interpreter is restarted.
I think usually is not a big deal.  I saw there are options to cache the compilation but didn't read much.


> * I've added tie-breaking of the minimax reward function using the number of walls remaining (suggested by @alejandromarcu when reviewing that last PR). This gives a nice bump to the results of the simple agent without slowing it down.
Yey!

> * For the moment I'm using numba-fied BFS for both shortest path and for "does this wall block a player's path to goal". For blocking checks, we can use DFS instead, but it wasn't a bottleneck so I didn't do that yet.
DFS may be a bit faster, but not sure if it would give a significant gain to make it worth implementing as well, since now everything is much faster

deep_quoridor/src/agents/simple.py
```diff
@@ -135,7 +125,19 @@ def choose_action(
...
+    # "player" tries to maximize the reward, while "opponent" tries to minimize it.
+    values = np.array(values)
+    if current_player_before_action == player:
+        best_value = values.max()
+    else:
+        best_value = values.min()
+
+    best_action_indices = np.flatnonzero(values == best_value)
+    # There may be multiple actions with the same value, so we randomly choose one of them.
+    # chosen_action_index = np.random.choice(best_action_indices)
```
alejandromarcu, line 137: Any reason to use the last rather than picking a random one?

jonbinney: Ah, great catch! Didn't mean to leave that in; fixed now. Interestingly fixing that makes simple a bit better as P2 but a bit worse as P1.....

jonbinney:
Added `@cache` to `index_to_action` and it did help a bit, thanks!

@alejandromarcu are you ok with merging this? I'll have another PR doing a bit more optimization in a few days.

PR #202, opened 2025-05-15
wandb: save best model, trigger metrics, ctrl-c

alejandromarcu:
From Diego's wishlist for wandb:
- When we save the model and compute the metrics, keep track of which one scored the highest in relative elo and (unless is the same as the final model) upload it to wandb
- When training, if you use for example: ` --trigger-metrics 95 100` , whenever 95 matches are won out of the last 100 during training, the model will be saved and the metrics computed.  Then, we clear the state so that we don't trigger it right after if there's another win, so you'll need at least 95 more matches to trigger another save.
- If you press ctrl-c or an exception is raised, it's caught and we still call end_arena, giving a chance to upload to wandb and other stuff.  An additional adavantage is that if you do ctrl-c during play it will show the results so far.

It's supposed to tag the best model with "{run_id}-best", but it's not working.  I think it may be related to the issue wandb is dealing with right now, so I don't think it's worth debugging more unless tags are not working after it's fixed.

### DIFF
--- a/deep_quoridor/src/arena.py
+++ b/deep_quoridor/src/arena.py
@@ -147,33 +147,43 @@ class Arena:
         match_id = 1
         results = []
 
-        if mode == PlayMode.ALL_VS_ALL:
-            for i in range(len(agents)):
-                for j in range(i + 1, len(agents)):
-                    for t in range(times):
+        try:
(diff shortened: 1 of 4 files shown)


muralx (approved):
Thanks Cucu!

PR #206, opened 2025-05-20
Prep for AlphaZero

alejandromarcu:
Adds to the training and the arena the possibility for self play, and defines the skeleton for AlphaZero

### DIFF
--- a/deep_quoridor/src/agents/alphazero.py
+++ b/deep_quoridor/src/agents/alphazero.py
@@ -1,0 +1,88 @@
+import os
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Optional, Tuple
+
+from quoridor import Quoridor
+from utils.subargs import SubargsBase
+
+from agents.core.trainable_agent import SelfPlayTrainableAgent
+
+
+@dataclass
+class AlphaZeroParams(SubargsBase):
+    # Just used to display a user friendly name
+    nick: Optional[str] = None
+    # If wandb_alias is provided, the model will be fetched from wandb using the model_id and the alias
+    wandb_alias: Optional[str] = None
+    # When loading from wandb, the project name to be used
+    wandb_project: str = "deep_quoridor"
+    # If a filename is provided, the model will be loaded from disc
+    model_filename: Optional[str] = None
+    # Directory where wandb models are stored
+    wandb_dir: str = "wandbmodels"
+    # Directory where local models are stored
+    model_dir: str = "models"
+
+    # After how many self play games we train the network
+    train_every: int = 100
+
+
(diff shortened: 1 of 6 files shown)


deep_quoridor/src/arena.py
```diff
@@ -129,6 +129,32 @@ def _play_game(self, agent1: Agent, agent2: Agent, game_id: str) -> GameResult:
         self.game.close()
         return result
 
+    def _self_play_game(self, agent: SelfPlayTrainableAgent, game_id: str) -> GameResult:
```
muralx, line 132: Don't you need to duplicate lot of what is in the arena (and arena plugins) already by doing self-play in a different way? Tracking, monitoring, renderers, all needs to be handled in the agent itself with this approach?

Wondering whether there is some other way of doing self play by following the same pattern we have been following?

deep_quoridor/src/agents/alphazero.py
```diff
@@ -0,0 +1,88 @@
...
+        # TO DO
+        # self.online_network.load_state_dict(torch.load(path, map_location=my_device()))
+
+    # def end_game(self, game):
+
+    def compute_loss_and_reward(self, length: int) -> Tuple[float, float]:
+        # TO DO
+        return 0.0, 0.0
+
+    def self_play(self, game: Quoridor) -> tuple[str, int]:
```
muralx, line 74: The agent should not care about whether it is self-playing or not, right? Why do we need a new method for this?

adamantivm: I agree. At a first glance, this seems to be mixing training with agent.

adamantivm:
I also get a similar reaction to @muralx . I feel like the "self training" should be a higher-level mechanism that simply exercises which agents it plays agains each other , without the agents really needing to implement something new. 
I would imagine something like the Arena dynamically switching which agents it runs against each other. It first runs an agent in training mode against random, for X amount of time, then it gets the current version of the agent being trained, and swaps the random agent for that one and keeps the tournament going, etc. something like that. WDYT?

alejandromarcu:
> I also get a similar reaction to @muralx . I feel like the "self training" should be a higher-level mechanism that simply exercises which agents it plays agains each other , without the agents really needing to implement something new. I would imagine something like the Arena dynamically switching which agents it runs against each other. It first runs an agent in training mode against random, for X amount of time, then it gets the current version of the agent being trained, and swaps the random agent for that one and keeps the tournament going, etc. something like that. WDYT?

The self-play means that the agent is playing against the same instance of itself (e.g. same NN). I'll experiment with another approach, either having the arena using the same agent twice, or having 2 different AlphaZeroAgent sharing parameters.

PR #207, opened 2025-05-21
2nd try: AlphaZero prep

alejandromarcu:
Basic skeleton for AlphaZero, and some framework changes to allow to have the 2 instances related. 
E.g. you would train it with: `/train.py -p alphazero:...`, and internally it would create 2 instances of the AlphaZero agent, one in training mode and one in play mode, and they will share the NN, temperature and probably other parameters as we realize we need.

### DIFF
--- a/deep_quoridor/src/agents/alphazero.py
+++ b/deep_quoridor/src/agents/alphazero.py
@@ -1,0 +1,104 @@
+import os
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Optional, Tuple
+
+import torch.nn as nn
+from utils.subargs import SubargsBase
+
+from agents.core.trainable_agent import TrainableAgent
+
+
+@dataclass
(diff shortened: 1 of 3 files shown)


deep_quoridor/src/agents/alphazero.py
```diff
@@ -0,0 +1,104 @@
...
+        super().__init__(**kwargs)
+        self.board_size = board_size
+        self.max_walls = max_walls
+        self.params = params
+        self.episode_count = 0
+        self.training_mode = training_mode
+
+        # If a training instance is passed, this instance is playing to train the other, sharing the NN and temperature
+        if training_instance:
+            self.nn = training_instance.nn
```
jonbinney, line 47: I like this concept of a "training instance" object with the context needed

jonbinney (approved):
This looks good to me. We can always modify it as needed.

PR #209, opened 2025-05-23
MCTS starting point

alejandromarcu:
based on https://github.com/foersterrobert/AlphaZeroFromScratch/blob/main/2.MCTS.ipynb

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -19,6 +19,15 @@
             "justMyCode": false
         },
         {
+            "name": "Python Debugger: MCTS debug",
+            "type": "debugpy",
+            "request": "launch",
+            "program": "${file}",
+            "console": "integratedTerminal",
+            "justMyCode": false,
+            "args": "-N 5 -W 3 -p human mcts -t 1 -r pygame -i 26"
+        },
+        {
             "name": "Test water level model (dataset)",
             "type": "debugpy",
             "request": "launch",
--- a/deep_quoridor/src/agents/alphazero.py
+++ b/deep_quoridor/src/agents/alphazero.py
@@ -1,12 +1,19 @@
+import math
 import os
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Optional, Tuple
 
+import torch
 import torch.nn as nn
+from numpy import copy
+from quoridor import Action, Quoridor
+from quoridor_env import QuoridorEnv
+from utils import my_device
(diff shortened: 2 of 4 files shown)


.vscode/launch.json
```diff
@@ -18,6 +18,15 @@
             "console": "integratedTerminal",
             "justMyCode": false
         },
+        {
+            "name": "Python Debugger: MCTS debug",
+            "type": "debugpy",
+            "request": "launch",
+            "program": "${file}",
```
adamantivm, line 25: Can you point this specifically to play.py?

deep_quoridor/src/agents/alphazero.py
```diff
@@ -1,12 +1,19 @@
+import math
```
adamantivm, line 1: This file is still WIP, right? Could you add a high-level comment clarifying that? Ideally a quick sentence on what is done and what is missing, which can be updated / removed when this work is completed.

Alternatively, why don't you omit these changes on this PR, and only send the MCTS part here? That may be better, so that the PR shows more clearly only what was needed to implement MCTS

deep_quoridor/src/agents/mcts.py
```diff
@@ -23,11 +23,16 @@ def __init__(self, game: Quoridor, parent: Optional["Node"] = None, action_taken
         self.depth = depth
         self.winner = ""
         # TODO maybe better to do it lazily
-        self.valid_actions = list(game.get_valid_actions())
-        random.shuffle(self.valid_actions)
+        # self.valid_actions = list(game.get_valid_actions())
```
adamantivm, line 26: I think this was unnecessary in the end, can you revert it?

deep_quoridor/src/agents/mcts.py
```diff
@@ -23,11 +23,16 @@ def __init__(self, game: Quoridor, parent: Optional["Node"] = None, action_taken
...
+        # random.shuffle(self.valid_actions)
+
+        wall_actions = game.get_valid_wall_actions()
+        random.shuffle(wall_actions)
+        self.valid_actions = game.get_valid_move_actions() + wall_actions
+
         # print("Created a node, valid actions:", self.valid_actions)
```
adamantivm, line 33: remove old code / comments, maybe?

alejandromarcu:
Sorry, I just pushed this to have it stored somewhere! I'm working on MCTS on a separate branch.

PR #215, opened 2025-06-04
Make ActionEncoder a singleton-ish

alejandromarcu:
We call ActionEncoder.index_to_action a bazillion times, and returning a new object each time when there's a very limited set of possible objects takes a lot of memory and is slow.  
We had it cached before but when using deepcopy to copy the games, it was copying the ActionEncoder object and therefore we were losing the cache, making it use a lot of memory.
With this approach, we can only have 1 ActionEncoder instance per board size, even when using deepcopy.
Running benchmark_B5W3 with this change took 9:32, vs 11:43 before it

### DIFF
--- a/deep_quoridor/src/quoridor.py
+++ b/deep_quoridor/src/quoridor.py
@@ -1,6 +1,7 @@
 import copy
 from dataclasses import dataclass
 from enum import IntEnum, unique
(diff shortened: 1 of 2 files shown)


muralx (approved):
Thanks!

PR #218, opened 2025-06-07
simple: param to disable heuristic

jonbinney:
Useful for debugging, since it makes it very obvious when the search depth isn't deep enough to find a win (or loss)


adamantivm (approved):
LGTM

PR #226, opened 2025-06-17
Initial version of MCTS for AlphaZero

alejandromarcu:
AlphaZero uses a modified version of MCTS.   I just copied the existing implementation and modified it.  There are many changes spread out across the class, so it seems that trying to re-use it would be tricky (I know OpenSpiel does it, but haven't checked how they do it).
I haven't tested it yet, but I wanted to send what I did so far in case anyone else is interested in working with this this week, since I won't have a lot of time.

### DIFF
--- a/deep_quoridor/src/agents/alphazero.py
+++ b/deep_quoridor/src/agents/alphazero.py
@@ -68,7 +68,7 @@ class AzNode:
 
     def expand(self, policy_probs: np.ndarray):
         """
-        Create a child of the current node.
+        Create all the children of the current node.
         """
         for action, prob in policy_probs:
             game = copy.deepcopy(self.game)
--- a/deep_quoridor/src/agents/alphazero.py
+++ b/deep_quoridor/src/agents/alphazero.py
@@ -1,12 +1,150 @@
+import copy
 import os
+import random
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Optional, Tuple
 
 import torch.nn as nn
+from quoridor import Action, Quoridor
 from utils.subargs import SubargsBase
 
 from agents.core.trainable_agent import TrainableAgent
+
+
+class Node:
+    def __init__(
+        self, game: Quoridor, parent: Optional["Node"] = None, action_taken: Optional[Action] = None, ucb_c: float = 1.0
+    ):
+        self.game = game
+        self.parent = parent
+        self.action_taken = action_taken
+
+        self.children = []
+        self.visit_count = 0
+        self.value_sum = 0.0
+
+        self.valid_actions = list(game.get_valid_actions())
+        random.shuffle(self.valid_actions)
+
+        self.ucb_c = ucb_c
+
+    def should_expand(self):
+        return len(self.valid_actions) > 0 or len(self.children) == 0
+
+    def expand(self):
+        """
+        Create a child of the current node.
+        """
+        action = self.valid_actions.pop()
+
+        game = copy.deepcopy(self.game)
+        game.step(action)
+
+        child = Node(game, parent=self, action_taken=action, ucb_c=self.ucb_c)
+        self.children.append(child)
+        return child
+
+    def select(self) -> "Node":
+        """
+        Return the child of the current node with the highest ucb
+        """
+        return max(self.children, key=self.get_ucb)
+
+    def get_ucb(self, child):
+        # value_sum is in between -1 and 1, so doing (avg + 1) / 2 would make it in the range [0, 1]
+        q_value = ((child.value_sum / child.visit_count) + 1) / 2
+        return q_value + self.ucb_c * math.sqrt(math.log(self.visit_count) / child.visit_count)
+
+    def backpropagate(self, value: float):
+        """
+        Update the nodes from the current node up to the tree by increasing the visit count and adding the value
+        """
+        self.value_sum += value
+        self.visit_count += 1
+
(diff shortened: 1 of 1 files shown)


deep_quoridor/src/agents/alphazero.py
```diff
@@ -30,6 +37,119 @@ class AlphaZeroParams(SubargsBase):
     # Exploration vs exploitation.  0 is pure exploitation, infinite is random exploration.
     temperature: float = 1.0
 
+    # Number of searches
+    n: int = 1000
+    # A higher number favors exploration over exploitation
+    c: float = 1.4
+
+
+class AzNode:
```
adamantivm, line 46: :donkey:

deep_quoridor/src/agents/alphazero.py
```diff
@@ -30,6 +37,119 @@ class AlphaZeroParams(SubargsBase):
...
+            # Traverse down the tree guided by maximum UCB until we find a node to expand
+            node = self.select(root)
+
+            if node.game.check_win(good_player):
+                value = 1
+            elif node.game.check_win(1 - good_player):
+                value = -1
+            else:
+                with torch.no_grad():
+                    policy, value = self.nn(node.game)
```
adamantivm, line 135: value is not used during evaluation, right?

deep_quoridor/src/agents/alphazero.py
```diff
@@ -30,6 +37,119 @@ class AlphaZeroParams(SubargsBase):
...
+        for _ in range(self.params.n):
+            # Traverse down the tree guided by maximum UCB until we find a node to expand
+            node = self.select(root)
+
+            if node.game.check_win(good_player):
+                value = 1
+            elif node.game.check_win(1 - good_player):
+                value = -1
+            else:
+                with torch.no_grad():
```
adamantivm, line 134: I think the way OpenSpiel generalizes the MCTS implementation is by plugging in an "Evaluator" here, which returns what the thing needs to expand and back-propagate. 

So in the default version the "evaluator" is the simulation you already implmeented. And in this other verison, the evaluator is the NN evaluator.

I don't think it's necessary to do it in that way, I just mention it in case it's useful to compare the implementaiton to confirm details and clear doubts as we go with our implementation.

deep_quoridor/src/agents/alphazero.py
```diff
@@ -30,6 +37,119 @@ class AlphaZeroParams(SubargsBase):
...
+
+        self.ucb_c = ucb_c
+        self.prior = prior
+
+    def should_expand(self):
+        return len(self.children) == 0
+
+    def expand(self, policy_probs: np.ndarray):
+        """
+        Create a child of the current node.
```
adamantivm, line 71: Here you create all the childs, right? Maybe update the comment to match?

Also consider if you wish, just to keep both evolving in parallel to make them easier to compare, to also update mcts.py to do it in this same way to match. If I understand things correctly, the default behavior would be to expand all with equal probability, maybe? I am probably missing something. Just trying to use the opportunity to understand a bit more, that's all.

adamantivm (approved):
Awesome to see progress with this, cucu!

I left some comments but I realize this is just incremental work, feel free to merge if you want and continue implementaiton as normal.

deep_quoridor/src/agents/alphazero.py
```diff
@@ -30,6 +37,119 @@ class AlphaZeroParams(SubargsBase):
...
+        self.value_sum += value
+        self.visit_count += 1
+
+        if self.parent is not None:
+            self.parent.backpropagate(-value)
+
+
+class AzMCTS:
+    def __init__(self, nn: nn.Module, params: AlphaZeroParams):
+        self.nn = nn
```
muralx, line 108: something to keep in mind. If you use an NN with Dropout (or the like), you might need to use eval() mode when not training or in the instance that is not training.

deep_quoridor/src/agents/alphazero.py
alejandromarcu: Didn't notice, I'm loosing my wittiness.  BTW, at Facebook I once called a class HubsExactIndex, and a while later someone noticed it could be HubSexActIndex 🤣 .  I think linters should start using AI tools to check for those cases and give you a warning

deep_quoridor/src/agents/alphazero.py
alejandromarcu: It's used in the backpropagation below.  It's hard to read because how the diff is presented

deep_quoridor/src/agents/alphazero.py
alejandromarcu: Yeah, the thing is that there are a few more changes spread out across the classes, which would be annoying to have ifs everywhere

deep_quoridor/src/agents/alphazero.py
alejandromarcu: We could change the other MCTS the same way to expand all the nodes (although besides code re-using, I don't think it provides any advantage).   In that one,  it doesn't use priors, so it's an uniform probability.  Basically the advantage here is that it uses the "intuition" of the NN to start exploring the nodes that it estimates are more likely to get to a win.
I'll add a TODO for myself to do the update in the other one

alejandromarcu, line 71:
```suggestion
        Create all the children of the current node.
```

PR #229, opened 2025-06-20
Show median value for computation times

jonbinney:
And add a title to the table. Median computation time is useful for cases where an agent does caching or jit-ing, or does training inside get_action.

### DIFF
--- a/deep_quoridor/src/renderers/computation_times.py
+++ b/deep_quoridor/src/renderers/computation_times.py
@@ -1,3 +1,5 @@
+import statistics
+
 from arena import GameResult
 from prettytable import PrettyTable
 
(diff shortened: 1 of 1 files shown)


jonbinney:
New output format:
```
+-----------------------------------------------+
|             Computation times (ms)            |
+-----------+---------+-------+--------+--------+
|   Player  | Average |  Min  |  Max   | Median |
+-----------+---------+-------+--------+--------+
| alphazero |  21.109 | 2.603 | 46.239 | 20.087 |
|   random  |  0.139  | 0.062 | 0.370  | 0.138  |
+-----------+---------+-------+--------+--------+
```

alejandromarcu (approved):

PR #230, opened 2025-06-22
Second pass at alphazero agent

jonbinney:
It does train now, but doesn't make moves that make any sense yet - I'm pretty sure there are multiple bad bugs. Leaving this as a draft PR until I debug a bit more.

### DIFF
--- a/deep_quoridor/src/quoridor.py
+++ b/deep_quoridor/src/quoridor.py
@@ -141,6 +141,16 @@ class Board:
         for player, position in zip(self._players, self._player_positions):
             self.set_player_cell(position, player)
 
+    def rotate_board(self):
+        self._grid = np.rot90(self._grid, k=2)
+        rotated_old_style_walls = np.zeros_like(self._old_style_walls)
+        for i in range(self._old_style_walls.shape[2]):
+            rotated_old_style_walls[:, :, i] = np.rot90(self._old_style_walls[:, :, i], k=2)
+        self._old_style_walls = rotated_old_style_walls
+        self._player_positions = np.array(
+            [(self.board_size - 1 - row, self.board_size - 1 - col) for (row, col) in self._player_positions]
+        )
+
     def get_player_position(self, player: Player) -> tuple[int, int]:
         """
         Get the position of the player's pawn.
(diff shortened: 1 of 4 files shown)


deep_quoridor/src/quoridor.py
```diff
@@ -378,6 +393,22 @@ def get_valid_wall_actions(self, player: Optional[Player] = None) -> list[WallAc
...
+        """
+        if isinstance(action, MoveAction):
+            return MoveAction(
+                (self.board.board_size - 1 - action.destination[0], self.board.board_size - 1 - action.destination[1])
+            )
+        elif isinstance(action, WallAction):
+            return WallAction(
+                (self.board.board_size - 2 - action.position[0], self.board.board_size - 2 - action.position[1]),
+                action.orientation,
+            )
```
alejandromarcu, line 408: Are you sure about this one? I'd think that rotating a horizontal wall would be different than rotating a vertical wall, because the walls are 2 in length and we use the top left corner, so when you rotate you need to account for the wall going "in the opposite direction".  E.g. a horizontal wall in the top left corner of the board would move to the bottom right, but you need to move it one to the left or it will overflow.
I think having tests for this would help

jonbinney:
This is still not performing correctly, but I think its worth merging and splitting up the next set of tasks at our meeting today. In my last commit, I split the alphazero agent into different files for the agent class, the mcts search code, the evaluator, and the network itself. I've put a list of suggested next steps in our meeting doc. Currently I run the code with the following command:
```
python deep_quoridor/src/train.py -N 3 -W 1 -e 100 -i 43 -p alphazero:training_mode=true random -r arenaresults2 computationtimes
```

adamantivm:
Proactively merging this without review before our meeting today

deep_quoridor/src/agents/alphazero/mcts.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def search(self, initial_game: Quoridor):
+        root = Node(initial_game, ucb_c=self.ucb_c)
+        good_player = initial_game.current_player
+
+        for _ in range(self.n):
+            # Traverse down the tree guided by maximum UCB until we find a node to expand
+            node = self.select(root)
+
+            if node.game.check_win(good_player):
```
muralx, line 105: I think the node always belongs to the winner, so value should always be one and backpropagate will take care of the sign.

jonbinney: You're right! I'll PR a fix

PR #233, opened 2025-06-26
Self play for training

alejandromarcu:
If you only pass one player to the train script, it will use the same instance twice for training, to allow for self play.  I added the current player in the observation so the agents can know as which player is playing in each turn.


jonbinney (approved):
That was quick :-)  Looks good!

PR #235, opened 2025-07-01
Fix bugs in numpy<->Tensor conversion

jonbinney:
This was included already in the code @adamantivm and @alejandromarcu and I were debugging today. Without this change, I was seeing errors on a computer where I had both CPU and GPU support for CUDA,.


alejandromarcu (approved):
Thanks! I run into those kind of issues as well but I just hardcoded to use CPU and was going to look later, I thought maybe was a mac gpu issue

PR #237, opened 2025-07-02
Adding DAZ, which is a mix of AlphaZero and DQN (MCTS and DQN)

muralx:
- Add new agent DAZ that was copied from DEXP and mixed with MCTS code (ugly)
-- Includes DQN (softmax) values as prior and use them to compute scores and selection during node expasion (when softmax exploration is enabled)
-- Has a cache to avoid re-exploring paths that have been already explored, just backpropagates the node. For this I added fast hash to the game 
-- It still trains as all past agents using Cucu's initial framework
-- Penalizes expansion of nodes that are already in the path.
-- Logs times per matchresults and for training status.


yargs/play/agent_benchmark.yaml
```diff
@@ -2,7 +2,7 @@ benchmark_B3W0:
   args: -t 100 -N 3 -W 0 --players greedy greedy:p_random=0.1,nick=greedy-ish random sb3ppo:wandb_alias=best -r progressbar arenaresults eloresults computationtimes
 
 benchmark_B5W3:
-  args: -t 100 -N 5 -W 3 --players random greedy greedy:p_random=0.1,nick=greedy-01 greedy:p_random=0.3,nick=greedy-03 dexp:wandb_alias=best simple sb3ppo:wandb_alias=best cnn3c:wandb_alias=best mcts -r progressbar arenaresults eloresults computationtimes
```
muralx, line 5: are you guys find with removing MCTS, it is really slow and it does not perform well at all?

deep_quoridor/src/play.py
```diff
@@ -79,6 +80,7 @@ def player_with_params(arg):
 
     args_dict = yargs(parser, "yargs/play")
 
+    sys.setrecursionlimit(10000)  # Increase recursion limit for deep games
```
jonbinney, line 83: Cool, i didn't know there was a way to change this!

deep_quoridor/src/quoridor.py
```diff
@@ -322,6 +349,7 @@ def rotate_board(self):
         """
         self.board.rotate_board()
         self._goal_rows = self._goal_rows[::-1]
+        self._rotated = not self._rotated
```
jonbinney, line 352: Ah, I like keeping track of this here!

deep_quoridor/src/quoridor_env.py
```diff
@@ -34,6 +34,38 @@
 from pettingzoo import AECEnv
 from pettingzoo.utils import wrappers
 from quoridor import ActionEncoder, Board, Player, Quoridor
+from utils.misc import get_opponent_player_id
+
+
+def make_observation(game, agent_id, player, is_player_turns):
```
jonbinney, line 40: Thanks for making separating this out into its own function - much cleaner!

jonbinney (approved):
LGTM

PR #238, opened 2025-07-02
Apply masking before softmax

adamantivm:
Adds a hack to deal with the case when the masked policy ends up being all zeroes. Now instead, we return a unifirm distribution over the valid actions.

### DIFF
--- a/deep_quoridor/src/agents/alphazero/nn_evaluator.py
+++ b/deep_quoridor/src/agents/alphazero/nn_evaluator.py
@@ -46,21 +46,18 @@ class NNEvaluator:
         # Mask the policy to ignore invalid actions
         valid_actions = game.get_valid_actions()
         valid_action_indices = [self.action_encoder.action_to_index(action) for action in valid_actions]
-        # We mask by setting all the values which are masked to minus infinity. This is because the softmax
-        # function will then assign a probability of zero to those values.
-        # Taken from the OpenSpiel implementation: https://github.com/deepmind/open_spiel/blob/master/open_spiel/python/algorithms/alpha_zero/model.py#L277
         policy_masked = np.zeros_like(unmasked_policy)
         policy_masked[valid_action_indices] = unmasked_policy[valid_action_indices]
 
         if np.all(policy_masked == 0):
-            # If the policy ends up as all zeros after masking, turn it into a uniform distribution amongh
+            # If the policy ends up as all zeros after masking, turn it into a uniform distribution among
             # the valid actions.
             policy_masked[valid_action_indices] = 1 / len(valid_action_indices)
         else:
             # Otherwise, just renormalize after masking
             policy_masked = policy_masked / policy_masked.sum()
 
-        # Validity checks
+        # Sanity checks
         assert np.all(policy_masked >= 0), "Policy contains negative probabilities"
         assert np.all(policy_masked <= 1), "Policy contains probabilities greater than 1"
         assert np.any(policy_masked > 0), "Policy is all zeros"


deep_quoridor/src/agents/alphazero/nn_evaluator.py
```diff
@@ -45,12 +45,16 @@ def evaluate(self, game: Quoridor):
...
-        # Re-normalize probabilities after masking.
-        policy_probs = policy_masked
-        policy_probs = policy_probs / policy_probs.sum()
+        if np.all(policy_masked == 0):
+            print("Policy is all zeros")
+            print(policy_masked)
+            print(game)
+
+        # Apply softmax to convert to probabilities
+        policy_probs = F.softmax(torch.from_numpy(policy_masked), dim=0).numpy()
```
jonbinney, line 57: Hmm... if you take the softmax here, then you need to take the logarithm of the training data you use for the network, right? Otherwise the network is training on probabilities, but you're treating its output as log probabilities.

deep_quoridor/src/agents/alphazero/nn_evaluator.py
```diff
@@ -45,12 +45,16 @@ def evaluate(self, game: Quoridor):
         # Mask the policy to ignore invalid actions
         valid_actions = game.get_valid_actions()
         valid_action_indices = [self.action_encoder.action_to_index(action) for action in valid_actions]
-        policy_masked = np.zeros_like(unmasked_policy)
+        policy_masked = np.full_like(unmasked_policy, -1e32)
```
jonbinney, line 48: `-1x10**32`? Could you add a comment explaining this here?

deep_quoridor/src/agents/alphazero/nn_evaluator.py
```diff
@@ -45,12 +45,16 @@ def evaluate(self, game: Quoridor):
...
         policy_masked[valid_action_indices] = unmasked_policy[valid_action_indices]
 
-        # Re-normalize probabilities after masking.
-        policy_probs = policy_masked
-        policy_probs = policy_probs / policy_probs.sum()
+        if np.all(policy_masked == 0):
+            print("Policy is all zeros")
+            print(policy_masked)
+            print(game)
```
jonbinney, line 54: I'm not a fan of adding debug prints to committed code - it gets out of hand fast.

deep_quoridor/src/renderers/computation_times.py
```diff
@@ -31,6 +31,8 @@ def end_arena(self, game, results: list[GameResult]):
                     f"{max(computation_times_ms):.3f}",
                     f"{statistics.median(computation_times_ms):.3f}",
                 ]
+                if num_moves > 0
```
jonbinney, line 34: Thanks for fixing this! I didn't encounter the problem often since it usually crashed earlier :-)

deep_quoridor/src/agents/alphazero/nn_evaluator.py
alejandromarcu: Won't this give a sligth chance of taking an invalid action? probably they won't be chosen if there's any valid action with a reasonable probability, but in cases like we were debugging where the masked policies were all almost 0, it may end up chosing an invalid action.  I think what we would need to ensure is that some of the valid actions are > 0, we could take one randomly otherwise to avoid division by 0 but I think it's a red flag in the first place that the network is so confidently wrong

deep_quoridor/src/agents/alphazero/nn_evaluator.py
alejandromarcu: Maybe we should be using a logging tool with different levels of log that we can turn on and off, and this should be a warning

jonbinney: I think that once the policy is all zeros, you're completely hosed, right? This could raise an exception.

deep_quoridor/src/agents/alphazero/nn_evaluator.py
adamantivm: :thinking:

deep_quoridor/src/renderers/computation_times.py
adamantivm: :handshake:

deep_quoridor/src/agents/alphazero/nn_evaluator.py
adamantivm: I'll leave the check (after the softmax now) anyway, JIC, but I posit it will never trigger.

deep_quoridor/src/agents/alphazero/nn_evaluator.py
jonbinney: ok i think i may have found it! For the policy loss, they use  [tf.nn.softmax_cross_entropy_with_logits_v2(
            logits=policy_logits, labels=policy_targets), ](https://github.com/google-deepmind/open_spiel/blob/b11f9eb62a67a0982c0183d7170be6411c7447ab/open_spiel/python/algorithms/alpha_zero/model.py#L283), which knows that it is comparing log-probabilities with probabilities. We just use `cross_entropy()` loss, which expects both inputs to be probability distributions.

adamantivm:
@jonbinney wdyt about merging this first and then working on the alleged missing mask rotation in a new PR? can I merge?

jonbinney:
Merging sounds good!

deep_quoridor/src/agents/alphazero/nn_evaluator.py
```diff
@@ -41,21 +41,29 @@ def evaluate(self, game: Quoridor):
...
 
-        # Re-normalize probabilities after masking.
-        policy_probs = policy_masked
-        policy_probs = policy_probs / policy_probs.sum()
-
-        if not (np.isfinite(policy_probs).all() and np.isfinite(value)):
-            raise ValueError("Non-finite number in policy")
-
-        return value, policy_probs
+        if np.all(policy_masked == 0):
```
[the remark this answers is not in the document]
adamantivm: why do you say it will fail the assert below? the content inside the `if` block will make it so that it's no longer all zeroes. Can you clarify @alejandromarcu ?

PR #243, opened 2025-07-04
Fix board rotation for AlphaZero agent

adamantivm:
Fixes a bug in how the decision was made to rotate or not and also moves things around so that the network is always trained and evaluated using the perspective of Player 1

### DIFF
--- a/deep_quoridor/src/agents/alphazero/nn_evaluator.py
+++ b/deep_quoridor/src/agents/alphazero/nn_evaluator.py
@@ -4,7 +4,7 @@ import random
 import numpy as np
 import torch
 import torch.nn.functional as F
-from quoridor import ActionEncoder, Board, Quoridor
+from quoridor import ActionEncoder, Board, Player, Quoridor
 
 from agents.alphazero.mlp_network import MLPNetwork
 
@@ -70,10 +70,10 @@ class NNEvaluator:
         player = game.get_current_player()
         opponent = int(1 - player)
 
-        rotate = True if (player == "player_2") else False
+        rotate = True if (player == Player.TWO) else False
         if rotate:
             game = copy.deepcopy(game)
-            game.rotate()
+            game.rotate_board()
 
         player_position = game.board.get_player_position(player)
         opponent_position = game.board.get_player_position(opponent)
--- a/deep_quoridor/src/agents/alphazero/nn_evaluator.py
+++ b/deep_quoridor/src/agents/alphazero/nn_evaluator.py
@@ -20,7 +20,7 @@ class NNEvaluator:
 
         # Create a temporary game just to see how  big the input tensors are.
         temp_game = Quoridor(Board(self.action_encoder.board_size))
-        temp_input, _ = self.game_to_input_array(temp_game)
+        temp_input = self.game_to_input_array(temp_game)
         self.input_size = len(temp_input)
 
         self.network = MLPNetwork(self.input_size, self.action_encoder.num_actions, self.device)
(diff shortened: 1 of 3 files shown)


adamantivm:
I think it's OK but I would appreciate a careful review.

adamantivm:
I did some step by step debugging of both training and evaluation and everything I see seems to make sense.
However, in a long run, it starts to have a lot of cases where it ends up with an all-zeroes policy, IDK why

jonbinney:
I'll take a look now, and step through it locally a bit to make sure I understand what happens when.

deep_quoridor/src/agents/alphazero/nn_evaluator.py
```diff
@@ -63,18 +63,27 @@ def evaluate(self, game: Quoridor):
         assert np.any(policy_masked > 0), "Policy is all zeros"
         assert np.isfinite(policy_masked).all() and np.isfinite(value), "Policy or value is non-finite"
 
+        # If the game was originally rotated, rotate the resulting back to player 2's perspective
+        if is_board_rotated:
+            policy_masked = policy_masked[self.rotated_action_mapping]
+
         return value, policy_masked
 
+    def rotate_game_for_player_two(self, game: Quoridor):
```
jonbinney, line 72: The name of this function is a bit misleading - given the name, I would expect it to always rotate the board to player two's perspective. Maybe call it `orient_game_for_current_player()`?

jonbinney: Also (I realize Im being a bit hypocritical here since i didnt document the code well originally), future us would probably appreciate a doc string for this function. Something like:
    """
    Rotates the game so that the current player's goal is the row with the largest index.

    This makes it easier for the neural network to learn, since it doesn't need to
    understand that one player wants to move up the board and the other wants to move down it. If the current player is Player.ONE, this is a no-op since the board 
    is already rotated correctly for it, and the same game instance is returned. If
    the current player is Player.TWO, the returned game is a copy of the original game
    which has beeen rotated appropriately.
    """

deep_quoridor/src/agents/alphazero/nn_evaluator.py
```diff
@@ -33,17 +33,16 @@ def __init__(
...
             unmasked_policy, value = self.network(torch.from_numpy(input_array).float().to(self.device))
             unmasked_policy = unmasked_policy.cpu().numpy()
             value = value.item()
 
-        if is_board_rotated:
-            unmasked_policy = unmasked_policy[self.rotated_action_mapping]
-            # TODO: Also rotate the action mask!
-
-        # Mask the policy to ignore invalid actions
+        # Mask the policy to ignore invalid actions. NOTE: Game is already rotated so the valid actions will be rotated too
```
jonbinney, line 45: I like your approach of rotating the game first, but since rotate_board changes the game in place, and python passes everything by reference, it could be confusing to callers of this function that their game changed under them.

deep_quoridor/src/agents/alphazero/nn_evaluator.py
```diff
@@ -63,18 +63,27 @@ def evaluate(self, game: Quoridor):
         assert np.any(policy_masked > 0), "Policy is all zeros"
         assert np.isfinite(policy_masked).all() and np.isfinite(value), "Policy or value is non-finite"
 
+        # If the game was originally rotated, rotate the resulting back to player 2's perspective
+        if is_board_rotated:
+            policy_masked = policy_masked[self.rotated_action_mapping]
```
jonbinney, line 68: Hmmm.... is the rotated_action_mapping that maps from an action on a regular board to an action on a rotated board the same as the that inverts that process? Rotating a board twice does put the board back in its originally orientation, but is the same true for rotating actions in this way...... hmmm.....

jonbinney (commented):
Posting this review now since I need to take a break, but later today I'm going to do some testing.

deep_quoridor/src/agents/alphazero/nn_evaluator.py
adamantivm: But the returned rotated game is a deep_copy of the original one, it's not the originally changed one.
What do you suggest instead?

deep_quoridor/src/agents/alphazero/nn_evaluator.py
```diff
@@ -95,7 +104,7 @@ def game_to_input_array(self, game: Quoridor) -> tuple[torch.FloatTensor, bool]:
             dtype=np.float32,
         )
 
-        return input_array, rotate
+        return input_array
 
     def train_network(self, replay_buffer, learning_rate, batch_size, optimizer_iterations):
         optimizer = torch.optim.Adam(self.network.parameters(), lr=learning_rate)
```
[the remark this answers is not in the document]
adamantivm: Seems like a good catch! Albeit unrelated to this PR.

deep_quoridor/src/agents/alphazero/nn_evaluator.py
adamantivm: Hmm ... I tried it with B3W0 and it seemed to work, but that only tested the moves part, not wall placement ...

deep_quoridor/src/agents/alphazero/nn_evaluator.py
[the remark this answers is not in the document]
jonbinney: I did that somewhat arbitrarily - is creating the optimizer high overhead? Or is there some state in it that is important to preserve between runs?

jonbinney (approved):
I didn't get testing done today, but I think we should merge this. It's fixing a definite bug and I believe the code changes are correct. I'll add unit tests in a separate PR within the next couple of days.

adamantivm:
Merging proactively, we can revisit how the rotation is done in a subsequent PR

PR #245, opened 2025-07-06
Tests to make sure evaluator rotates board

jonbinney:
If the evaluator rotates the board properly for player two, then rotationally symmetric games should give the same policy and value, regardless of whose turn it is.  These tests failed at first, and I realized that it's conventional to call `model.eval()` to put the model in "evaluation" mode before running inference. That disables the dropout, which otherwise would introduce randomness to the results. Then before training, we call `model.train()` to put the model back into training mode.

### DIFF
--- a/deep_quoridor/src/agents/alphazero/nn_evaluator.py
+++ b/deep_quoridor/src/agents/alphazero/nn_evaluator.py
@@ -34,6 +34,8 @@ class NNEvaluator:
         # Rotate the board if player 2 is playing so that we always work with player 1's perspective.
         game, is_board_rotated = self.rotate_if_needed_to_point_downwards(game)
 
+        self.network.eval()  # Disables dropout
+
         with torch.no_grad():
             input_array = self.game_to_input_array(game)
             unmasked_policy, value = self.network(torch.from_numpy(input_array).float().to(self.device))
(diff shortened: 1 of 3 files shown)


deep_quoridor/src/quoridor.py
```diff
@@ -341,6 +341,9 @@ def __init__(
     def create_new(self):
         return Quoridor(self.board.create_new(), self.current_player, self.action_encoder, self._goal_rows)
 
+    def copy(self):
```
adamantivm, line 344:
```suggestion
    def clone(self):
```

Small nit: slightly better name like this?

deep_quoridor/src/agents/alphazero/nn_evaluator.py
```diff
@@ -34,6 +34,8 @@ def evaluate(self, game: Quoridor):
         # Rotate the board if player 2 is playing so that we always work with player 1's perspective.
         game, is_board_rotated = self.rotate_if_needed_to_point_downwards(game)
 
+        self.network.eval()  # Disables dropout
```
adamantivm, line 37: Just to be 100% sure, we *always* want to disable dropout for evaluation, right? 
I mean, I see how we want to do this for testing in order to achieve repeatability. But during actual play / execution ... ? Isn't the point of the dropout to generate some randomness?
I suppose we would only use that during training, as regularization, but during execution as a "currently frozen" policy it should be OK to remove the Dropout, right?
Just want to make sure before this ends up hardcoded here (instead of with a "is_testing" flag). 
I think it's OK as it is, just wanted to suggest making sure of it.

deep_quoridor/src/quoridor.py
jonbinney: I chose copy because it's what the python standard library seems to use for similar cases. For example if you have a list `foo` with some some stuff in it, `foo.copy()` returns a new list with the same elements in the same order. I'm fine with clone too - how strongly do you feel about it?

adamantivm: Very weakly. keep the copy.

PR #250, opened 2025-07-08
Handle temperature==0

alejandromarcu:
I noticed that when playing it would often pick a sub-optimal choice (e.g. probs were 0.4, 0.3 and 0.3 and it would pick 0.3).  That was because it's not correclty handling the temperature.  When playing, temperature is 0 which means choose the best option (randomly between them if there's more than 1 with the max value)

### DIFF
--- a/deep_quoridor/src/agents/alphazero/alphazero.py
+++ b/deep_quoridor/src/agents/alphazero/alphazero.py
@@ -231,7 +231,13 @@ class AlphaZeroAgent(TrainableAgent):
             raise RuntimeError("No nodes visited during MCTS")
 
         visit_probs = visit_counts / visit_counts_sum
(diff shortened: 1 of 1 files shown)


adamantivm (approved):
LGTM

jonbinney:
Is there any easy way to set `temperature=_some_number_` for training, but do the intermittent mini-tournaments with `temperature=0`?

alejandromarcu:
> Is there any easy way to set `temperature=_some_number_` for training, but do the intermittent mini-tournaments with `temperature=0`?

Good catch! for cases like this we introduced training_only_params. I think if we put temperature in that set, it will be removed when playing the tournament in the training and the default of 0 will be used.  I'll try it out.

PR #252, opened 2025-07-10
AZ: remove training_mode in mini tournaments

alejandromarcu:
By adding `training_mode ` in `training_only_params`, this parameter is removed when re-instantiating the agent to play the mini tournament for metrics.  This way, the temperature is set to 0.

Additionally, I removed the dumb_score computation, since we would need to set the temperature to 0 (and maybe other changes), or re-instantiate the agent to make sure we don't interfere with training.  Not sure if it's very useful so just removing it for now.


jonbinney (approved):
LGTM!

PR #253, opened 2025-07-11
Move softmax out of network

adamantivm:


jonbinney:
Side note: since a perfect NN will have a policy loss of at least the entropy of the target distribution, the` (cross_entropy(logit, target) - entropy(target)` might be a good metric to track for the training.

adamantivm:
Very interesting analysis @jonbinney , thank you. I was also under the wrong impression that the loss should be 0 when it converges. It's much better to have this crisper understanding. It will take a while to commit this understanding to intuition, but now there is no arguiing with this.

PR #254, opened 2025-07-12
Update masking method to make it safer to calculate cross entropy loss

adamantivm:
Apparently -1e32 is better for masking than -np.inf

```
>>> import torch
>>> import torch.nn.functional as F
>>> x = torch.Tensor([0,1])
>>> y = torch.log(x)
>>> F.cross_entropy(y,x)
tensor(nan)
>>> y
tensor([-inf, 0.])
>>> y[0] = -1e32
>>> F.cross_entropy(y,x)
tensor(-0.)
```

### DIFF
--- a/deep_quoridor/src/agents/alphazero/nn_evaluator.py
+++ b/deep_quoridor/src/agents/alphazero/nn_evaluator.py
@@ -9,6 +9,8 @@ from quoridor import ActionEncoder, Board, Player, Quoridor
 from agents.alphazero.mlp_network import MLPNetwork
 from agents.core.rotation import create_rotation_mapping
 
+INVALID_ACTION_VALUE = -1e32
+
 
 class NNEvaluator:
     def __init__(
@@ -47,7 +49,8 @@ class NNEvaluator:
         with torch.no_grad():
             input_array = self.game_to_input_array(game)
             policy_logits, value = self.network(torch.from_numpy(input_array).float().to(self.device))
-            policy_logits[invalid_mask] = -1e32
+            assert torch.isfinite(policy_logits).all(), "Policy logits contains non-finite values"
+            policy_logits[invalid_mask] = INVALID_ACTION_VALUE
             policy_masked = F.softmax(policy_logits, dim=-1).cpu().numpy()
             value = value.item()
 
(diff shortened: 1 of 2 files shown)


adamantivm:
ready for review @jonbinney

deep_quoridor/src/agents/alphazero/nn_evaluator.py
```diff
@@ -36,20 +38,20 @@ def evaluate(self, game: Quoridor):
...
-            unmasked_policy_logits = unmasked_policy_logits.cpu().numpy()
-
-            # Set logits of invalid actions to negative infinity to ensure they are never chosen
-            # TODO: Mask entirely in Torch to avoid going to Numpy and back
-            masked_logits = np.full_like(unmasked_policy_logits, -np.inf)
-            masked_logits[valid_action_indices] = unmasked_policy_logits[valid_action_indices]
-
-            policy_masked = F.softmax(torch.from_numpy(masked_logits), dim=-1).cpu().numpy()
+            policy_logits, value = self.network(torch.from_numpy(input_array).float().to(self.device))
+            assert torch.isfinite(policy_logits).all(), "Policy logits contains non-finite values"
```
jonbinney, line 52: The lines after the assert will never do anything, right?

jonbinney: (if we decide that getting -inf out of the network is expected behavior, then it will be nice to have those two lines)

deep_quoridor/src/agents/alphazero/nn_evaluator.py
```diff
@@ -36,20 +38,20 @@ def evaluate(self, game: Quoridor):
 
         self.network.eval()  # Disables dropout
 
+        # Prepare a tensor with True only on INVALID action positions so that we can set those values to
+        # a large negative number below, to discourage choosing them.
```
jonbinney, line 42: I don't think that this comment quite describes what the following lines do. They're creating a tensor with the valid indices as integers, eg. instead of `[True, False, True, True]` it would contain `[0, 2, 3]` if i understand correctly. This could explain why you had trouble getting ignore_indices to do what you expected.

jonbinney (approved):
A couple questions, but looks good to merge once those are answered.

PR #255, opened 2025-07-13
Simple tests for alphazero NN training

jonbinney:
Been working on these tests for a while, but with the other changes made this past week, they pass now! For both tests, the game state used is always the initial game state. The tests are:

1. Fill the replay buffer with many instances of `(<game at start state>, <policy that chooses Move((0, 0)) 100% of the time>, value=1.0)`
2. 1. Fill the replay buffer with many instances of  `(<game at start state>, <policy that chooses Move((0, 0)) 20% of the time and Move((0, 2)) 80% of the time>, value=1.0)`

In both cases, it trains the evaluator network with the given replay buffer and makes sure that calling `Evaluator.evaluate(<game at start state>)` produces a policy that looks like the one we fed in.

Interestingly, it converges pretty quickly to the solution in test (1) but takes much longer for test (2). It will be interesting to see if this changes when we introduce other NN structures.

### DIFF
--- a/deep_quoridor/test/agents/alphazero_test.py
+++ b/deep_quoridor/test/agents/alphazero_test.py
@@ -1,7 +1,9 @@
 import numpy as np
+import torch
 from agents.alphazero.nn_evaluator import NNEvaluator
 from agents.core import rotation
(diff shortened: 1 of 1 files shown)


adamantivm:
Nice work, I like it. LGTM

adamantivm (approved):

PR #257, opened 2025-07-17
Using single pass for UCB computation

muralx:
It goes from 
Total: 228.131 seconds

  1395387   39.471    0.000   39.471    0.000 {built-in method torch._C._nn.linear}
  1348140   20.816    0.000   25.723    0.000 mcts.py:72(get_child_ucbs)
   996705   11.841    0.000   11.841    0.000 {built-in method torch.relu}
to

Total: 193.437 seconds
  1327599   35.419    0.000   35.419    0.000 {built-in method torch._C._nn.linear}
  6114317   11.997    0.000   11.997    0.000 mcts.py:69(get_child_ucb)
   948285   10.845    0.000   10.845    0.000 {built-in method torch.relu}

Note that it does not generate the same UCBs, it seems there are precision aspects making numbers differ in the last decimal digits, this causes the performance numbers to not be fully comparable. If you look at the ratio with the other two lines, it seems there is a big improvement.


deep_quoridor/src/quoridor.py
```diff
@@ -355,6 +355,7 @@ def rotate_board(self):
         self.board.rotate_board()
         self._goal_rows = self._goal_rows[::-1]
         self._rotated = not self._rotated
+        return self
```
jonbinney, line 358: did you mean to add this back in?

PR #261, opened 2025-07-27
Early version of multiprocess training script

jonbinney:
There's a lot more to do on this, but I think getting this reviewed and merged at this point will be good - both to get feedback and avoid a huge PR later.

#### Summary
New "train_alphazero.py" script that trains alphazero via self play
using the main process for NN training and coordination, and worker
subprocesses to run multipled games in parallel.

#### Example command
This will run 40 self-play games, then train the NN, then run 40 more games, then train again. It will use 2 worker subprocesses. The `-p` is for setting the subargs parameters of the alphazero agent (but you omit the agent name since it is always alphazero).

```python -O deep_quoridor/src/train_alphazero.py --games-per-epoch 40 --epochs 2 --num-workers=2 -p mcts_n=1000```

On my laptop (which runs the NN on CPU) 2 threads is more than enough to overwhelm all the cores, since the torch NN inference runs in many system threads. On my desktop, which has an nvidia card, I can use up to 20 workers and stil have spare cores.

#### Things that still need to be done
- saving models to disk at checkpoints
- wandb integration
- better stats on what is taking the most computation time, and some optional GPU profiling
- adding noise for the policy used by the mcts search in early moves of each game, as the Alphago Zero paper does

### DIFF
--- a/deep_quoridor/src/renderers/pygame.py
+++ b/deep_quoridor/src/renderers/pygame.py
@@ -1,4 +1,3 @@
-import re
 import time
 import warnings
 
@@ -423,7 +422,6 @@ class PygameQuoridor:
             # if the user wants a verticar or horizontal wall
             return
 
-        type_action = 0
         if v_wall:
             action = WallAction((row, col), WallOrientation.VERTICAL)
         elif h_wall:
--- a/ruff.toml
+++ b/ruff.toml
@@ -50,6 +50,12 @@ unfixable = []
 # Allow unused variables when underscore-prefixed.
 dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
 
+[lint.isort]
+# Ruff sometimes mistakenly thinks the wandb python module is "first-party"
+# and puts it in that section  of the imports if the wandb cache directory
+# exists. See https://github.com/astral-sh/ruff/issues/10519
+known-third-party = ["wandb"]
+
 [format]
 # Like Black, use double quotes for strings.
 quote-style = "double"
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: deep_quoridor/src/renderers/pygame.py [hunk 1]
```diff
@@ -1,4 +1,3 @@
-import re
```
jonbinney, line 1: I Removed the lines here just because the linter was complaining about them being unused.

ruff.toml
```diff
@@ -50,6 +50,12 @@ unfixable = []
 # Allow unused variables when underscore-prefixed.
 dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
 
+[lint.isort]
+# Ruff sometimes mistakenly thinks the wandb python module is "first-party"
+# and puts it in that section  of the imports if the wandb cache directory
+# exists. See https://github.com/astral-sh/ruff/issues/10519
+known-third-party = ["wandb"]
```
jonbinney, line 57: This has happened a bunch - we've been bouncing the wandb import up and down the import list because if the "wandb" cache directory and thinks that means it's a "first-party" package instead of "third-party"

alejandromarcu (approved):
I'm a bit worried that we'll end up with a lot of duplicate code, since we need to re-do a lot of the core functionality, but we can try to see where it goes

jonbinney:
@alejandromarcu I agree - it's a fair bit of re-implementation, particularly of the arena class. I'm hoping that once we understand more about how the multiprocess/multithreaded/batching needs to be done, we'll be able to unify this and train.py (or at least share more code).

PR #262, opened 2025-07-28
Introduce mcts_k

alejandromarcu:
With this PR, if you specify mcts_k instead of mcts_n, the number of searches in MCTS will be mcts_k * number_of_actions.  This way, we ensure we have enough searches to explore the space, but we don't waste time if there are just a few actions (usually when there are no walls)


alejandromarcu:
Closing because I rather write a new PR than figure out conflicts in git :'(
New PR: https://github.com/jonbinney/deep_rabbit_hole/pull/264

PR #265, opened 2025-08-02
Adding util to generate a gexf file with the MCTS graph

muralx:
- Also adds wins and loss when when MCTS find terminal state nodes.


jonbinney (approved):
Looks good - I'm excited to try this!

PR #266, opened 2025-08-08
Show the state value in pygame

alejandromarcu:
I thought it would be useful to display the state value to better understand what's going on, i.e. how likely MCTS thinks that the player will win.  We can see that at the beginning the values are close to 0, because we end up evaluating mid-game states and the network is untrained.  Later in the game, MCTS is able to reach winning states and we can see that the values are higher and seem reasonable (e.g. getting closer to 1 when it has a good chance of winning)

```
/Users/amarcu/code/deep_rabbit_hole/.venv/bin/python /Users/amarcu/code/deep_rabbit_hole/deep_quoridor/src/train.py -N 5 -W 3 -e 1000 -i 43 -p alphazero:training_mode=true,mcts_n=1000,learning_rate=0.0001,batch_size=64,replay_buffer_size=100000,mcts_ucb_c=2,optimizer_iterations=50 -r arenaresults2 computationtimes -f 20 -r pygame
```

https://github.com/user-attachments/assets/fa5b235b-a716-4e58-b162-baa6d990ce2f

### DIFF
--- a/deep_quoridor/src/agents/alphazero/alphazero.py
+++ b/deep_quoridor/src/agents/alphazero/alphazero.py
@@ -7,14 +7,14 @@ from typing import Optional, Tuple
 
 import numpy as np
 import torch
-from agents.alphazero.mcts import MCTS
-from agents.alphazero.nn_evaluator import NNEvaluator
-from agents.core import TrainableAgent
+import wandb
 from quoridor import ActionEncoder, MoveAction, construct_game_from_observation
 from utils import my_device
 from utils.subargs import SubargsBase
 
-import wandb
+from agents.alphazero.mcts import MCTS
+from agents.alphazero.nn_evaluator import NNEvaluator
+from agents.core import TrainableAgent
 
 
 @dataclass
--- a/deep_quoridor/src/agents/alphazero/alphazero.py
+++ b/deep_quoridor/src/agents/alphazero/alphazero.py
@@ -7,14 +7,14 @@ from typing import Optional, Tuple
 
 import numpy as np
 import torch
-import wandb
-from quoridor import ActionEncoder, construct_game_from_observation
-from utils import my_device
-from utils.subargs import SubargsBase
-
 from agents.alphazero.mcts import MCTS
 from agents.alphazero.nn_evaluator import NNEvaluator
 from agents.core import TrainableAgent
+from quoridor import ActionEncoder, MoveAction, construct_game_from_observation
+from utils import my_device
+from utils.subargs import SubargsBase
+
+import wandb
 
 
 @dataclass
@@ -233,6 +233,8 @@ class AlphaZeroAgent(TrainableAgent):
         self,
         visit_probs,
         root_children,
+        root_value,
+        root_action,
     ):
         if not self.action_log.is_enabled():
             return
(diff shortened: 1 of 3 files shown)


deep_quoridor/src/agents/alphazero/alphazero.py
```diff
@@ -7,14 +7,14 @@
...
-from utils.subargs import SubargsBase
-
 from agents.alphazero.mcts import MCTS
 from agents.alphazero.nn_evaluator import NNEvaluator
 from agents.core import TrainableAgent
+from quoridor import ActionEncoder, MoveAction, construct_game_from_observation
+from utils import my_device
+from utils.subargs import SubargsBase
+
+import wandb
```
jonbinney, line 17: hmmm... looks like my wandb import order fix for our ruff config didn't quite work in this case. If you have time, could you check that you're using ruff for import ordering (not just formatting, which vscode treats as separate). [This](https://github.com/jonbinney/deep_rabbit_hole/blob/2db4eb18e89fe812b6eb2d3c40e048e41a2143cf/ruff.toml#L57) line should cause wandb to show up in the third-party section, up above the first-party imports.

deep_quoridor/src/agents/alphazero/nn_evaluator.py
```diff
@@ -65,9 +65,9 @@ def evaluate(self, game: Quoridor, extra_games_to_evaluate: list[Quoridor] = [])
             # If the game was originally rotated, rotate the resulting back to player 2's perspective
             if g.get_current_player() == Player.TWO:
                 policy_masked[i] = policy_masked[i][self.action_mapping_rotated_to_original]
-            self.cache[all_hashes[i]] = (values[i], policy_masked[i])
+            self.cache[all_hashes[i]] = (values[i][0], policy_masked[i])
```
jonbinney, line 68: Was this a bug before? Why wasn't it breaking things?

deep_quoridor/src/agents/alphazero/mcts.py
```diff
@@ -131,5 +131,7 @@ def search(self, initial_game: Quoridor):
                 self.new_nodes.extend(node.children)
 
             node.backpropagate(value)
-
-        return root.children
+        # Negate the value because the value is actually the value that the opponent
+        # got for getting to that state.
+        root_value = -(root.value_sum / root.visit_count)
```
jonbinney, line 136: Shouldn't this only be negated if the opponent wins? I'm probably missing something.

jonbinney (approved):
Looks good - left a few comments just for my own understanding of the code.

deep_quoridor/src/agents/alphazero/alphazero.py
alejandromarcu: I uninstalled another extension that I had installed (isort) that is just to sort imports, and now it seems to work well

deep_quoridor/src/agents/alphazero/nn_evaluator.py
alejandromarcu: I was debugging for this PR and noticed that instead of getting a float for the value I was getting a ndarray with 1 float.  Nothing changed when I did this change, so I don't think it was breaking anything, probably somewhere else it was casted as a float or something.  In any case, this seems the right way to do it.

deep_quoridor/src/agents/alphazero/mcts.py
alejandromarcu: I spent a lot of time debugging this because initially I thought it was a bug. But actually, let's say that it's turn for player 1, so we have the root node, and every children will have a value saying how good this move is for player 1.   When you look at those children from the point of view of player 2 (because once you get into that children it will be player 2 turn), you need to negate it because when we computed it, it was from the pov of player 1.

jonbinney: Ah, I see!

PR #267, opened 2025-08-08
Client/Server Evaluator for multiprocess training

jonbinney:
The biggest change in this PR is that instead of each worker process having its own evaluator with its own GPU memory allocation, they now use an "EvaluatorClient" which does the CPU work of rotating the game and creating the action mask, and then sends them to an "EvaluatorServer" in another subprocess to do the GPU work. This allows adding more worker processes to use more of the CPU without running out of GPU memory.

Other changes:
- **Adds caching** (like in @alejandromarcu 's improvements to NNEvaluator). Caches both on the client side and on the server side. The server side cache has more game states in it since it includes results from every worker, but takes more time to get to. The client cache gets hit far more often. Actually I'm not sure its worth the complexity of having the server side cache, but I left it in for now.
- **Supports evaluating multiple game states at once**, so that mcts can pre-evaluate nodes
- **Adds statistics for the evaluator client and server** so that we can see how much time they're taking and how often their caches are hit.
- **Makes things deterministic** so that if you run it with the same parameters multiple times you should get the same result. I haven't tested this thoroughly though. The changes include setting random seeds in each subprocess, and ordering the results from the worker processes by their worker id so that their running time doesn't affect the order of things in the replay buffer.

You can try this with the following command:
```
time python -O deep_quoridor/src/train_alphazero.py -g 40 -e 2 --num-workers=10 -p mcts_n=100
```

The equivalent command using train.py would be:
```
python -O deep_quoridor/src/train.py -N 5 -W 3 -e 80 -p alphazero:training_mode=true,mcts_n=1000,train_every=40 -r arenaresults2 computationtimes
```

Before this PR the multiprocess training took about 24 minutes on my desktop. train.py takes about 17 minutes on my PC with the equivalent command. After these new changes multiprocess training takes about 2.5 minutes. I'm curious to hear how it performs on other machines.

### DIFF
--- a/deep_quoridor/src/agents/alphazero/multiprocess_evaluator.py
+++ b/deep_quoridor/src/agents/alphazero/multiprocess_evaluator.py
@@ -6,6 +6,8 @@ import queue
 import threading
 import time
 from collections import deque
+from dataclasses import dataclass
+from typing import Optional
 
 import numpy as np
 import torch
@@ -17,6 +19,55 @@ from agents.alphazero.nn_evaluator import NNEvaluator
 from agents.core.rotation import create_rotation_mapping
 
 
+@dataclass
+class EvaluationInfo:
+    """
+    Information about one call to the evaluator.
+    """
+
+    evaluation_start_time: float
+    evaluation_end_time: float
+    num_games_evaluated: int
+    used_cache: bool
+
+
+@dataclass
+class EvaluatorStatistics:
+    """
+    Statistics about all evaluations since this evaluator was created.
+    """
+
+    duty_cycle: float
+    average_evaluation_time: float
+    num_evaluations: int
+    num_cache_hits: int
+
+
+def compute_statistics(evaluation_log: list[EvaluationInfo]) -> Optional[EvaluatorStatistics]:
+    evaluation_time_sum = 0
+    num_evaluations = len(evaluation_log)
+    num_cache_hits = 0
+    first_evaluation_time = None
+    last_evaluation_time = None
+    for eval_info in evaluation_log:
+        if first_evaluation_time is None:
+            first_evaluation_time = eval_info.evaluation_start_time
+        last_evaluation_time = eval_info.evaluation_end_time
+        evaluation_time_sum += eval_info.evaluation_end_time - eval_info.evaluation_start_time
+        if eval_info.used_cache:
+            num_cache_hits += 1
+
+    if num_evaluations > 0:
+        if first_evaluation_time is None or last_evaluation_time is None:
+            duty_cycle = None
+        else:
+            duty_cycle = evaluation_time_sum / (last_evaluation_time - first_evaluation_time)
+
+        return EvaluatorStatistics(duty_cycle, evaluation_time_sum / num_evaluations, num_evaluations, num_cache_hits)
+    else:
+        return None
+
+
 class EvaluatorClient:
     def __init__(
         self,
@@ -24,6 +75,7 @@ class EvaluatorClient:
         client_id: int,
         request_queue: mp.Queue,
         result_queue: mp.Queue,
+        max_log_len=int(1e8),
     ):
         self._client_id = client_id
         self._request_queue = request_queue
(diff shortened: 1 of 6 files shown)


alejandromarcu:
In my Mac with this change I'm getting:
```
python3 -O deep_quoridor/src/train_alphazero.py -g 40 -e 2 --num-workers=10 -  116.96s user 14.07s system 255% cpu 51.226 total
```

And using the regular training:
```
python3 -O deep_quoridor/src/train.py -N 5 -W 3 -e 80 -p  -r arenaresults2   691.46s user 22.35s system 92% cpu 12:51.36 total
```

So the wall clock went down from almost 13 minutes to less than 1!! That's great!
Do you think we'll be able to somehow use the existing plug ins with this architecture?

deep_quoridor/src/agents/alphazero/multiprocess_evaluator.py
```diff
@@ -0,0 +1,223 @@
...
+            value, policy = self._cache.get(cache_key)
+            values_array = np.expand_dims(value, 0)
+            policies_array = np.expand_dims(policy, 0)
+            used_cache = True
+        else:
+            inputs_tensor = torch.from_numpy(inputs_array).to(my_device())
+            action_masks_tensor = torch.from_numpy(action_masks_array).to(my_device())
+
+            # Do the actual evalution
+            values_tensor, policies_tensor = self._evaluator.evaluate_tensors(inputs_tensor, action_masks_tensor)
```
alejandromarcu, line 198: I realized something: we are evaluating all the "extra games", but some of them may be in the cache already, because they could have been used in a previous game or even in this same game.  Not sure if this is worth trying, but we could remove from the input everything that is already in the cache (can be done separately).

deep_quoridor/src/agents/alphazero/multiprocess_evaluator.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def run(self):
+        while not self._shutdown:
+            try:
+                inputs_array, action_masks_array, client_id = self._input_queue.get(timeout=0.1)
+            except queue.Empty:
+                continue
+
+            values_array, policies_array = self._evaluate_arrays(inputs_array, action_masks_array)
+            self._send_result_to_client(client_id, values_array, policies_array)
```
alejandromarcu, line 175: One thing that we can consider in the future is to batch together requests from multiple workers.
I.e. we would loop getting an element from the queue until we hit a timeout and accumulate the inputs.  Then we would evaluate everything together, and use the cache to reply to each client what they expect

deep_quoridor/src/train_alphazero.py
```diff
@@ -64,117 +81,164 @@ def self_play_worker(
...
+        for r in results:
+            print(r.evaluator_statistics)
+
+
+def setup(
+    board_size: int, max_walls: int, max_game_length: int, num_workers: int, random_seed: int
+) -> tuple[EvaluatorServer, list[Worker]]:
+    set_deterministic(random_seed)
+
+    set_deterministic(args.seed)
```
alejandromarcu, line 163: You're setting it twice, I think it's the same value, but probably this one should be removed

alejandromarcu:
I tried running with just 1 worker and I got 1:40 of clock time, so half the speed than running with 10 workers, but much faster than the 13 minutes of the previous version.   So, it's not only multiple workers in parallel that increased the performance, there has to be something else that made it so much faster. Do you know what it is?

alejandromarcu:
I ran the original train script with --profile and I got this:

```
         1136356005 function calls (1061665614 primitive calls) in 993.069 seconds

   Ordered by: internal time

141123627  311.620    0.000  311.620    0.000 mcts.py:67(get_child_ucb)
 19211655   65.040    0.000  101.628    0.000 quoridor.py:164(create_new)
59732360/6389000   63.482    0.000   63.482    0.000 mcts.py:78(backpropagate)
  6027444   62.774    0.000   83.998    0.000 mcts.py:50(expand)
53345508/53345496   59.573    0.000  371.200    0.000 {built-in method builtins.max}
   143662   49.117    0.000   49.117    0.000 {method 'cpu' of 'torch._C.TensorBase' objects}
 53343360   48.654    0.000  419.845    0.000 mcts.py:61(select)
   144290   36.853    0.000   36.853    0.000 {method 'to' of 'torch._C.TensorBase' objects}
   504217   19.212    0.000   19.212    0.000 {built-in method torch._C._nn.linear}
 18712477   18.102    0.000   18.102    0.000 mcts.py:9(__init__)
  6389000   15.276    0.000  447.051    0.000 mcts.py:97(select)
 38954712   15.086    0.000   15.086    0.000 {built-in method numpy.array}
     6389   11.935    0.002  987.525    0.155 mcts.py:107(search)
```

It's frustrating that something as simple as get_child_ucb is taking 1/3 of the wall time, and also create_new about 1/10th.
I tried to profile your new script (using `-m cProfile`) but I got an error, not sure if it doesn't support multithreading or what's going on

deep_quoridor/src/agents/alphazero/multiprocess_evaluator.py
```diff
@@ -0,0 +1,223 @@
...
+
+        for row_i in range(len(values_array)):
+            # Sanity checks
+            assert np.all(policies_array[row_i] >= 0), "Policy contains negative probabilities"
+            assert np.abs(np.sum(policies_array[row_i]) - 1) < 1e-6, "Policy does not sum to 1"
+            assert np.isfinite(values_array[row_i]), "Policy or value is non-finite"
+
+            # Update our local cache
+            cache_key = all_games_inputs_array[row_i].tobytes()
+            self._cache[cache_key] = (values_array[row_i], policies_array[row_i], game.copy())
```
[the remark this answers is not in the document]
jonbinney: I've removed this in my local branch, and I'll include that in my next PR. I included the game because I was debugging some odd caching issues and wanted to make sure I knew which game state each cache entry corresponded to (creating a utility function to convert from input_array to game_state would be a better way to do this)

deep_quoridor/src/agents/alphazero/multiprocess_evaluator.py
jonbinney: Yes! I had this implemented earlier but it was hard to set the timeout so that it didn't delay the worker processes too much, and so I removed it. I think it's worth adding back in. @muralx suggested trying it without the timeout - you grab as many items are in the queue, stack them together, and evaluate them at once. This should help if a few requests get backed up in the queue while we're evaluating the previous request.

jonbinney:
I have a few things things to do before this will be ready to merge:
- make sure that the single-worker speedup over train.py isn't the result of some bug
- a few small fixes based on review comments
- fix merge conflicts
I'm hoping to have these done before our meeting today.... we'll see.

deep_quoridor/src/agents/alphazero/multiprocess_evaluator.py
```diff
@@ -0,0 +1,223 @@
...
+    def evaluate(self, game: Quoridor, extra_games_to_evaluate: list[Quoridor] = []):
+        evaluation_start_time = time.time()
+
+        game, _ = self.rotate_if_needed_to_point_downwards(game)
+        input_array = self.game_to_input_array(game)
+        cache_key = input_array.tobytes()
+        if cache_key in self._cache:
+            cached_value, cached_policy, _ = self._cache[cache_key]
+            if game.is_rotated():
+                cached_policy = cached_policy[self._action_mapping_rotated_to_original]
```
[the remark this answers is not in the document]
jonbinney: Done

deep_quoridor/src/agents/alphazero/multiprocess_evaluator.py
```diff
@@ -0,0 +1,223 @@
...
+        input_queue: mp.Queue,
+        output_queues: list[mp.Queue],
+        max_log_len=int(1e8),
+    ):
+        super().__init__()
+        self._evaluator = evaluator
+        self._cache = {}
+        self._input_queue = input_queue
+        self._output_queues = output_queues
+        self._max_l = max_log_len
```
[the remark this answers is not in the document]
jonbinney: Removed

PR #268, opened 2025-08-09
Compute metrics against each opponent

jonbinney:
In metrics, and in the wandb training plugin, this PR adds a wandb metric for the win rate as p1 against each opponent agent, and against p2 against each other agent. This should be helpful since the win rate against random often gets to 100% quickly, whereas as P2 against simple is much harder. You can't see it in this screenshot, but if you hover over the graph it shows you which color is which opponent agent.

image


muralx (approved):
Thanks!

PR #270, opened 2025-08-14
VideoRenderer: include the ActionLog

alejandromarcu:
Show the action log, which allows to track internal data.  E.g. in AlphaZero, we show in the pawn position the estimated value for the pawn at that state, and the top 5 possible next actions.
Below we can see how it works in a game that is truncated.  I'm wondering if we may be mixing things up in the training with the rotation and have contradictory things on where's the target row, and that's why it goes back and forth. 

https://github.com/user-attachments/assets/b2c2fc5f-cd3d-4459-8e41-e33b04adb656


jonbinney (approved):
LGTM!

PR #271, opened 2025-08-15
Interactive Evaluator Visualizer

alejandromarcu:
This new utility allows us to set up a board in any position and visualize the output from the NN.
You can select the black or white pawn by clicking on it, and then when you click on a cell you will move the pawn there, or if you click on a wall you will toggle it.
Every time you make a change, it calls the NN to re-evaluate the policy and shows it.
You just need to pass a trained network ot the script, or if you don't pass anything it will use an untrained network.
The network that I'm showing in this demo was trained with only 80 episodes.  My observations from it is that it doesn't have too strong of an opinion (e.g. #1 is usually around 0.06), and that even when you move things around, the output of the network is not shifting a lot.  No idea why this is happening, but at least it's good to know. I think my next step would be to visualize what we're using to train the NN.

At leat one positive thing is that the 2nd option for the white is to move forward in the first move, and for the black moving forward is the 1st option.  

```
 python deep_quoridor/src/interactive_evaluator_visualizer.py models/alphazero_B5W3_mv1.0_episode_80.pt
```

https://github.com/user-attachments/assets/9aa5d3e7-8792-455e-be66-081c5f7f72b1

### DIFF
--- a/deep_quoridor/src/agents/alphazero/nn_evaluator.py
+++ b/deep_quoridor/src/agents/alphazero/nn_evaluator.py
@@ -204,29 +204,3 @@ class NNEvaluator:
             self.optimizer.step()
 
         return {"total_loss": total_loss.item(), "policy_loss": policy_loss.item(), "value_loss": value_loss.item()}
-
-    def action_log_for_game(self, game: Quoridor) -> ActionLog:
-        """Generate ActionLog showing neural network evaluation for the current game state."""
-        al = ActionLog()
-        al.set_enabled(True)
-
-        # Get neural network evaluation
-        value, policy = self.evaluate(game)
-        # Get valid actions and their scores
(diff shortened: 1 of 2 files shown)


deep_quoridor/src/agents/alphazero/nn_evaluator.py
```diff
@@ -203,3 +204,29 @@ def train_iteration(self, replay_buffer):
             self.optimizer.step()
 
         return {"total_loss": total_loss.item(), "policy_loss": policy_loss.item(), "value_loss": value_loss.item()}
+
+    def action_log_for_game(self, game: Quoridor) -> ActionLog:
```
muralx, line 208: can we keep this outside the evaluator so any evaluator could use the functionality?

PR #273, opened 2025-08-16
Save Replay Buffers to files and allow to use it for bootstrapping

alejandromarcu:
Now if you use`save_replay_buffer=true` it will save the replay buffer to a file before each training.
Then, you can "bootstrap" it by starting AlphaZero by training it right away with the replay buffer.  This allow us to quickly try different training parameters without having to run all the initial games, which won't change if you don't change the parameters.  
Another part of this, that I'll do in a separate PR, is to have an utility that will load a replay buffer and visualize it (e.g. show a state and the policy as a board).

Here's an example usage.
First I run a training session with `save_replay_buffer=true`.  I just want to play the first N (=100 in this case) games and then save it, so I use -e 100 and train_every=100:

```
python -O /Users/amarcu/code/deep_rabbit_hole/deep_quoridor/src/train.py -N 5 -W 3 -e 100 -i 43 -p alphazero:training_mode=true,mcts_n=100,learning_rate=0.0001,batch_size=64,replay_buffer_size=100000,mcts_ucb_c=2,optimizer_iterations=100,save_replay_buffer=true,train_every=100 -r arenaresults2 computationtimes -f 20
```

This generated the file `replay_buffers/replay_buffer_episode_100.pkl`.

Now I can try using the generated file, so that it will run the training right away.   With this I could try for example different values for the learning_rate, batch_size, optimizer_iterations to see how different trainings parameters affect the result.

```
python -O /Users/amarcu/code/deep_rabbit_hole/deep_quoridor/src/train.py -N 5 -W 3 -e 1 -i 43 -p alphazero:training_mode=true,mcts_n=1000,learning_rate=0.0001,batch_size=64,replay_buffer_size=100000,mcts_ucb_c=2,optimizer_iterations=100,bootstrap_replay_buffer=replay_buffers/replay_buffer_episode_100.pkl -r video arenaresults2 computationtimes -f 20
Starting DQN training...
Board size: 5x5
Max walls: 3
Training for 1 episodes
Using step rewards: False
Device: mps
Initializing random seed 43
Loaded 5196 samples from replay_buffers/replay_buffer_episode_100.pkl
Running bootstrap training iteration...
Training the network (buffer size: 5196, batch size: 64)...done in 0.92s
alphazero P2 Episode     1/1 [ ], Steps: 150, Time: 13622ms,  Avg Reward:   0.00, Avg Loss:  3.9474, Epsilon: 0.0000 opponent: alphazero
alphazero P2 Episode     1/1 [ ], Steps: 150, Time: 13622ms,  Avg Reward:   0.00, Avg Loss:  3.9474, Epsilon: 0.0000 opponent: alphazero
AlphaZero model saved to /Users/amarcu/code/deep_rabbit_hole/models/alphazero_B5W3_mv1.0_final.pt
```

From here, I could for example check the generated video to see how it played the game, or I could use the interactive visualizer to see how the NN responds to different states, e.g. by doing:

```
python deep_quoridor/src/interactive_evaluator_visualizer.py models/alphazero_B5W3_mv1.0_final.pt
```

### DIFF
--- a/deep_quoridor/src/agents/alphazero/alphazero.py
+++ b/deep_quoridor/src/agents/alphazero/alphazero.py
@@ -79,6 +79,9 @@ class AlphaZeroParams(SubargsBase):
     # If True, save replay buffer contents before training
     save_replay_buffer: bool = False
 
+    # If specified, load replay buffer from this file path to bootstrap training
+    bootstrap_replay_buffer: Optional[str] = None
+
     @classmethod
     def training_only_params(cls) -> set[str]:
         """
@@ -93,6 +96,7 @@ class AlphaZeroParams(SubargsBase):
             "batch_size",
             "replay_buffer_size",
             "save_replay_buffer",
+            "bootstrap_replay_buffer",
         }
 
 
@@ -151,6 +155,21 @@ class AlphaZeroAgent(TrainableAgent):
         from metrics import Metrics
 
         self.metrics = Metrics(board_size, max_walls)
+
+        # Load bootstrap replay buffer if specified (after all initialization)
+        if params.bootstrap_replay_buffer is not None:
+            if os.path.exists(params.bootstrap_replay_buffer):
+                with open(params.bootstrap_replay_buffer, 'rb') as f:
+                    bootstrap_data = pickle.load(f)
+                self.replay_buffer.extend(bootstrap_data)
+                print(f"Loaded {len(bootstrap_data)} samples from {params.bootstrap_replay_buffer}")
+
+                # Run initial training if we have enough data
+                if len(self.replay_buffer) >= self.params.batch_size:
+                    print("Running bootstrap training iteration...")
+                    self.train_iteration()
+            else:
+                print(f"Warning: Bootstrap replay buffer file not found: {params.bootstrap_replay_buffer}")
 
     def version(self):
         return "1.0"
--- a/deep_quoridor/src/agents/alphazero/alphazero.py
+++ b/deep_quoridor/src/agents/alphazero/alphazero.py
@@ -1,4 +1,5 @@
 import os
+import pickle
 import time
 from collections import deque
 from dataclasses import dataclass
@@ -75,6 +76,9 @@ class AlphaZeroParams(SubargsBase):
     # If True, the agent will penalize visited states in MCTS to avoid cycling
     penalized_visited_states: bool = False
 
+    # If True, save replay buffer contents before training
+    save_replay_buffer: bool = False
+
     @classmethod
     def training_only_params(cls) -> set[str]:
         """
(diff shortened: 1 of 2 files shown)


deep_quoridor/src/agents/alphazero/alphazero.py
```diff
@@ -234,6 +251,10 @@ def train_iteration(self):
         if len(self.replay_buffer) < self.params.batch_size:
             return
 
+        # Save replay buffer if requested
+        if self.params.save_replay_buffer:
+            self.save_replay_buffer_to_file(self.episode_count)
```
muralx, line 256: we should only save the buffer once (first time before training the network), right? Unless you see a point on keeping all game states and policies, etc. 

Also if you loaded the buffer, you would be saving it again here, right?

deep_quoridor/src/agents/alphazero/alphazero.py
```diff
@@ -249,6 +270,28 @@ def train_iteration(self):
 
         print(f"done in {time.time() - t0:.2f}s")
 
+    def save_replay_buffer_to_file(self, episode_number: int):
+        """Save replay buffer contents to a file."""
+        replay_buffer_dir = "replay_buffers"
+        os.makedirs(replay_buffer_dir, exist_ok=True)
+        filename = f"replay_buffer_episode_{episode_number}.pkl"
```
muralx, line 277: ideally we should not overun past runs, right?

deep_quoridor/src/agents/alphazero/alphazero.py
```diff
@@ -147,6 +156,14 @@ def __init__(
 
         self.metrics = Metrics(board_size, max_walls)
 
+        # Load bootstrap replay buffer if specified (after all initialization)
+        if params.bootstrap_replay_buffer is not None:
+            if self.load_replay_buffer_from_file(params.bootstrap_replay_buffer):
```
muralx, line 161: nit: I think it would be better to make this save/load automatic. If there is a buffer that match the configuration of the current run then load it and skip the initial epoch.

Keep in mind that "in theory" NN should match the configuration of the NN that generated that initial file, so random seed and NN configuration would be relevant to be in the name of the file.

deep_quoridor/src/agents/alphazero/alphazero.py
muralx: Adding to my comment, we should always save if we did not load it and also just make load/save automatic.

deep_quoridor/src/agents/alphazero/alphazero.py
muralx, line 277: use a more specific name to avoid unexpected overwrites.

alejandromarcu: yeah, I can do what you mentioned of including the parameters in the name of the file

PR #275, opened 2025-08-19
Visualize the Replay Buffer and the model prediction

alejandromarcu:
This allows us to see what's in the replay buffer (i.e. training data) and how the model predicts the same input.

For example, I trained a model using:
```
python -O /Users/amarcu/code/deep_rabbit_hole/deep_quoridor/src/train.py -N 5 -W 3 -e 1000  -p alphazero:training_mode=true,mcts_n=1000,learning_rate=0.0001,batch_size=64,replay_buffer_size=100000,mcts_ucb_c=2,optimizer_iterations=100,train_every=100,save_replay_buffer=always -r arenaresults2 computationtimes -f 100 -w
```

This saved the model and the replay buffer files, so then I can do:

```
python deep_quoridor/src/replay_buffer_visualizer.py replay_buffers/alphazero_B5W3_ep100_i42_t100_rbs100000_n1000_ucbc200.pkl models/alphazero_B5W3_mv1.0_episode_100.pt
```
And navigate through the replay buffer entries

image

I barely wrote any code for it, Claude Code did almost everything, except the last commit.  Here are the prompts for reference:

Initial Version
```
I want to create a tool to visualize the contents of the replay buffer and optionally see how it differs from the trained network.
- The tool will take one mandatory parameter (the filename of the saved replay buffer) and one optional parameter (the filename of the trained model)
- The tool will use matplotlib and take advantage of the visualizations that we already have
- In the tool we will see one panel for the replay buffer and, if present, another panel for the trained model
- In the replay buffer panel we'll have buttons to go to the previous and next reply buffer entry, as well as to the previous and next entries with a different state
- The replay buffers will be sorted descending by the number of times the input array appears.
- The input_array in the replay buffer needs to be used to re-construct a game
- The mcts_policy in the replay buffer will be shown using ActionLog, and the value as well, shown where the pawn is.

This is a long task so think hard and ask if you need further clarifications
```

Fix wall reconstruction
```
I never see any walls when looking through different entries, but there should be walls in at least some entries.
```

Fix navigation issue
```
When I start it, it's in entry 1/1841 and it says Frequency 100.  So, I expect the first 100 entries to all have the same input_array (i.e. the same game, but they will probably have different policies and values).  However, when I press the right key, the game is different.
When I try to skip to the next state pressing shift+right, I was expecting to jumpt to the entry 101/1841, but it goes to entry 2.
```

Then when I was looking at the code I realized that it was not using the number of walls placed to compute the total number of walls, so I gave it this prompt:
```
The walls variable will have an 1 per wall, so we
   can count the number of placed walls counting 
  the numbers of 1.  Then, to know the 
  max_walls_per_players exactly, we can add to the 
  remaining walls of both players the walls that 
  are already placed, and that would give us twice 
  as much walls per players
```
But it put the `//2` just for walls_placed, not for everything, which I fix by hand, and then I fixed it so that the walls remaining after the placement match what it should be, since the NN may produce different results if this is wrong

### DIFF
--- a/deep_quoridor/src/replay_buffer_visualizer.py
+++ b/deep_quoridor/src/replay_buffer_visualizer.py
@@ -15,15 +15,14 @@ import pickle
 import sys
 from collections import Counter
 from pathlib import Path
-from typing import Optional, Tuple
+from typing import Optional
 
 import matplotlib.pyplot as plt
 import numpy as np
 import torch
 from agents import ActionLog
 from agents.alphazero.nn_evaluator import NNEvaluator
-from mpl_visualizer import visualize_board
-from quoridor import ActionEncoder, Board, MoveAction, Player, Quoridor
+from quoridor import ActionEncoder, Board, MoveAction, Player, Quoridor, WallOrientation
 from utils.misc import my_device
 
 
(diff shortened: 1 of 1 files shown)


muralx (approved):
Nice!

PR #278, opened 2025-09-08
Add dirichlet noise to policy in MCTS

jonbinney:
Choosing the `alpha` parameter for the noise is a bit tricky for quoridor; I put some thoughts into a comment in the alphazero parameters class. For now I set the default values to what I think makes the most sense. I'd also be fine with having dirichlet noise be off for now (by setting AlphazeroParams.mcts_noise_epsilon to default to zero). I think we'll want it on eventually to match what the alphazero paper does.

I didn't see a noticeable improvements in results yet, but I think it is encouraging explortation of various different moves.


alejandromarcu (approved):
LGTM

PR #283, opened 2025-09-16
Load the model if passed in AlphaZero

alejandromarcu:
We were not loading the model if passed using a filename or wandb alias.  Since the metrics during training rely on loading the model, they were just measuring a random initialized network.  Same thing if you tried to play against it by using the play script and passing for example a wandb_alias

### DIFF
--- a/deep_quoridor/src/agents/alphazero/alphazero.py
+++ b/deep_quoridor/src/agents/alphazero/alphazero.py
@@ -186,7 +186,7 @@ class AlphaZeroAgent(TrainableAgent):
         else:
             self.evaluator = evaluator
 
-        self._fetch_model_from_wand_and_update_params()
+        self._fetch_model_from_wandb_and_update_params()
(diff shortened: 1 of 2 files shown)


deep_quoridor/src/agents/alphazero/alphazero.py
```diff
@@ -240,6 +245,72 @@ def __init__(
                 self.train_iteration(is_replay_buffer_bootstrap=True)
                 self.first_replay_buffer_loaded = True
 
+    # TO DO, this was copy-pasted from AbstractTrainableAgent, need to refactor it
+    def _fetch_model_from_wand_and_update_params(self):
```
jonbinney, line 249: wand is "weights and ..........."? :-P

alejandromarcu: oh, no, just a magic wand 😉

PR #284, opened 2025-09-18
Respawn self-play workers to avoid memory leak

jonbinney:
Adds a bit of overhead but does fix the leak! `train_alphazero.py` was getting quite large, so I moved the worker starting/restarting/etc to its own class in self_play_manager.py. In this PR I also updated the ruff and pylance settings so that I don't keep missing type warnings.

A few of other minor things:
- updated the check in AlphaZeroAgent.train_iteration to make sure that the replay buffer is longer in batch size even after pulling out the validation set
- fixed the model path passed to the benchmarking
- switched to calling AlphaZeroAgent.train_iteration instead of calling NNEvaluator.train_iteration directly from train.py, because that was bypassing the code that saves the replay buffer.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,8 +3,6 @@
     "python.analysis.autoImportCompletions": true,
     "python.analysis.useLibraryCodeForTypes": true,
     "[python]": {
-        // Type checking with Pylance (choose: "off", "basic", or "strict")
-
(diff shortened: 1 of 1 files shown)


jonbinney:
This PR isn't as big and scary as it looks - most of it is moving the self-play code from train_alphazero.py to self_play_manager.py.

PR #285, opened 2025-09-19
Misc fixes

alejandromarcu:
Fixing a few small things that I run into recently.  More context in code.

### DIFF
--- a/deep_quoridor/src/agents/alphazero/alphazero.py
+++ b/deep_quoridor/src/agents/alphazero/alphazero.py
@@ -24,6 +24,9 @@ from agents.core import TrainableAgent
 class AlphaZeroParams(SubargsBase):
     training_mode: bool = False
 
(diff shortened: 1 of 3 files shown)


deep_quoridor/src/agents/alphazero/mcts.py
```diff
@@ -106,8 +106,8 @@ def backpropagate_result(self, value: float):
         """
         self.value_sum += value
         self.visit_count += 1
-        self.wins = self.wins + 1 if value == 1 else 0
-        self.losses = self.losses + 1 if value == -1 else 0
+        self.wins = self.wins + (1 if value == 1 else 0)
+        self.losses = self.losses + (1 if value == -1 else 0)
```
alejandromarcu, line 110: this would was resetting the wins and losses rather than not incrementing the value

PR #287, opened 2025-09-21
MCTS: bug fix and optimization

alejandromarcu:
When experimenting with a regular size board, the play was exteremely slow, usually more than 10 minutes per game!
I discovered that there was a bug that was slowing it down: `self.new_nodes` was not cleared, so if there were left overs from the previous search (which didn't happen that often in a smaller board because the number of children is much less) we will be spending compute time in evaluating games from the previous move, that usually won't be needed, instead of the games from this move, that will probably be needed.  Just that fix improved the speed 7x in the 9x9 board (but doesn't make a difference in the 5x5 board).

I also had an idea: instead of queuing all the children for piggy-back evaluation, just queue the top K based on the prior, since those are the first ones to be evaluated.   That way, we evaluate games that are more likely to be used.  So we get less cache hits, but we can significantly reduce the evaluation size to the point of making it worth.  This gave another almost 2x in speed in 9x9.  Notice that the gains vary greatly depending on the board size and mcts_n.

### Benchmarks

**9x9 board**

I used the following command:
```
.venv/bin/python -O deep_quoridor/src/train.py \
-N 9 -W 10 -e 1 -f 500 -i 26 --benchmark random greedy:p_random=0.5,nick=greedy-05 greedy:p_random=0.3,nick=greedy-03 greedy:p_random=0.1,nick=greedy-01 simple \
-p alphazero:training_mode=true,replay_buffer_size=100000,mcts_ucb_c=2,\
train_every=500,save_replay_buffer=first,mcts_noise_epsilon=0.1,\
mcts_n=2000,learning_rate=0.0001,batch_size=128,optimizer_iterations=200,validation_ratio=0.2
```

Running it in main took 827 seconds.
Just by clearing the nodes (`self.new_nodes = []`) the time decreased to 112s

I tried also to run it with mcts_pre_evaluate_nodes_total=1024 but the time increased to 166s.

Then, with this pr I tried with the following values of mcts_top_k_pre_evaluate, and got the following times and cache hits, out of 233k evaluations
- 4: 77s (203k hits)
- 8: 72s (209k hits)
- 10: 67s (212k hits)
- 12: 64s (214k hits)
- 16: 62s (217k hits)
- 20: 63s (218k hits)
- 24: 65s (218k hits)

As expected, when increasing the value the number of hits increases (but gets slower), and also the number of elements in the evaluations increase, making the evaluation slower.  Around 16 seems to be ideal in this case.

** 5x5 board**
I used the same command than 9x9 but with `-N 5 -W 3`

Running it in main took 8s, and when clearing the nodes was about the same.
I tried a few different values of mcts_pre_evaluate_nodes_total (in main) and got:
- 4: 8.2s
- 8: 7.4s
- 64: 8s (the default)

With this PR and different values of mcts_top_k_pre_evaluate I got (out of 16k evaluations)
- 4: 7.6s (12k hits)
- 8: 6.9s (12.5k hits)
- 16: 6.5s (13k hits)
- 20: 6.6s (13k hits)
- 24: 6.6s (13k hits)

So, again around 16 works the best.   With higher numbers of mcts_n we may need a higher value, because it's more likely that we'll explore more of its children.

### DIFF
--- a/deep_quoridor/src/agents/alphazero/alphazero.py
+++ b/deep_quoridor/src/agents/alphazero/alphazero.py
@@ -112,9 +112,14 @@ class AlphaZeroParams(SubargsBase):
     mcts_noise_epsilon: float = 0.25
     mcts_noise_alpha: Optional[float] = None
 
-    # Number of nodes to pre-evaluate in MCTS. This is to take advantage of the GPU
-    # being efficient in batching.
-    mcts_pre_evaluate_nodes_total: int = 64
+    # When expanding a node, we take the top k children based on the prior (since those are the nodes that will
+    # be expanded first), and queue them for piggy-back evaluation.
+    # A value too high will compute many evaluations that are not used, and a value too low will make it run more evaluations,
+    # making it slower in both cases.
+    # The optimal value will depend on mcts_n as well as the board complexity, requiring a higher value when either increases.
+    # Having said that, it may not be worth trying to tune this to the ideal value, but if it's too far off it will slow it down.
+    # Notice that this value won't change the results, just the speed.
+    mcts_top_k_pre_evaluate: int = 16
 
     # If wandb_alias is provided, the model will be fetched from wandb using the model_id and the alias
     wandb_alias: Optional[str] = None
@@ -212,7 +217,7 @@ class AlphaZeroAgent(TrainableAgent):
             self.max_steps,
             self.evaluator,
             self.visited_states,
-            params.mcts_pre_evaluate_nodes_total,
+            params.mcts_top_k_pre_evaluate,
         )
 
         if params.training_mode and params.train_every is not None:
--- a/deep_quoridor/src/agents/alphazero/mcts.py
+++ b/deep_quoridor/src/agents/alphazero/mcts.py
@@ -124,7 +124,7 @@ class MCTS:
         max_steps: int,  # -1 means no limit
         evaluator,
         visited_states: set,
-        pre_evaluate_nodes_total: int = 64,
+        top_k_pre_evaluate: int = 16,
     ):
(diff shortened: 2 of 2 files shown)


alejandromarcu:
I also tried with a regular board and mcts_n=5000, and for different values of mcts_top_k_pre_evaluate I got (960k evaluations)
-8: 205s (901k hits)
-16: 194s (914k hits)
-20: 198s (918k hits)

Again, the best seems to be around 16 this time.  I think what happens is that the branching is exponential, so for example when doubling mcts_n, we don't get twice as many children in average, we get a logarithmic increase.


So, I tried with a much smaller mcts_n=100 (because if I keep going higher it's going to take too long).
I got (out of 11k eval):

1: 4.7s (17k hits out of 19k eval, the number of evals increased!)
2: 4.7s (8.9k hits)
3: 4.8s (8.9k hits)
4: 4.8s (9k hits)
6: 5s (9k hits)
8: 5.1s (9.1k hits)
12: 5.4s (9.2k hits)
16: 5.6s (9.3k hits)

So in this case we should use 1 or 2, but using 16 is not terrible (and probably we shouldn't use mcts_n=100 anyway in a normal board because it's even less than the max number of children)

jonbinney:
I think the best value for mcts_top_k_pre_evaluate will be pretty different on different machines. For example I ran the command you posted on my desktop with an nvidia 2070 with different values of mcts_to_k_pre_evaluate to see how long it took:
```
0: 51.766s
1: 55.589s
2: 6.317s
4: 62.561s
8: 65.483s
16: 71.844s
32: 82.753s
64: 122.795s
```
I was surprised because I thought the slower communication between the CPU and the GPU on my machine would favor evaluating more nodes at once, but the result suggests the opposite.

PR #291, opened 2025-09-26
Log loss in WandB and other fixes

alejandromarcu:
The main thing in this PR is to log the loss in WandB so we easily see how it evolves.  I ended up adding a lot of other changes, some related some completely orthogonal. 
The loss track ended up being tricky because we were using "steps" but you can only increase them monotonically and they need to be int, and I needed a different way to track losses.  I find it very confusing the way wandb does it, but I defined the metrics "Loss step" and "Episode" which would be the x axis:

```
        wandb.define_metric("Loss step", hidden=True)
        wandb.define_metric("Episode", hidden=True)
```

And then I indicated that all the metrics starting with loss_ will use 'Loss Step' and the rest 'Episode`:
```
        wandb.define_metric("loss_*", "Loss step")
        wandb.define_metric("*", "Episode")
```

E.g. from [this run](https://wandb.ai/the-lazy-learning-lair/deep_quoridor/runs/amarcu-20250926-000648?nw=nwuseramarcu)

image

The training for each epoch is in the X coordinates between that epoch number and extends for 0.5, so we can distinguish epochs.

E.g. when zooming in:
image
Between 0 and 0.5 I see how the training evolved after epoch 0.  Then from 1 to 1.5 how it evolved in epoch 1 and so on.

### DIFF
--- a/deep_quoridor/src/agents/alphazero/alphazero.py
+++ b/deep_quoridor/src/agents/alphazero/alphazero.py
@@ -11,6 +11,7 @@ from typing import Optional, Tuple
 import numpy as np
 import torch
 import wandb
+import wandb.wandb_run
 from quoridor import ActionEncoder, MoveAction, construct_game_from_observation
 from utils import get_initial_random_seed, my_device, resolve_path
 from utils.subargs import SubargsBase
@@ -238,11 +239,6 @@ class AlphaZeroAgent(TrainableAgent):
         # Just to make the training status plugin happy
         self.epsilon = 0.0
 
-        # Avoid circular imports
-        from metrics import Metrics
-
-        self.metrics = Metrics(board_size, max_walls)
-
         self.first_replay_buffer_saved = False
         self.first_replay_buffer_loaded = False
 
@@ -252,6 +248,11 @@ class AlphaZeroAgent(TrainableAgent):
                 print("Running bootstrap training iteration...")
                 self.train_iteration(is_replay_buffer_bootstrap=True)
                 self.first_replay_buffer_loaded = True
+
+        self.wandb_run = None
+
+    def set_wandb_run(self, wandb_run: wandb.wandb_run.Run):
+        self.wandb_run = wandb_run
 
     # TO DO, this was copy-pasted from AbstractTrainableAgent, need to refactor it
     def _fetch_model_from_wandb_and_update_params(self):
@@ -270,16 +271,6 @@ class AlphaZeroAgent(TrainableAgent):
         print(f"{self.name()} - Fetching model from wandb: {path}")
         artifact = api.artifact(path, type="model")
         local_filename = resolve_path(self.params.wandb_dir, self.wandb_local_filename(artifact))
-
-        param_fields = set(self.params_class().__dataclass_fields__.keys())
-        filtered_metadata = {k: v for k, v in artifact.metadata.items() if k in param_fields}
-
-        all_params = self.params_class()(**filtered_metadata)
-
-        # Override params, but only the ones that are not training only
-        for key, value in artifact.metadata.items():
-            if key not in all_params.training_only_params():
-                setattr(self.params, key, value)
 
         self.params.model_filename = str(local_filename)
 
@@ -415,13 +406,32 @@ class AlphaZeroAgent(TrainableAgent):
 
         return float(avg_loss), 0.0
 
-    def train_iteration(self, is_replay_buffer_bootstrap=False) -> bool:
+    def train_iteration(self, is_replay_buffer_bootstrap=False, epoch=None) -> bool:
         """
         Train the neural network on collected data.
 
         Returns:
             True if training was done, False if not enough data was available.
         """
+
+        def log_loss_entry(entry):
+            i = entry["step"]
+            t, p, v = entry["loss_total"], entry["loss_policy"], entry["loss_value"]
+
+            if "loss_total_val" in entry:
+                vt, vp, vv = entry["loss_total_val"], entry["loss_policy_val"], entry["loss_value_val"]
+                print(f"{i:>7}     {t:>7.3f} {vt:>7.3f}     {p:>7.3f} {vp:>7.3f}     {v:>7.3f} {vv:>7.3f}")
+            else:
+                print(f"{i:>7} {t:>7.3f} {p:>7.3f} {v:>7.3f}")
+
+            if self.wandb_run:
+                # Log this to take care of half of the epoch, so we can
+                # see clearly the different phases of training
+                entry["Loss step"] = epoch + entry["completion"] * 0.5
+                del entry["step"]
+                del entry["completion"]
+                self.wandb_run.log(entry)
+
         if len(self.replay_buffer) * (1.0 - self.params.validation_ratio) < self.params.batch_size:
             return False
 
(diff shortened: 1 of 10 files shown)


deep_quoridor/src/agents/alphazero/alphazero.py
```diff
@@ -491,7 +503,7 @@ def save_replay_buffer_to_file(self, episode_number: int):
 
     def load_replay_buffer_from_file(self) -> bool:
         """Load replay buffer from file if it exists. Returns True if the file exists."""
-        if self.params.train_every is None:
+        if self.params.train_every is None or not self.is_training():
```
alejandromarcu, line 506: It was trying to load the replay buffer when training and causing an error

deep_quoridor/src/agents/alphazero/nn_evaluator.py
```diff
@@ -187,6 +188,9 @@ def split_data(self, replay_buffer, validation_ratio: float) -> tuple[list, list
         return training_set, validation_set
 
     def compute_losses(self, batch_data):
+        if len(batch_data) == 0:
+            return None, None, None
```
alejandromarcu, line 192: this makes simpler the code that calls it

deep_quoridor/src/agents/alphazero/nn_evaluator.py
```diff
@@ -221,44 +225,69 @@ def compute_losses(self, batch_data):
...
-        return {"total_loss": total_loss.item(), "policy_loss": policy_loss.item(), "value_loss": value_loss.item()}
+        # Get a last entry for the losses after the last backprop
+        with torch.no_grad():
+            batch_data = random.sample(training_set, self.batch_size)
+            policy_loss, value_loss, total_loss = self.compute_losses(batch_data)
+            v_policy_loss, v_value_loss, v_total_loss = self.compute_losses(validation_set)
+            entry = make_entry(i + 1, total_loss, policy_loss, value_loss, v_total_loss, v_policy_loss, v_value_loss)
+
+            if on_new_entry is not None:
+                on_new_entry(entry)
```
alejandromarcu, line 291: We were not computing the loss after the last training step, but I think it's good to have it as well, and allows us to see how the loss changes when we have different data in the next epoch

deep_quoridor/src/agents/alphazero/self_play_manager.py
```diff
@@ -64,6 +63,14 @@ def run(self):
...
             self._run_central_evaluation()
 
+    # Return an array with the number of games per worker, such that they add up to the total number of games
+    # and each worker gets X or X+1 jobs.
+    # E.g. for 20 games and 8 workers it returns: [3, 3, 3, 3, 2, 2, 2, 2]
+    def _games_per_worker(self):
+        base = self.num_games // self.num_workers
+        extra = self.num_games - base * self.num_workers
+        return [base + 1 if i < extra else base for i in range(self.num_workers)]
+
```
alejandromarcu, line 73: I noticed when playing 20 games with 8 workers that it was assigning 3 per worker, so it was playing 24 games.
Then I realized it was missing `games_remaining_to_allocate -= this_worker_num_games`, but this didn't work as I expected, since in that case it was doing 6 workers with 3 games, 1 worker with 2 and another worker empty.

deep_quoridor/src/metrics.py
```diff
@@ -12,11 +12,20 @@ class Metrics:
...
+        benchmarks: list[str | Agent] = [],
+        benchmarks_t: int = 10,
+        max_steps=200,
+    ):
         self.board_size = board_size
         self.max_walls = max_walls
         self.stored_elos = {}
         self.benchmarks = benchmarks
+        self.benchmarks_t = benchmarks_t
+        self.max_steps = max_steps
```
alejandromarcu, line 28: This ones were hardcoded before, and I made it so that you can specify from the arguments now

deep_quoridor/src/plugins/wandb_train.py
```diff
@@ -148,23 +147,19 @@ def compute_tournament_metrics(self, model_filename: str) -> int:
             self.best_model_filename = model_filename
             print("Best Relative Elo so far!")
 
-        wandb_elo_table = wandb.Table(
-            columns=["Player", "elo"], data=[[player, elo] for player, elo in elo_table.items()]
-        )
```
alejandromarcu, line 153: this table was useless, so just removing it

deep_quoridor/src/quoridor.py
```diff
@@ -60,7 +60,7 @@ def __init__(self, board_size: int):
         self.board_size = board_size
         self.wall_size = board_size - 1
         self.num_actions = self.board_size**2 + self.wall_size**2 * 2
-        self.constant_action_mask_template = np.zeros(self.num_actions, dtype=np.float32)
+        self.constant_action_mask_template = np.zeros(self.num_actions, dtype=np.int8)
```
alejandromarcu, line 63: When playing against greedy:p_random=0.5, I was getting the following error:

```
  File "/Users/amarcu/code/deep_rabbit_hole/deep_quoridor/src/metrics.py", line 130, in compute
    results = arena._play_games(players, self.benchmarks_t, PlayMode.ALL_VS_ALL)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/amarcu/code/deep_rabbit_hole/deep_quoridor/src/arena.py", line 180, in _play_games
    result = self._play_game(agent_1, agent_2, f"game_{match_id:04d}")
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/amarcu/code/deep_rabbit_hole/deep_quoridor/src/arena.py", line 94, in _play_game
    action = int(agent.get_action(observation_before_action))
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/amarcu/code/deep_rabbit_hole/deep_quoridor/src/agents/greedy.py", line 185, in get_action
    return self.action_space.sample(action_mask)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/amarcu/code/deep_rabbit_hole/.venv/lib/python3.12/site-packages/gymnasium/spaces/discrete.py", line 93, in sample
    mask.dtype == np.int8
```

So I just changed it to be int8 as requested.  Notice that if you are running with -O the assertion is disabled, it took me a while to understand why when training it was not giving that error (I was copy pasting a training with -O, but I was running something else without -O)

deep_quoridor/src/train_alphazero.py
```diff
@@ -46,6 +47,12 @@ def train_alphazero(
         # members, so we pass in a GameParams object. We have to call start_game
         # because it calls the plugin's internal _intialize method which sets up metrics.
         wandb_train_plugin.start_game(game=args, agent1=training_agent, agent2=training_agent)
+        training_agent.set_wandb_run(wandb_train_plugin.run)
+
+        # Compute the tournament metrics with the initial model, possibly random initialized, to
+        # be able to see how it evolves from there
+        wandb_train_plugin.episode_count = 0
+        wandb_train_plugin.compute_tournament_metrics(str(current_filename))
```
alejandromarcu, line 55: The untrained model can already beat some agents during MCTS, so it's good to have it as a base

deep_quoridor/src/agents/alphazero/self_play_manager.py
jonbinney: I definitely need to do a better job of debugging before making PR's :-( Sorry!

deep_quoridor/src/metrics.py
jonbinney: Ah, this is great! I've been wishing that the number of games in the benchmarks was configurable!

deep_quoridor/src/quoridor.py
jonbinney: I think that @muralx used float32 here to make the vector easier to do math with during training? But I think it makes sense as an int8 if that's what PettingZoo expects.

jonbinney (approved):
LGTM! Lots of good changes.

deep_quoridor/src/agents/alphazero/alphazero.py
```diff
@@ -438,12 +448,14 @@ def train_iteration(self, is_replay_buffer_bootstrap=False) -> bool:
...
+            print("  Epoch       Total   Val Total   Policy  Val Policy  Value   Val Value")
+        else:
+            print("==== Training Loss ====")
+            print("  Epoch   Total   Policy  Value")
 
-        # Store loss for metrics
-        self.recent_losses.append(metrics["total_loss"])
-        if len(self.recent_losses) > 100:  # Keep only recent losses
-            self.recent_losses = self.recent_losses[-100:]
+        self.evaluator.train_iteration(self.replay_buffer, self.params.validation_ratio, on_new_entry=log_loss_entry)
```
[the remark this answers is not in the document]
alejandromarcu: We could consider making it into a generator (i.e. a function that uses yield to give control to the caller, like range).
Then we would do something like:
```
for entry in evaluator.train_step(num_batches=100):
    # entry will have the losses for the step
```

PR #293, opened 2025-10-05
Time self-play and benchmarks

alejandromarcu:
I wanted to measure how long it takes to do the self-play and running the tournament benchmarks, so it's easier to measure gains when doing performance work in just a specific part, and I think in general it's interesting to see.
I created the Timer class to make it easier to use it.
I tried to also measure the time in evaluate and mcts, but the aggregation doesn't work because they are in separate process, we need a bit more work for that to happen (e.g. when the process is done, send the aggregated data back to the root process).

E.g. when I ran:

```
.venv/bin/python -O deep_quoridor/src/train_alphazero.py -N 5 -W 3 \
-g 100 -e 8 --max-steps=50 --num-workers 6 \
-p replay_buffer_size=50000,mcts_ucb_c=1.5,\
save_replay_buffer=first,mcts_n=500,\
learning_rate=0.0001,batch_size=16,optimizer_iterations=10,\
validation_ratio=0.2 \
-w project="cucutest"
```
I got this in wandb:

image

### DIFF
--- a/deep_quoridor/src/train_alphazero.py
+++ b/deep_quoridor/src/train_alphazero.py
@@ -10,7 +10,7 @@ from agents.alphazero.self_play_manager import GameParams, SelfPlayManager
 from agents.core.agent import AgentRegistry
 from metrics import Metrics
 from plugins.wandb_train import WandbParams, WandbTrainPlugin
-from utils import parse_subargs, set_deterministic
+from utils import Timer, parse_subargs, set_deterministic
 
 
 def train_alphazero(
@@ -61,6 +61,8 @@ def train_alphazero(
         # Set the filename so that each process loads the most recent model.
         self_play_params.model_filename = str(current_filename)
 
+        Timer.start("self-play")
+
         # Create a self-play manager to run self-play games across multiple processes. The
         # worker processes get re-spawned each epoch to make sure all cached values get freed.
         self_play_manager = SelfPlayManager(
@@ -77,6 +79,9 @@ def train_alphazero(
             new_replay_buffer_items = self_play_manager.get_results(timeout=0.1)
         training_agent.replay_buffer.extend(new_replay_buffer_items)
         self_play_manager.join()
+        game_num = (epoch + 1) * args.games_per_epoch
+
+        Timer.finish("self-play", game_num)
 
         # Do training if we have enough samples in the replay buffer.
         # training_agent.episode_count = game_num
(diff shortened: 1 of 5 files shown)


deep_quoridor/src/train_alphazero.py
```diff
@@ -77,23 +79,27 @@ def train_alphazero(
...
         # Do training if we have enough samples in the replay buffer.
         # training_agent.episode_count = game_num
         training_occured = training_agent.train_iteration(epoch=epoch)
         if not training_occured:
             print("Not enough samples - skipping training")
 
-        game_num = (epoch + 1) * args.games_per_epoch
         current_filename = training_agent.save_model_with_suffix(f"_epoch_{epoch}")
-        if wandb_train_plugin is not None:
+        if wandb_train_plugin is not None and epoch < args.epochs - 1:
```
alejandromarcu, line 93: This is an unrelated change, it was computing the tournament here for the last epoch, and then again below when doing end_arena, so I removed this one, since the other does a few more things like uploading the models

deep_quoridor/src/utils/timer.py
```diff
@@ -0,0 +1,83 @@
...
+    Class to measure elapsed time and log it to the console and optionally wandb.
+    It's a class just to create a container for the functions and variables, it can't be instantiated.
+    TO DO: would be nice if it would easily support multi-process, so we can gather information coming
+    from different processes and compile it.
+    """
+
+    starts = {}
+    counts = {}
+    totals = {}
+    wandb_run: Optional[wandb.wandb_run.Run] = None
```
jonbinney, line 21: Since wandb_run is a class variable and not an instance variable, this means that if you create multiple timer instances and set the wandb_run for one of them, they all will use that, right? (this seems like a good thing)

deep_quoridor/src/utils/timer.py
```diff
@@ -0,0 +1,83 @@
...
+import wandb.wandb_run
+
+from utils.misc import format_time
+
+
+class Timer:
+    """
+    Class to measure elapsed time and log it to the console and optionally wandb.
+    It's a class just to create a container for the functions and variables, it can't be instantiated.
+    TO DO: would be nice if it would easily support multi-process, so we can gather information coming
```
jonbinney, line 14: I think that the "right" way to do multiprocess logging with wandb is to have a separate wandb run for each process, and have them all be in the same wandb "group". I prototyped that a while back and it works well, but i didn't PR it and our code has changed quite a bit since then. It's a bit ugly now that we start new worker processes for each self-play session - that will generate a lot of runs. But if we fix the memory leak then it would work well. It's pretty easy to generate graphs that show all the info for all runs in the wandb group.

deep_quoridor/src/utils/timer.py
```diff
@@ -0,0 +1,83 @@
...
+        cls.wandb_run = wandb_run
+
+    @classmethod
+    def start(cls, name: str):
+        if name in cls.starts:
+            print(f"TIMER: WARNING - timer for {name} was already started")
+
+        cls.starts[name] = time.perf_counter()
+
+    @classmethod
```
jonbinney, line 37: Having these be class methods makes some things easier - we'll just need to be careful if we ever add multi-threading.

jonbinney (approved):
I've always wished python had some built-in convenience class like this for timing. Thanks for adding it! It's also pretty interesting that we're spending so much time on benchmarking.

deep_quoridor/src/utils/timer.py
alejandromarcu: You can't create instances of this class, the method below won't let you.  The idea is just to use the static methods.  I could have created just normal functions, but I liked this as a way of keeping everything in the same "namespace"

PR #294, opened 2025-10-06
Self-play games in batches in each process

alejandromarcu:
Currently, each process plays one game at a time, so when we call the evaluator, we have only 1 state to evaluate (I did the pre-evaluation thing but didn't help much, e.g. in Jon's computer it was faster with 0).  This technique of playing in batches seems to be common (e.g. The Kid explains it), and the idea is that you take N games, do the first move in the N games, then do the second move for the N games and so on.  The advantage is that you can then just batch the N evaluations together, which makes it significantly faster.

I ended up removing mcts_top_k_pre_evaluate because sometimes it was performing worse, sometimes a bit better, but I thought it was not worth the complexity. 

The result is that this runs a bit more than 2x faster.

For example, running the following:
```
time .venv/bin/python -O deep_quoridor/src/train_alphazero.py -N 5 -W 3 \
-g 500 -e 1 --max-steps=50 --num-workers 8 \
-p replay_buffer_size=50000,mcts_ucb_c=1.5,\
save_replay_buffer=first,mcts_noise_epsilon=0.15,mcts_n=1000,\
learning_rate=0.0001,batch_size=256,optimizer_iterations=10,\
validation_ratio=0.2 
```
Took 358 seconds in main, vs 175 in this PR with the default of 32 parallel games (seems to be the fastest in my machine).  So that's a 2x in speed.

To make sure this is still working, I ran a longer training session, and verified that the results are similar to the base in main:
image

There's a significant speedup, but not as much as the above because the time to play the tournaments for the metrics is still the same and it's a significant percentage of the total time.
Additionally, I made the output to console more compact, e.g. it looks like this now:

```
Worker 3: (37.64s) Games 83...101 / 83 ended after [14, 14, 15, 16, 17, 17, 17, 17, 18, 18, 18, 19, 21, 21, 25, 26, 26, 26, 29] turns. 0 truncated
Worker 2: (40.77s) Games 83...101 / 83 ended after [14, 14, 15, 17, 17, 18, 18, 18, 18, 18, 18, 20, 22, 30, 33, 33, 45, 46] turns. 1 truncated
Worker 1: (37.48s) Games 84...103 / 84 ended after [10, 13, 14, 16, 16, 17, 17, 18, 18, 18, 18, 18, 18, 20, 21, 22, 29, 38, 41] turns. 1 truncated
Worker 0: (38.98s) Games 84...103 / 84 ended after [14, 15, 16, 16, 17, 18, 18, 20, 20, 20, 20, 22, 22, 24, 25, 26, 28, 29, 40] turns. 1 truncated
Worker 5: (42.62s) Games 83...101 / 83 ended after [11, 14, 16, 17, 18, 18, 18, 20, 21, 22, 22, 24, 26, 33, 33, 33, 42, 44] turns. 1 truncated
Worker 4: (35.55s) Games 83...101 / 83 ended after [12, 15, 15, 16, 16, 16, 16, 17, 18, 18, 18, 18, 18, 21, 21, 22, 26, 29, 46] turns. 0 truncated
```
So, we have all the parallel games in one line. If we wanted we could make it even more compact by doing something like [12x3, 15x2, 16, 18x5], etc, or even just showing the quartiles or some statistical information to give a sense of the length of the matches.

### DIFF
--- a/deep_quoridor/src/agents/alphazero/alphazero.py
+++ b/deep_quoridor/src/agents/alphazero/alphazero.py
@@ -118,15 +118,6 @@ class AlphaZeroParams(SubargsBase):
     mcts_noise_epsilon: float = 0.25
     mcts_noise_alpha: Optional[float] = None
 
-    # When expanding a node, we take the top k children based on the prior (since those are the nodes that will
-    # be expanded first), and queue them for piggy-back evaluation.
-    # A value too high will compute many evaluations that are not used, and a value too low will make it run more evaluations,
-    # making it slower in both cases.
-    # The optimal value will depend on mcts_n as well as the board complexity, requiring a higher value when either increases.
-    # Having said that, it may not be worth trying to tune this to the ideal value, but if it's too far off it will slow it down.
-    # Notice that this value won't change the results, just the speed.
-    mcts_top_k_pre_evaluate: int = 16
-
     # If wandb_alias is provided, the model will be fetched from wandb using the model_id and the alias
     wandb_alias: Optional[str] = None
 
@@ -222,7 +213,6 @@ class AlphaZeroAgent(TrainableAgent):
             self.max_steps,
             self.evaluator,
             self.visited_states,
-            params.mcts_top_k_pre_evaluate,
         )
 
         if params.training_mode and params.train_every is not None:
@@ -383,12 +373,12 @@ class AlphaZeroAgent(TrainableAgent):
         model_state = torch.load(path, map_location=my_device())
         self.set_model_state(model_state)
 
-    def multi_start_game(self, envs):
+    def start_game_batch(self, envs):
         self.visited_states.clear()
         self.game_envs = envs
         self.replay_buffers_in_progress = [[] for _ in range(len(envs))]
 
-    def multi_end_game(self):
+    def end_game_batch(self):
         if not self.params.training_mode:
             return
 
@@ -577,7 +567,7 @@ class AlphaZeroAgent(TrainableAgent):
         self.action_log.action_score_ranking(scores)
         self.action_log.action_text(root_action, f"{root_value:0.2f}")
 
-    def multi_get_action(self, observations_with_ids: list[tuple[int, dict]]) -> list[tuple[int, int]]:
+    def get_action_batch(self, observations_with_ids: list[tuple[int, dict]]) -> list[tuple[int, int]]:
         if not observations_with_ids:
             return []
 
@@ -597,7 +587,7 @@ class AlphaZeroAgent(TrainableAgent):
         #     root_childrens.append(root_children)
         #     root_values.append(root_value)
 
-        root_childrens, root_values = self.mcts.multi_search(games)
+        root_childrens, root_values = self.mcts.search_batch(games)
 
         actions = []
         for game_idx, root_children, root_value, game, player in zip(
@@ -609,10 +599,14 @@ class AlphaZeroAgent(TrainableAgent):
                 raise RuntimeError("No nodes visited during MCTS")
 
             visit_probs = visit_counts / visit_counts_sum
-            # TODO disabled on multi
-            self._log_action(
-                visit_probs, root_children, float(root_value), MoveAction(game.board.get_player_position(player))
-            )
+
+            # _log_action is used to display information about the action (e.g. probabilities of moves) in the GUI.
+            # We allow that only when there's one game at the time, since we won't be playing multiple games and
+            # displaying them
+            if len(observations_with_ids) == 1:
+                self._log_action(
+                    visit_probs, root_children, float(root_value), MoveAction(game.board.get_player_position(player))
+                )
 
             temperature = self.initial_temperature
             if self.params.drop_t_on_step is not None and game.completed_steps >= self.params.drop_t_on_step:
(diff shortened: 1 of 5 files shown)


deep_quoridor/src/agents/alphazero/alphazero.py
```diff
@@ -242,6 +232,7 @@ def __init__(
                 self.initial_temperature = 0
 
         self.replay_buffer = deque(maxlen=params.replay_buffer_size)
+        self.replay_buffers_in_progress = []
```
alejandromarcu, line 235: Now we store the replay buffer(s) in progress (i.e. the current game or games) here, and when the game is done we set the value for each entry and store it in the replay_buffer

deep_quoridor/src/agents/alphazero/alphazero.py
```diff
@@ -546,56 +563,88 @@ def _log_action(
...
+        games = []
+        players = []
+        game_indices = []
+        for game_idx, observation in observations_with_ids:
+            g, p, _ = construct_game_from_observation(observation["observation"])
+            games.append(g)
+            players.append(p)
+            game_indices.append(game_idx)
+
+        root_children_batch, root_value_batch = self.mcts.search_batch(games)
```
alejandromarcu, line 583: this is the main change for this method: instead of doing mcts search for just one game, we do it for many games, and we get arrays of results.  Below, we iterate through those results, but hat we end up doing is almost the same.  It's hard to follow what changed in the diff

deep_quoridor/src/agents/alphazero/alphazero.py
```diff
@@ -546,56 +563,88 @@ def _log_action(
...
+                    policy_target[action_index] = child.visit_count / visit_counts_sum
+                self.store_training_data(game, policy_target, player, game_idx)
+
+        return actions
+
+    def get_action(self, observation) -> int:
+        action_batch = self.get_action_batch([(0, observation)])
+        return action_batch[0][1]
+
+    def store_training_data(self, game, mcts_policy, player, game_idx):
```
alejandromarcu, line 641: Added game_idx, because the training data for each parallel game needs to be store separately, so that when we finish we set the right value.  (Actually now that I think about it I could have stored everything in 1 replay buffer and add the game_idx as an entry, but not sure if it makes anything better anyway)

deep_quoridor/src/agents/alphazero/mcts.py
```diff
@@ -147,32 +145,38 @@ def select(self, node: Node) -> Node:
...
             assert self.n is not None, "n must be specified if k is not"
-            num_iterations = self.n
+            num_iterations = [self.n for _ in initial_games]
+
+        max_iterations = max(num_iterations)
+
+        roots = [Node(g, ucb_c=self.ucb_c) for g in initial_games]
+        for iteration in range(max_iterations):
+            need_evaluation = []  # (root, node)
+            for game_idx, root in enumerate(roots):
```
alejandromarcu, line 160: Now we need to perform the search in each one of the roots, so this is the added loop

deep_quoridor/src/agents/alphazero/mcts.py
```diff
@@ -147,32 +145,38 @@ def select(self, node: Node) -> Node:
...
-            else:
-                games_to_evaluate = [n.game for n in new_nodes]
-                new_nodes = []
+                if node.game.is_game_over():
+                    # The player who just made a move must have won.
+                    node.backpropagate_result(1)
+                elif self.max_steps >= 0 and node.game.completed_steps >= self.max_steps:
+                    node.backpropagate_result(0)
+                else:
+                    need_evaluation.append((root, node))
```
alejandromarcu, line 174: Instead of evaluating the games, we gather what we need and evaluate a batch below

deep_quoridor/src/agents/alphazero/nn_evaluator.py
```diff
@@ -73,37 +73,42 @@ def evaluate_tensors(
...
+        action_masks = torch.stack([torch.from_numpy(g.get_action_mask()) for g in games_to_evaluate]).to(
+            device=self.device
+        )
 
-        for i, g in enumerate(all_games):
-            pm = policy_masked[i]
-            # Sanity checks
-            assert np.all(pm >= 0), "Policy contains negative probabilities"
-            assert np.abs(np.sum(pm) - 1) < 1e-6, "Policy does not sum to 1"
-            assert np.isfinite(values[i]), "Policy or value is non-finite"
```
alejandromarcu, line 99: We are doing those assertions in evaluate_tensor, so I don't think it's useful to have them here as well

deep_quoridor/src/agents/alphazero/self_play_manager.py
```diff
@@ -170,28 +180,50 @@ def run_self_play_games(
...
 
-            action_index = alphazero_agent.get_action(observation)
-            environment.step(action_index)
-            num_turns += 1
+        num_turns = [0] * n
+        finished_in = []
+        finished = [False] * n
 
+        t0 = time.time()
+        while not all(finished):
```
alejandromarcu, line 197: this is the added loop to go through all the games

alejandromarcu:
> Looks good and I'm excited to see how fast it runs on my machine! I didn't debug every line, but it looks like it does what it is intended to do. Once this is merged I'll do some more testing.

@jonbinney , when you test it, try a different number of workers, maybe less.  E.g. for me 8 is the best without parallel games, but with parallel games using 6 workers performs a bit better.

PR #295, opened 2025-10-07
Add Resnet option to NNEvaluator

jonbinney:
There is a new param in AlphazeroParams called "nn_type" which defaults to "mlp" but can be changed to "resnet". Some of this code was generated by claude. I've gone through and read it and compared to the resnet structure used by alphazero, and I think it is doing the right thing.

In a future PR, I'd like to make the number of channels and number of residual blocks of resnet configurable via parameters. For now I've set the number of channels to 32 and the number of blocks to the width of the combined grid times 2 plus 2. Alphazero used 256 channels internally, but that is very slow - if things work well we can up it in later training runs. Alphazero used 20 residual blocks for chess and 40 for Go, so I followed the pattern of "a bit more than twice the width of the input array".

Since the resnet is much bigger than the mlp, it is also much slower. Self-play is ok when run on a big machine in lots of parallel processes. Training is still fast. Benchmarking is veeeeery slow since it only uses one process.

Here's the command I'm using to train on a hyperstack machine with an Nvidia L40:
```
train_alphazero.py -N 5 -W 3 -g 500 -e 10000 --max-steps=40 --num-workers 20 -w --benchmarks random greedy simple:branching_factor=8,nick=simple-bf8 simple:branching_factor=16,nick=simple-bf16 -p replay_buffer_size=50000,mcts_ucb_c=1.5,save_replay_buffer=first,mcts_noise_epsilon=0.25,mcts_n=2000,learning_rate=0.0001,batch_size=256,optimizer_iterations=100,validation_ratio=0.2,nn_type=resnet
```

### DIFF
--- a/deep_quoridor/src/agents/alphazero/resnet_network.py
+++ b/deep_quoridor/src/agents/alphazero/resnet_network.py
@@ -26,7 +26,7 @@ class ResidualBlock(nn.Module):
 
 
 class ResnetNetwork(nn.Module):
-    def __init__(self, action_encoder: ActionEncoder, device, num_blocks: Optional[int] = None, num_channels: int = 16):
+    def __init__(self, action_encoder: ActionEncoder, device, num_blocks: Optional[int] = None, num_channels: int = 32):
         """
         ResNet-based network following AlphaZero architecture.
 
--- a/deep_quoridor/src/agents/alphazero/resnet_network.py
+++ b/deep_quoridor/src/agents/alphazero/resnet_network.py
@@ -1,0 +1,141 @@
+from typing import Optional
+
+import numpy as np
+import torch
+import torch.nn as nn
+from quoridor import ActionEncoder, Player, Quoridor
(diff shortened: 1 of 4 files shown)


deep_quoridor/src/agents/alphazero/resnet_network.py
```diff
@@ -0,0 +1,145 @@
...
+
+        # Value head
+        value = torch.relu(self.value_bn(self.value_conv(out)))
+        value = value.view(value.size(0), -1)
+        value = torch.relu(self.value_fc1(value))
+        value = torch.tanh(self.value_fc2(value))
+
+        return policy, value
+
+    def game_to_input_array(self, game: Quoridor) -> np.ndarray:
```
jonbinney, line 111: The combined grid includes an extra 2 layers of solid walls around the board. I left this in since it seems like it could be useful to the convolution kernels.

PR #296, opened 2025-10-08
Run benchmarks every N epochs

alejandromarcu:
When we do longer runs with a lot of epochs, there's no need to run an evaluation after every epoch, and considering how slow it is, we are better off running it less often.
With this, you can pass `--benchmarks_every` to train_alphazero to do it once in every N epochs.
As a bonus I included 2 unrelated small fixes

### DIFF
--- a/deep_quoridor/src/train_alphazero.py
+++ b/deep_quoridor/src/train_alphazero.py
@@ -91,13 +91,14 @@ def train_alphazero(
             print("Not enough samples - skipping training")
 
         current_filename = training_agent.save_model_with_suffix(f"_epoch_{epoch}")
-        if wandb_train_plugin is not None and epoch < args.epochs - 1:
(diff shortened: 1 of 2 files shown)


deep_quoridor/src/train_alphazero.py
```diff
@@ -183,6 +184,13 @@ def main(args):
         default=10,
         help="How many time to play against each opponent during benchmarks",
     )
+    parser.add_argument(
+        "-be",
```
jonbinney, line 188: Technically in posix style command line options (which i think is pretty good) arguments with a single dash are only allowed to be one letter. This works fine in argparse though, so I don't feel too strongly about it.

alejandromarcu: ok, I'll leave it for now since we have a few more and eventually we can fix them all

PR #363, opened 2026-03-27
Consistency text fix: use production policy builder

adamantivm:
Use AlphaZeroAgent helpers in the real-model parity script so policy\nconstruction follows the production path and does not drift over time.\n\nAlso remove duplicated visit-probability logic in get_action_batch by\nsharing a single helper used for both sampling and policy targets.


adamantivm:
The test passes (it's in the CI) but more importantly, I was able to confirm that if I hack the policy values in python it now fails.

The PR description is screwed up b/c I impatiently closed the PR that Claude had opened and then just copy/pasted it in a new one / quick and dirty

PR #364, opened 2026-03-31
Evaluate model using policy database

jonbinney:
Adds a new script `train_policy_db_evaluator.py` which trains an NNEvaluator on the policy from a database.

There's also a few other fixes, and I've moved policy saving/loading into `policy_database.rs`


adamantivm:
LGTM

PR #366, opened 2026-04-21
Use a learning rate scheduler and fix bug with LR and weight decay

alejandromarcu:
Claude recommended using a LR scheduler, in particular this one, so I made it implement it.
Here's how the LR looks using it:
image

I notices that the initial LR was 0.001 despite the configuration saying 0.0002, which ended up being a bug of some config parameters not copied, so Claude found and fixed the bug.  That bug made in the past changes in LR and weight_decay useless 🤦

### DIFF
--- a/deep_quoridor/src/v2/common.py
+++ b/deep_quoridor/src/v2/common.py
@@ -127,6 +127,8 @@ def alphazero_params_dict_from_config(
     params_dict = {
         "mcts_n": config.alphazero.mcts_n,
         "mcts_ucb_c": config.alphazero.mcts_c_puct,
+        "learning_rate": config.training.learning_rate,
+        "weight_decay": config.training.weight_decay,
     }
 
     if config.training.initial_model:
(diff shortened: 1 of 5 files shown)


deep_quoridor/src/v2/common.py
```diff
@@ -127,6 +127,8 @@ def alphazero_params_dict_from_config(
     params_dict = {
         "mcts_n": config.alphazero.mcts_n,
         "mcts_ucb_c": config.alphazero.mcts_c_puct,
+        "learning_rate": config.training.learning_rate,
```
jonbinney, line 130: Is this the fix for the bug you mentioned about the learning rate being ignored?

jonbinney (approved):
Very cool! I'm glad there's a principled way to do this. I've tried reducing the learning rate over time, but never looked into the right way to do it.

deep_quoridor/src/v2/common.py
alejandromarcu: yes

PR #367, opened 2026-05-04
Better training from policy database

jonbinney:
Improvements:
- Can now download a precomputed policy database from W&B
- Switched database format from sqlite to parquet, which has made it much much faster.

First build the python extension
```
(cd deep_quoridor/rust && maturin develop --release)
```

Then run the script that trains an evaluator using a policy database. Using the `wandb:` syntax in this command tells it to download the policy database from W&B. I've only uploaded this one so far; it's for 5x5, 2 walls, max_steps=20.

```
python deep_quoridor/src/train_policy_db_evaluator.py wandb:the-lazy-learning-lair/policydb/policydb_5_2_20:latest -p nn_type=resnet,nn_resnet_num_blocks=2,nn_resnet_num_channels=32,learning_rate=0.001,batch_size=2048,weight_decay=0.0001 -w --test-batch-size 2048 --log-interval 10
```

Training stats will be logged to W&B where you can look at them.

### DIFF
--- a/deep_quoridor/src/quoridor.py
+++ b/deep_quoridor/src/quoridor.py
@@ -551,7 +551,15 @@ class Quoridor:
         )
 
     def __str__(self):
-        return str(self.board)
+        p1_walls = self.board.get_walls_remaining(Player.ONE)
+        p2_walls = self.board.get_walls_remaining(Player.TWO)
+        return (
+            f"{self.board}\n"
+            f"Current player: {self.current_player + 1}  |  "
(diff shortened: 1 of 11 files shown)


deep_quoridor/src/quoridor.py
```diff
@@ -551,7 +551,15 @@ def __eq__(self, other: "Quoridor") -> bool:
         )
 
     def __str__(self):
-        return str(self.board)
```
jonbinney, line 554: Now prints the rest of the game state below the board.

alejandromarcu (approved):
YOLO

PR #368, opened 2026-05-14
Fix corrupted ONNX checkpoints from save_model_onnx caching

jonbinney:
After this change, using `train_v2.py` with rust for self-play appears to be doing the right thing (good win_perc graph).

save_model_onnx cached the ONNX protobuf on the first call and only patched initializers on subsequent saves. With do_constant_folding=True, BatchNorm parameters get folded into preceding Conv weights, producing 21 transformed initializers from 55 state_dict tensors. The subsequent-save path then overwrote those folded initializers with the raw state_dict tensors, silently corrupting every checkpoint after model_0.

Rust self-play consumes these ONNX files; the corruption produced garbage NN evaluations and collapsed training (raw_win_perc 18% vs Python's 94%). Always do a full torch.onnx.export instead; the cost is negligible against the seconds-per-training-step budget.


alejandromarcu (approved):
yey, great catch!

PR #369, opened 2026-05-17
Use compact representation for rust selfplay

jonbinney:
I made this change in preparation for adding parallelism and caching to the rust selfplay. The compact states will save a lot of space. In this PR, I also updated the compact representation to handle the full B9W10 case, which takes 24 bytes instead of just 8 for B5W3. This slows down the the generation of the policy database a bit, but it isn't terrible. And the resulting policy database parquet files for B5W2 are actually about the same size, I assume because of compression.

There are a ton of changed lines in this PR for three main reasons:
- the game state type shows up in lots of places, so lots of code gets touched
- some things like board rotation and NN feature generation is now implemented for both grid and compact representations
- more tests

I've tested that B5W2 training still works with this PR, and that the policy database stuff still works. I've also uploaded new policy databases to W&B for B5W2 and B5W1.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -29,7 +29,7 @@
         "source.organizeImports": "explicit"
     },
     "rust-analyzer.linkedProjects": [
-        "./deep_quoridor/rust/Cargo.toml"
+        "${workspaceFolder}/deep_quoridor/rust/Cargo.toml"
     ],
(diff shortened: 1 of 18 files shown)


alejandromarcu (approved):
Can't really review the code but what you're doing LGTM

PR #371, opened 2026-05-31
Quoridor web player

jonbinney:
This PR includes two unrelated changes, since I am lazy: W&B metrics for rust self play and a web app for playing against the rust alphazero implementation.

### W&B Metrics for rust self-play

To avoid complicating the rust code too much, the rust self-play just logs metrics to files in the run directory being used. A separate python script, started by train_v2.py, reads those files and uploads the metrics to W&B. So far the metrics are all about the MCTS search. They provide some insight into how many paths in the search tree are being explored, how many reach terminal states, the entropy of the distribution at the root node, etc.

### Web-app (and server) for playing against rust implementation of alphazero

Mostly so we can easily try playing against new models without having to compile and run everything locally. One rust binary serves a simple JS+html page and acts as an HTTP API server. Not much intelligence in the frontend. Lets the user choose between any model that is in a particular directory. Each session gets its own state and can have its own game against the AI.

### DIFF
--- a/deep_quoridor/rust/Cargo.toml
+++ b/deep_quoridor/rust/Cargo.toml
@@ -39,6 +39,7 @@ clap = { version = "4.5", features = ["derive"], optional = true }
 ort = { version = "=2.0.0-rc.12", optional = true }
 anyhow = "1"
 serde_yaml = { version = "0.9", optional = true }
+serde_json = { version = "1", optional = true }
 ndarray-npy = { version = "0.9", optional = true }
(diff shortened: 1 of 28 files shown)


jonbinney:
Although this is a big change, it isn't as big as the diff seems. A couple thousand lines of changes are just the planning docs that claude made. Then there's all the web frontend and API stuff, which ..... isn't "real" code :-P

alejandromarcu (approved):

